PR #22, opened 2022-04-06
Helm charts refactor

ferpizza:
isolate containerRegistry from containerImage, so we can operate on each independently (eg. when moving to production, we want to use a different registry, with the same image and tag than in stage)


sunbit (approved):

ferpizza:
discarded in favor of #58

PR #56, opened 2022-04-25
ingest | add istio sidecar and port exceptions

ferpizza:
### Description
Add an istio sidecar to the Ingest component


bloodbare (changes requested):
Ingest connects to multiple nats, to cache, pubsub. It offers GRPC services throught the generic service for all sts...

bloodbare (approved):

PR #106, opened 2022-05-23
Chitchat

r3bu1ld3r:
### Description

### How was this PR tested?


nucliadb_cluster/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tracing = "0.1"
 uuid = "0.8"
 log = "0.4.14"
 env_logger = "0.9.0"
+chitchat = { git = "https://github.com/r3bu1ld3r/chitchat.git", branch="main" }
```
bloodbare, line 27: If we need a fork we should fork at nuclia org

bloodbare (changes requested):
PR fails on tests

nucliadb_cluster/Cargo.toml
r3bu1ld3r: Switched back on quickwit repository

PR #173, opened 2022-07-05
Updated nucliadb_writer API path regex [sc-1830]

albertnadal:
### Description
Update the nucliadb_writer API path regex in the Virtual Service config file. Previous regex was too much generic, it causes some collisions with similar endpoints defined in other services (for instance, supervisor_ml). Current regex is more specific to nucliadb_writer.

### How was this PR tested?

### DIFF
--- a/charts/nucliadb_writer/templates/writer.vs.yaml
+++ b/charts/nucliadb_writer/templates/writer.vs.yaml
@@ -15,7 +15,7 @@ spec:
       method:
         regex: 'PUT|DELETE|POST|PATCH|OPTIONS'
     - uri:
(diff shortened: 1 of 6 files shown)


bloodbare (commented):
Why we need to change the creation of a KB. It should not be exposed on istio

bloodbare (approved):
After talking to @sunbit I agree with the aproach

albertnadal:
> Why we need to change the creation of a KB. It should not be exposed on istio

I just seen that the path defined for creating a KB is not consistent with the REST naming guidelines. When adding or listing the contents of a collection we usually use the plural name of the path segment not the singular.

PR #176, opened 2022-07-06
Allow OPTIONS to search endpoint

albertnadal:
### Description

### How was this PR tested?


bloodbare (commented):
Is it normal that some places is 'GET|OPTIONS' and some other GET|OPTIONS without quotes ?

albertnadal:
> Is it normal that some places is 'GET|OPTIONS' and some other GET|OPTIONS without quotes ?

It has no effect in the YAML. Those are needed in case you use special chars. To be consistent let us put the quotation marks.

bloodbare (approved):

PR #198, opened 2022-07-15
Fixing paragraphs

hermeGarcia:
### Description
Distance was changed after parsing, not after searching.

### How was this PR tested?
Local tests.


r3bu1ld3r (approved):
lgtm

PR #232, opened 2022-08-30
Sc 1793  nucliadb metrics bytes fields and searches

sunbit:
- Extend audit to provide insight of how many fields have changed and it's file fields storage movements
- Fix Audit of new resources was being set to MODIFIED
- Fix storage error when overwriting a file field
- Fix delete of a field when it has nested data


nucliadb_ingest/nucliadb_ingest/orm/processor.py
```diff
@@ -442,7 +589,7 @@ async def create_kb(
             raise e
 
         if not failed:
-            storage = await get_storage(service_name=SERVICE_NAME)
+            storage = await get_storage()
```
bloodbare, line 592: Why we remove service_name ?

nucliadb_ingest/nucliadb_ingest/tests/fixtures.py
```diff
@@ -274,14 +284,17 @@ async def txn(redis_driver):
 
 @pytest.fixture(scope="function")
 async def cache(redis):
+
     url = f"redis://{redis[0]}:{redis[1]}"
     pubsub = RedisPubsub(url)
+    set_utility(Utility.PUBSUB, pubsub)
```
bloodbare, line 290: Could we check if already exist ?

nucliadb_utils/nucliadb_utils/storages/gcs.py
```diff
@@ -130,7 +130,7 @@ async def copy(
                 )
             else:
                 data = await resp.json()
-                assert data["name"] == destination_uri
+                assert data["resource"]["name"] == destination_uri
```
bloodbare, line 133: Are you sure?

bloodbare (changes requested):

nucliadb_ingest/nucliadb_ingest/fields/base.py
```diff
@@ -122,11 +122,15 @@ async def db_set_value(self, payload: Any):
...
-        await self.resource.txn.delete(
-            KB_RESOURCE_FIELD.format(
-                kbid=self.kbid, uuid=self.uuid, type=self.type, field=self.id
-            )
+        field_base_key = KB_RESOURCE_FIELD.format(
+            kbid=self.kbid, uuid=self.uuid, type=self.type, field=self.id
         )
+        # Make sure we explicitly delete the field and any nested key
+        keys_to_delete = []
+        async for key in self.resource.txn.keys(field_base_key):
```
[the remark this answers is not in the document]
sunbit: I'll check, but the first result of the `.keys` includes the provided key, so a single key`.keys` call will emit the same key.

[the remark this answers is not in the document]
bloodbare: just make sure it works with the three drivers

bloodbare (approved):

PR #244, opened 2022-09-13
Add X-SYNCHRONOUS header to all writer endpoints

lferran:
### Description
Includes the X-SYNCHRONOUS header on all writer endpoints

### How was this PR tested?
Corresponding integration tests have been added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ target
 .eggs
 .egg-info
 *.so
+venv/
 
 # Coverage
 .coverage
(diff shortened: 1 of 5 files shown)


jotare (commented):
I think it would be nice to add X-Synchronous also to knowledgebox requests (nucliadb_writer/api/v1/knowledgebox.py)

.gitignore
```diff
@@ -26,6 +26,7 @@ target
 .eggs
 .egg-info
 *.so
+venv/
```
bloodbare, line 29: We are using mostly pyenv virtualenvs just as a comment

bloodbare (approved):

.gitignore
lferran: Ah nice! I am already using pyenv but I didn't know this feature 👍🏼 
Still, it doesn't harm to leave `venv` here I think

PR #245, opened 2022-09-14
Add X-SYNCHRONOUS header to regular upload endpoint

lferran:
### Description
Added the header to the upload endpoint

### How was this PR tested?
With an integration test

### DIFF
--- a/nucliadb_writer/nucliadb_writer/tests/test_files.py
+++ b/nucliadb_writer/nucliadb_writer/tests/test_files.py
@@ -411,3 +411,47 @@ async def test_knowledgebox_file_upload_field_headers(
         key=writer.files[field].file.uri,
     )
     assert len(data.read()) == 30472
+
+
+@pytest.mark.asyncio
+async def test_knowledgebox_file_upload_field_sync(
+    writer_api, knowledgebox_writer, resource
+):
+    async with writer_api(roles=[NucliaDBRoles.WRITER]) as client:
+
+        filename = base64.b64encode(b"image.jpg").decode()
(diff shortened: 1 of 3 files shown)


nucliadb_writer/nucliadb_writer/tests/test_files.py
```diff
@@ -411,3 +411,47 @@ async def test_knowledgebox_file_upload_field_headers(
...
+    assert payload["kbid"] == knowledgebox_writer
+    path = resp.headers["ndb-field"]
+    field = path.split("/")[-1]
+
+    storage = await get_storage()
+    data = await storage.downloadbytes(
+        bucket=writer.files[field].file.bucket_name,
+        key=writer.files[field].file.uri,
+    )
+    assert len(data.read()) == 30472
```
lferran, line 457: @nuclia/nuclia-db here I'm not sure how to properly assert that everything went OK.
I pretty much copied from the test above, but the fact that we are making the writer ack-ing here in the test confuses me, as I was expecting it to happen as a result of adding the X-SYNCHRONOUS header.

bloodbare: Here what we check is that writer sended the proper broker message to ingest, with the information that is required. Then checks that writer wrote the file on the proper place. This test checks that writer does what is supossed to do, it does not check that ingest transactioned it. In order to check it you can call the ingest utility and check if the resource exist, or do the test at nucliadb_one level were you can call the reader on the same test and check that the resource is properly set on the DB after the writer call.

bloodbare (approved):

PR #264, opened 2022-09-20
Fix pagination

lferran:
### Description
Pagination was broken before.

### How was this PR tested?
An integration test was added


nucliadb_search/nucliadb_search/search/merge.py
```diff
@@ -274,8 +273,8 @@ async def merge_paragraph_results(
         for result in paragraph_response.results:
             raw_paragraph_list.append(result)
 
-    if len(raw_paragraph_list) > 1:
```
lferran, line 277: The actual bug was here: We were always sorting the list of paragraphs. Also results were sorted on increasing score, which made subsequent pages be duplicates of the first page.

bloodbare (approved):

PR #266, opened 2022-09-21
Added  boolean `custom` field on EntityGroups

lferran:
### Description
Adds a new field on `EntityGroup` object (both in pydantic model and in protobuf) so that the API team can distinguish entities created by the user (`custom=True`) from those that are generated by the processing pipeline

### How was this PR tested?
Adding an integration test


bloodbare (approved):
Reader sideeffect is not reflected on this code. It should work but adding a test on nucliadb may be a good idea to test read/write behavior.

nucliadb_protos/knowledgebox.proto
```diff
@@ -124,6 +124,7 @@ message EntitiesGroup {
     map<string, Entity> entities = 1;
     string title = 2;
     string color = 3;
+    bool custom = 4;
```
lferran, line 127: According to https://developers.google.com/protocol-buffers/docs/proto#other , bools on proto2 have a default value of `False`, so we should be fine with already existing entities.

PR #267, opened 2022-09-21
Adding fieldmetadata on resources

jotare:
### Description
- GET a resource will now return the field `fieldmetadata`.
- fieldmetadata now is able to overwrite content for the same field
- `TokenSplit` now has `start` and `end` fields to specify token boundaries

### How was this tested?
- With integration and e2e backend tests


bloodbare (changes requested):
Fields are defined not only by the field itself , also by the field_type. There is the funcion `self.generate_field_id` that generates a string with the fieldId.

bloodbare (approved):

nucliadb_one/nucliadb_one/tests/test_fieldmetadata.py
```diff
@@ -0,0 +1,153 @@
...
+            f"/{KB_PREFIX}/{knowledgebox_one}/resource/{rid}?show=basic&show=extracted",
+        )
+        assert resp.status_code == 200
+        fieldmetadata = resp.json()["fieldmetadata"]
+
+        assert len(fieldmetadata) == 2
+        assert fieldmetadata[0] == fieldmetadata_2
+        assert fieldmetadata[1]["field"] == fieldmetadata_1["field"]
+        assert fieldmetadata[1]["paragraphs"] == fieldmetadata_1["paragraphs"]
+        assert fieldmetadata[1]["token"] == []
```
lferran, line 153: What about adding some dummy tokens on the test too?
This way we make sure that the `start` and `end` fields that we added are covered by a test.

nucliadb_protos/resources.proto
```diff
@@ -291,6 +291,8 @@ message UserMetadata {
 message TokenSplit {
     string token = 1;
     string klass = 2;
+    uint32 start = 3;
+    uint32 end = 4;
```
lferran, line 295: What will happen with already existing tokens that do not have `start` and `end` values set?

nucliadb_ingest/nucliadb_ingest/orm/resource.py
```diff
@@ -166,9 +166,28 @@ async def set_basic(self, payload: PBBasic, slug: Optional[str] = None):
...
         if self.basic is not None and self.basic != payload:
             self.basic.MergeFrom(payload)
+
             # We force the usermetadata classification to be the one defined
             if payload.HasField("usermetadata"):
                 self.basic.usermetadata.CopyFrom(payload.usermetadata)
+
+            if len(payload.fieldmetadata):
+                # keep only the last fieldmetadata item for each field. API
+                # users are responsible to manage fieldmetadata properly
```
lferran, line 176: WDYT about adding a test with this case (modify the same field twice on the same payload)? Just to be even more explicit.

nucliadb_one/nucliadb_one/tests/test_fieldmetadata.py
jotare: `start` and `end` are implicitly tested on the asserts like `assert fieldmetadata[0] == fieldmetadata_2`. But I can add a more explicit one if you think so

nucliadb_ingest/nucliadb_ingest/orm/resource.py
```diff
@@ -166,9 +166,28 @@ async def set_basic(self, payload: PBBasic, slug: Optional[str] = None):
...
+                    if field_id not in fields:
+                        fields.append(field_id)
+                    positions[field_id] = i
+
+                updated = [
+                    self.basic.fieldmetadata[positions[field]] for field in fields
+                ]
+
+                del self.basic.fieldmetadata[:]
+                self.basic.fieldmetadata.extend(updated)
```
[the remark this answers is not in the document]
jotare: That's because assignment is now allowed to a repeated field like this. You need to use methods for repeated fields (like `add`, `append` or `extend`). So I need to clear it first and then extend with the updated list

nucliadb_protos/resources.proto
lferran: seems like proto2 will assign a default value of 0

lferran (approved):

PR #268, opened 2022-09-21
fix: [sc-2390] PATCH thumbnail property on resource doesn't work

jotare:
Story details: https://app.shortcut.com/flaps/story/2390

### DIFF
--- a/nucliadb_one/nucliadb_one/tests/test_basic.py
+++ b/nucliadb_one/nucliadb_one/tests/test_basic.py
@@ -1,0 +1,71 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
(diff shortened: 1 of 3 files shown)


nucliadb_one/nucliadb_one/tests/test_basic.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pytest
+from httpx import AsyncClient
+
+from nucliadb_models.resource import NucliaDBRoles
+from nucliadb_search.api.v1.router import KB_PREFIX
+
+
+@pytest.mark.asyncio
+async def test_basic_path_thumbnail_sc_2390(
```
lferran, line 30: typo `path` should be `patch`

lferran (approved):
lgtm! 👍🏼

nucliadb_one/nucliadb_one/tests/test_basic.py
jotare: Fixed

PR #270, opened 2022-09-21
Fixing Export

bloodbare:
### Description

### How was this PR tested?


jotare (approved):

hermeGarcia (approved):

nucliadb_ingest/nucliadb_ingest/orm/resource.py
```diff
@@ -244,7 +244,7 @@ async def generate_index_message(self) -> ResourceBrain:
                     brain.apply_field_vectors(field_key, vo, [], {})
         return brain
 
-    async def generate_broker_message(self) -> BrokerMessage:
+    async def generate_broker_message(self, cf: bool = True) -> BrokerMessage:
```
lferran, line 247: Does `cf` mean `cloud_file`? Why not naming the parameter `cloud_file` or to something more meaningful?

nucliadb_ingest/nucliadb_ingest/orm/resource.py
```diff
@@ -263,27 +263,70 @@ async def generate_broker_message(self) -> BrokerMessage:
...
+                bm.conversations[field_id].CopyFrom(value)
+            elif type_id == FieldType.KEYWORDSET:
+                value = await field.get_value()
+                bm.keywordsets[field_id].CopyFrom(value)
+            elif type_id == FieldType.DATETIME:
+                value = await field.get_value()
+                bm.datetimes[field_id].CopyFrom(value)
+            elif type_id == FieldType.LAYOUT:
+                value = await field.get_value()
+                bm.layouts[field_id].CopyFrom(value)
```
lferran, line 287: What if the type is non of these? Do we need a `else: raise ValueError()` branch at the end?

lferran (approved):
I left a couple of comments, but LGTM!

nucliadb_ingest/nucliadb_ingest/orm/resource.py
bloodbare: changing to `full` in order to make it more clear, it means all the bm, not the pointers

nucliadb_ingest/nucliadb_ingest/orm/resource.py
bloodbare: in case there is else we do not want to raise, we do not want to export.

PR #272, opened 2022-09-23
Store the filename metadata properly on gcloud

lferran:
### Description
This bug was preventing downloads from getting the correct `Content-Disposition` header.

### How was this PR tested?
Added an e2e test that uploads a file and downloads it later.

### DIFF
--- a/nucliadb_writer/nucliadb_writer/tus/gcs.py
+++ b/nucliadb_writer/nucliadb_writer/tus/gcs.py
@@ -194,7 +194,8 @@ class GCloudFileStorageManager(FileStorageManager):
             filename = "file"
         else:
             filename = dm.filename
-        metadata = json.dumps({"NAME": filename})
+        metadata = json.dumps({"metadata": {"FILENAME": filename}})
+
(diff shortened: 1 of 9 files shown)


nucliadb_writer/nucliadb_writer/tus/gcs.py
```diff
@@ -194,7 +194,8 @@ async def start(self, dm: FileDataMangaer, path: str, kbid: str):
             filename = "file"
         else:
             filename = dm.filename
-        metadata = json.dumps({"NAME": filename})
+        metadata = json.dumps({"metadata": {"FILENAME": filename}})
```
lferran, line 197: Following google documentation: https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations

Custom metadata needs to be nested under `metadata` key. 
As of the `name` key, we don't really need it, as it is already being set at the `init_url` on line 189

bloodbare: Great catch!

PR #275, opened 2022-09-27
File field with externally hosted file

lferran:
### Description
These changes provide a way for api users to create a file field that will point to a url which is not in Nuclia's domain.

### How was this PR tested?
- [x] e2e tests on nucliadb_one
- [x] integration tests at ingest


nucliadb_ingest/nucliadb_ingest/processing.py
```diff
@@ -216,6 +216,28 @@ async def convert_filefield_to_str(self, file: models.FileField) -> str:
...
+            "aud": "urn:proxy",
+            "iat": now,
+            "exp": expiration,
+            "jti": uuid.uuid4().hex,
+            "source": 1,  # To indicate that this files comes internally
+            "driver": models.CloudFile.Source.EXTERNAL.value,
+            "uri": file_field.uri,
+            "extra_headers": file_field.extra_headers,
+        }
+        return jwt.encode(payload, self.nuclia_jwt_key, algorithm="HS256")
```
lferran, line 239: @sunbit FYI: This is what is inside of the jwt that I'll be sending along on the `PushPayload` to the processing-slow

nucliadb_models/nucliadb_models/file.py
```diff
@@ -60,6 +65,11 @@ class FileField(BaseModel):
     file: FileB64
 
 
+class ExternalFileField(BaseModel):
+    uri: str
+    extra_headers: Dict[str, str] = {}
```
jotare, line 70: I think this field should be `Optional`, public links may not need extra headers

nucliadb_writer/nucliadb_writer/api/v1/field.py
```diff
@@ -293,16 +294,34 @@ async def add_resource_field_conversation(
 )
 @requires(NucliaDBRoles.WRITER)
 @version(1)
-async def add_resource_field_file(
+async def _add_resource_field_file(
```
jotare, line 297: Using underscore to avoid name collision seems a bit ugly to me. Maybe we can rename the other function `add_resource_field_file_encoded` or something like that and keep this without the underscore?

jotare (changes requested):

nucliadb_models/nucliadb_models/file.py
lferran: yup! The fact that it has a default value makes it "optional" in terms of API: if the user doesn't set it, it won't break.
However, adding the `Optional` pydantic type means that it can be `None` too, which I don't want to.

jotare: Yes, my mistake

nucliadb_ingest/nucliadb_ingest/processing.py
```diff
@@ -216,6 +216,28 @@ async def convert_filefield_to_str(self, file: models.FileField) -> str:
...
+        expiration = now + datetime.timedelta(days=self.days_to_keep)
+        payload = {
+            "iss": "urn:nucliadb",
+            "sub": "file",
+            "aud": "urn:proxy",
+            "iat": now,
+            "exp": expiration,
+            "jti": uuid.uuid4().hex,
+            "source": 1,  # To indicate that this files comes internally
+            "driver": models.CloudFile.Source.EXTERNAL.value,
```
sunbit, line 235: If i read this correctly, this would resolve to "EXTERNAL", where in the other token payload self.driver is an `int`. This values needs to match the stashify_protos.processing.Driver protobuffer, that is not a dependency of nucliadb, but its is in terms of what this values needs to be.

I think there is a bit of naming problem/overlapping between the concept of source and driver that makes this a little diffucult to follow sometimes.

sunbit: BTW, the value needs to be `3`

jotare (approved):

sunbit (approved):

PR #276, opened 2022-09-27
Store processing sequence id on Basic protobuf

jotare:


nucliadb_writer/nucliadb_writer/resource/basic.py
```diff
@@ -213,3 +213,7 @@ def set_status(basic: Basic, item: CreateResourcePayload):
 
 def set_status_modify(basic: Basic, item: UpdateResourcePayload):
     basic.metadata.status = Metadata.Status.PENDING
+
+
+def set_seqid(bm: BrokerMessage, seqid: int):
```
lferran, line 218: to be more accurate: `set_last_seqid`?

lferran (approved):
lgtm! 👍🏼

nucliadb_writer/nucliadb_writer/resource/basic.py
jotare: Changed

lferran (commented):
I think you are still missing some cases, like when updating a field text, a field link, and others.

PR #277, opened 2022-09-27
github actions | disable arm64 builds

ferpizza:
### Description
Disable arm64 builds on nucliadb docker images, as it's taking ours to compile.
We will re-enable when a better solution is found.


bloodbare:
ALready fixed thanks !

PR #279, opened 2022-09-28
/upload returns a JSON response instead of only headers

jotare:


nucliadb_models/nucliadb_models/writer.py
```diff
@@ -112,3 +112,9 @@ class ResourceDeleted(BaseModel):
 
 
 ComminResourcePayload = Union[CreateResourcePayload, UpdateResourcePayload]
+
+
+class ResourceFileUploaded(BaseModel):
+    seqid: Optional[int] = None
+    rid: Optional[str] = None
```
lferran, line 119: Some questions that come to my mind:

- Should this be `uuid` instead, to be "coherent" with `ResourceCreated` model?
- Should it be stored in `nucliadb_models`. What policy are we following?

nucliadb_writer/nucliadb_writer/api/v1/upload.py
```diff
@@ -537,13 +538,15 @@ async def generate_buffer(storage_manager, request):
     except LimitsExceededError as exc:
         raise HTTPException(status_code=402, detail=str(exc))
 
-    headers = {}
-    headers["NDB-Seq"] = f"{seqid}"
-    headers["NDB-Resource"] = f"/{KB_PREFIX}/{kbid}/resources/{rid}"
-    headers["NDB-Field"] = f"/{KB_PREFIX}/{kbid}/resources/{rid}/field/{valid_field}"
+    response.headers["NDB-Seq"] = f"{seqid}"
+    response.headers["NDB-Resource"] = f"/{KB_PREFIX}/{kbid}/resources/{rid}"
```
lferran, line 542: What about changing `resources` for the `RESOURCE_PREFIX` constant instead?

lferran (approved):
Some NIT comments, but LGTM! 👍🏼

nucliadb_writer/nucliadb_writer/api/v1/upload.py
jotare: This headers will be removed, I'm uploading a new version

nucliadb_models/nucliadb_models/writer.py
jotare: Renamed `rid` to `uuid`

PR #289, opened 2022-10-04
Pb pip

bloodbare:
### Description

### How was this PR tested?


nucliadb_models/requirements.txt
```diff
@@ -1,2 +1,4 @@
 pydantic==1.9.0
-protobuf
\ No newline at end of file
+protobuf
+nucliadb-ingest>=1.0.0
+nucliadb-protos>=0.0.29
```
lferran, line 4: It feels wrong that nucliadb_models is depending on nucliadb_ingest and nucliadb_protos.

lferran: ah I see, it's because of the serialization logic

lferran: This should remove the models -> ingest dependency:

https://github.com/nuclia/nucliadb/pull/291

nucliadb/requirements.txt
```diff
@@ -1,2 +1,4 @@
 uvicorn
-pydantic_argparse
\ No newline at end of file
+pydantic_argparse
+nucliadb-one>=1.0.0
```
jotare, line 3: nucliadb has dependency with nucliadb_writer. Now is solved with nucliadb_one, but we want to be more explicit?

jotare (changes requested):

PR #290, opened 2022-10-05
[nucliadb_reader] Support accessing by slug on resource endpoints

lferran:
### Description
This PR aims to provide API support to do READ operations on resources by the resource slug, instead of the resource uuid.

List of changes:

 - Renamed the `/resource/{rid}` url prefix to `/rid/{rid}`.
 - The previous `resource` prefix is now used to access by slug: `/resource/{rslug}`.
 - When slug is not set, it defaults to uuid. Therefore old code still using `/resource/ruuid` will work.

Endpoints changed:
- reader: 
  - [x] get_resource 
  - [x] get_resource_field
  - [ ] download_extract_field
  - [ ] download_field_file
  - [ ] download_field_layout
  - [ ] download_field_conversation
 
Won't do: writer and search changes. Will come in other PRs.
  
Note: these are breaking changes. Should we notify it somehow?

### How was this PR tested?
Added new integration tests to check access by slug.
Existing tests should ensure we don't break anything.


nucliadb_models/nucliadb_models/serialize.py
```diff
@@ -361,5 +389,4 @@ async def serialize(
                         field_type_name,
                         extracted,
                     )
-    await txn.abort()
```
lferran, line 364: note: now we close the transaction much earlier (inside `get_orm_resource`)

PR #302, opened 2022-10-11
External File Fields v2

lferran:
### Description
- Instead of a different model, we reuse the FileField model now
- We are able to create a resource with an external file field on the post request

### How was this PR tested?
existing tests and new ones

### DIFF
--- a/nucliadb/nucliadb/writer/resource/field.py
+++ b/nucliadb/nucliadb/writer/resource/field.py
@@ -210,8 +210,8 @@ async def parse_internal_file_field(
     writer.files[key].file.CopyFrom(
         await storage.upload_b64file_to_cloudfile(
             sf,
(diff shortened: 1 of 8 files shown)


nucliadb/nucliadb/writer/resource/field.py
```diff
@@ -413,3 +419,7 @@ async def parse_conversation_field(
         convs.messages.append(processing_message)
 
     toprocess.conversationfield[key] = convs
+
+
+def is_external_file(field: models.FileField) -> bool:
+    return field.file.uri is not None
```
jotare, line 425: I think this should be part of the model definition (as a property for example). To avoid spreading implementation details on other parts of the code

jotare (changes requested):

jotare (approved):

PR #305, opened 2022-10-11
Writer resource and field endpoints by slug

lferran:
### Description
This PR adds endpoints to the writer so that resources and fields can be modified using the custom resource slug.

### How was this PR tested?
Adding integration tests


nucliadb/nucliadb/writer/api/v1/resource.py
```diff
@@ -317,10 +367,51 @@ async def delete_resource(
...
 @version(1)
-async def reindex_resource(request: Request, kbid: str, rid: str):
+async def reindex_resource(
+    request: Request,
+    kbid: str,
+    rid: Optional[str] = None,
+    rslug: Optional[str] = None,
+):
+
+    rid = await get_rid_from_params_or_raise_error(kbid, rid, rslug)
```
bloodbare, line 377: Just as a comment, when you redefine the rid from Optional[str] to str its better that you create a different var name that type checker knows it may not be empty

bloodbare (approved):

PR #312, opened 2022-10-17
Fix highlight

lferran:
### Description
Improves highlight search results logic

### How was this PR tested?
Adding another case in existing test

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -37,15 +37,6 @@ flag_management:
           target: 20%
         - type: patch
           target: 100%
-    - name: nucliadb
-      paths:
-        - nucliadb/**
-      carryforward: true
-      statuses:
-        - type: project
-          target: 20%
-        - type: patch
-          target: 80%
     - name: train
       paths:
         - nucliadb/nucliadb/train/**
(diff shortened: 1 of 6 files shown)


.github/.codecov.yml
```diff
@@ -37,15 +37,6 @@ flag_management:
...
           target: 100%
-    - name: nucliadb
-      paths:
-        - nucliadb/**
-      carryforward: true
-      statuses:
-        - type: project
-          target: 20%
-        - type: patch
-          target: 80%
```
jotare, line 48: Why have you deleted it? Shouldn't it be `nucliadb/nucliadb/**`?

nucliadb/nucliadb/writer/tests/test_resources.py
```diff
@@ -117,6 +117,8 @@ async def test_resource_crud(writer_api, knowledgebox_writer):
             },
         )
 
+        if resp.status_code != 201:
+            print(resp.content)
```
jotare, line 121: Have you forgotten to delete this?

jotare (changes requested):

.github/.codecov.yml
lferran: Because "general rules" is managed by the "default_rules" key from a bit above. 
Essentially, this was making most of PRs show with the red cross

jotare (approved):

PR #324, opened 2022-10-21
Use latest uvicorn version

lferran:
### Description
Use latest uvicorn version

### How was this PR tested?
Existing tests should cover it, and I started the reader and the writer servers locally


bloodbare (changes requested):
removing debug is bot replaced for some other parameter on uvicorn?

lferran:
> removing debug is bot replaced for some other parameter on uvicorn?

(sorry, I should have included this as part of the description)

no, it is not replaced by anything. According to https://github.com/encode/uvicorn/pull/1640, the parameter wasn't really used... and even less when you are using with starlette

bloodbare (approved):

PR #331, opened 2022-10-26
splitting fuzzy from search

hermeGarcia:
### Description

### How was this PR tested?


hermeGarcia:
> @hermeGarcia is the `booster` score field not used anymore? can we get rid of it?

It is still being used. Fuzzy is still being done, but only when no results are  found.

PR #338, opened 2022-11-02
Forbid use of slash (/) in field ids

jotare:
- NucliaDB API validates field ids and don't allow use of the slash character (`/`) on any model
- Fix slug validation

### DIFF
--- a/nucliadb/nucliadb/models/utils.py
+++ b/nucliadb/nucliadb/models/utils.py
@@ -23,4 +23,4 @@ import pydantic
 
 
 class SlugString(pydantic.ConstrainedStr):
-    regex = re.compile(r"[a-z0-9_-]+")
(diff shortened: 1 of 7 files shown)


hermeGarcia (approved):

nucliadb/nucliadb/models/utils.py
```diff
@@ -22,5 +22,9 @@
 import pydantic
 
 
+class FieldIdString(pydantic.ConstrainedStr):
+    regex = re.compile(r"^[^/]+$")
+
+
 class SlugString(pydantic.ConstrainedStr):
-    regex = re.compile(r"[a-z0-9_-]+")
+    regex = re.compile(r"^[a-z0-9_-]+$")
```
lferran, line 30: WDYT about adding some unit tests that make it clearer what would be some valid and invalid examples of each?

PR #339, opened 2022-11-03
Unit test field ids

jotare:


nucliadb/nucliadb/tests/test_field_ids.py
```diff
@@ -0,0 +1,55 @@
...
+            TestFieldIdModel(field_id=valid)
+        except ValidationError:
+            assert False, f"Field id '{valid}' should be valid!"
+
+    invalid_field_ids = [
+        "",
+        "foo/bar",
+    ]
+    for invalid in invalid_field_ids:
+        try:
```
lferran, line 50: NIT comment: you can use `with pytest.raises` instead

lferran (approved):

PR #341, opened 2022-11-03
Tag vectors search

hermeGarcia:
### Description
[Closest up node ](https://github.com/nuclia/nucliadb/blob/df5deca049f381a24b789e9fa5871ac76593176b/nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs#L114) computes for a given node the closest in the layer (in terms of cosine similarity) that matches certain restrictions, which are:
- The vector is not deleted.
- The vector has the required labels. (new addition).

This process is done only for the layer 0 with the best results being already found.  In this new version we avoid finding a node that is already part of the solution and the closest node is also computed for vectors that dont meet the label requirements.
### How was this PR tested?
Local tests


nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -114,17 +112,25 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+        solution: &mut HashSet<Address>,
+        mut filters: impl Iterator<Item = &'a [u8]>,
         layer: L,
         x: Address,
         y: Address,
     ) -> Option<(Address, f32)> {
+        solution.remove(&x);
         let mut candidates = BinaryHeap::from([Cnx(x, self.cosine_similarity(x, y))]);
-        let mut visited = HashSet::from([x]);
+        let mut visited = HashSet::new();
```
alekece, line 124: Ditto. `visited_nodes` for clarity ?

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -194,8 +194,7 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
         entry_points: &[Address],
     ) -> Vec<Address> {
         use params::*;
-        let s_result =
-            self.layer_search::<&RAMLayer>(x, layer, ef_construction(), NO_FILTER, entry_points);
+        let s_result = self.layer_search::<&RAMLayer>(x, layer, ef_construction(), entry_points);
```
alekece, line 197: Ditto. `search_result` ?

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -114,17 +112,25 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
         self.tracker.consine_similarity(x, y)
     }
     fn closest_up_node<L: Layer>(
-        &self,
+        &'a self,
+        solution: &mut HashSet<Address>,
+        mut filters: impl Iterator<Item = &'a [u8]>,
```
alekece, line 117: I think we could pass a slice here directly (since the `search` method takes a slice too as argument) and then avoid an useless copy at call site.

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -251,24 +250,26 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
             let neighbours = result
                 .neighbours
                 .into_iter()
-                .map(|(n, _)| self.closest_up_node(hnsw.get_layer(0), n, x))
-                .filter(Option::is_some)
-                .flatten()
+                .flat_map(|(n, _)| {
+                    self.closest_up_node(
+                        &mut solution,
+                        with_filter.iter().copied(),
```
alekece, line 267: This copy can be avoided in the `closest_up_node` method takes a slice as argument

alekece (changes requested):

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -114,17 +112,25 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
         self.tracker.consine_similarity(x, y)
     }
     fn closest_up_node<L: Layer>(
-        &self,
+        &'a self,
+        solution: &mut HashSet<Address>,
```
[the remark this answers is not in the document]
hermeGarcia: But what I want to express is that the addresses on the set are already part of the solution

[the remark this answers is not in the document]
alekece: Maybe `retained_nodes` is not a good name as well :cat:

alekece (approved):
Great work !

PR #342, opened 2022-11-04
Be able to search without duplicates

lferran:
### Description
Added query parameters to search endpoints to be able to filter out duplicates (paragraphs and vectors).
The actual filtering happens at the node. This PR just makes sure the wiring is right.

When calculating a brain resource, we detect which paragraphs are duplicated in a field by computing the hash of the text.

### How was this PR tested?
Added integration and unit tests.


nucliadb/nucliadb/ingest/orm/brain.py
```diff
@@ -41,6 +43,18 @@
...
 
+class DuplicateParagraphsChecker:
+    def __init__(self):
+        self.seen = set()
+
+    def check(self, paragraph: Paragraph) -> bool:
+        """Returns whether paragraph has already been checked"""
+        par_md5 = hashlib.md5(paragraph.text.encode()).hexdigest()
+        already_checkd = par_md5 in self.seen
+        self.seen.add(par_md5)
```
bloodbare, line 54: maybe skip if its not inside ?

bloodbare (approved):

PR #347, opened 2022-11-07
clean_and_upgrade function

hermeGarcia:
### Description
Update and clean function enabled for a node. This function will:

- Remove all data from the given shard id
- Create new (and empty) indexes for the shard id
- Returns the versions of the indexes after the update.

### How was this PR tested?
Local tests


lferran (approved):

nucliadb_node/src/writer/grpc_driver.rs
```diff
@@ -379,18 +379,18 @@ mod tests {
...
-        for id in request_ids {
+        for (id, expected) in request_ids.iter().map(|v| (v.clone(), v.clone())) {
             let response = client
-                .delete_shard(Request::new(ShardId { id: id.clone() }))
+                .delete_shard(Request::new(ShardId { id }))
                 .await
                 .expect("Error in delete_shard request");
             let deleted_id = response.get_ref().id.clone();
-            assert_eq!(deleted_id, id);
+            assert_eq!(deleted_id, expected);
```
alekece, line 393: We may use `debug_assert_eq` here

alekece (approved):

PR #350, opened 2022-11-08
Fix Processor transactional error

jotare:
Fix transactional error with TiKV when trying to get a nonexistent KB by slug


lferran:
it would be nice to refactor all the transaction logic to use some kind of context manager that ensures we don't miss this kind of things

lferran (approved):

PR #353, opened 2022-11-08
filtering out duplicate vectors

hermeGarcia:
### Description
This PR opens the possibility for removing duplicates from a vector search (~~based on cosine similarity~~). This is achieved by applying the same logic of label filtering to score redundancy in "closest_up_node".
The proto VectorSearchRequest allows to include if duplicates are wanted or not.
###EDIT
Cosine similarity may cause false duplicates (i.e. different vectors can have the same cosine similarity). For this reason commit [f3ad06b](https://github.com/nuclia/nucliadb/pull/353/commits/f3ad06bd7ca1dbacecfbda5b2fccf34cdc236ac5) uses vector comparison.
### How was this PR tested?
Local tests


nucliadb_vectors2/src/vectors/data_point/mod.rs
```diff
@@ -122,6 +122,14 @@ impl<'a, Dlog: DeleteLog> Retriever<'a, Dlog> {
...
 
 impl<'a, Dlog: DeleteLog> DataRetriever for Retriever<'a, Dlog> {
+    fn get_vector(&self, x @ Address(addr): Address) -> &[u8] {
+        if addr == key_value::get_no_elems(self.nodes) {
+            self.temp
+        } else {
+            let x = self.find_node(x);
+            Node::vector(x)
+        }
+    }
```
alekece, line 132: Maybe we can remove the destructing pattern `x @ Address(addr)` parameter and implement on `Address` type a `inner` method to improve maintainability and readability ?
Because, even if the destructing pattern is super useful time to time, this is rarely the case in function parameter (the code will break in a instant if we modify a bit the `Address` type).

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -110,29 +111,29 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
         &'a self,
-        solution: &mut HashSet<Address>,
-        filters: &[&[u8]],
-        layer: L,
         x: Address,
-        y: Address,
+        query: Address,
+        layer: L,
+        filters: &[&[u8]],
+        sol_addresses: &HashSet<Address>,
```
alekece, line 118: I would rather prefer `solution_addresses` (`sol` can stand for an acronyme)

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -110,29 +111,29 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
         loop {
             match candidates.pop() {
                 None => break None,
-                Some(Cnx(n, _))
+                Some(Cnx(n, score))
                     if !self.tracker.is_deleted(n)
-                        && !solution.contains(&n)
+                        && vec_counter.get(score) == 0
+                        && !sol_addresses.contains(&n)
                         && filters.iter().all(|label| self.tracker.has_label(n, label)) =>
```
alekece, line 130: This predicate is a bit hard to understand at first glance. Can you add some comments or extract it in a proper function ?

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
```diff
@@ -232,32 +233,78 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+            let mut vec_counter = RepCounter::new(!with_duplicates);
+            let mut filtered_result = Vec::new();
+            result.iter().copied().for_each(|(addr, score)| {
+                sol_addresses.insert(addr);
+                vec_counter.add(score);
+            });
+            result.into_iter().for_each(|(addr, score)| {
+                sol_addresses.remove(&addr);
+                vec_counter.sub(score);
+                if let Some((adr, score)) = self.closest_up_node(
```
alekece, line 265: I think you can reuse `addr` variable name here (and shadow the previous one) instead of having a variable names way too similar

alekece (changes requested):

nucliadb_vectors2/src/vectors/data_point/mod.rs
hermeGarcia: Address is a internal type that must never be exposed, as well as the methods that use such type. For this reason I dont want to implement a inner method nor Deref trait, if Address changes, the code that uses it should complain and carefully modified. 
The same thing is taken in the Rust compiler when dealing with the MIR, matching variants of the IR must be  exhaustive for safety reasons.

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: Will do

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: Will do

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: Its a typo, will do

nucliadb_vectors2/src/vectors/data_point/mod.rs
alekece: Ok, sounds fair (even if I truly believe that internal types should be treated as exposed one)

hermeGarcia: Lets revisit this when I get back, maybe there is an approach that gives both flexibility and security.

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: fixed by [363880a](https://github.com/nuclia/nucliadb/pull/353/commits/363880a41a5053d81c5f3b42f5abbc68d8f4bd92)

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: fixed by [363880a](https://github.com/nuclia/nucliadb/pull/353/commits/363880a41a5053d81c5f3b42f5abbc68d8f4bd92)

nucliadb_vectors2/src/vectors/data_point/ops_hnsw.rs
hermeGarcia: fixed by [363880a](https://github.com/nuclia/nucliadb/pull/353/commits/363880a41a5053d81c5f3b42f5abbc68d8f4bd92)

alekece (approved):

PR #357, opened 2022-11-10
Add some contexts on LMDB opening failure

alekece:
### Description
This PR aims to replace the `unwrap` calls which cause a panic after a  node cleaning.

### How was this PR tested?
Tests are green.


hermeGarcia:
Fixed in [this PR](https://github.com/nuclia/nucliadb/pull/365).

PR #365, opened 2022-11-14
using create_database when opening

hermeGarcia:
### Description
When the clean function appeared a  new scenario appeared for the index. A shard that is not yet created (this is done by the query) may be queried by the reader. In the old version of relations this causes a lot of problems because I didn't knew that "create_database" in heed opens or creates a db. In the new version this is fixed.

### How was this PR tested?
Local test

### DIFF
--- a/nucliadb_relations/src/storage_system/mod.rs
+++ b/nucliadb_relations/src/storage_system/mod.rs
@@ -137,13 +137,23 @@ impl StorageSystem {
         unsafe {
             env_builder.flag(Flags::MdbNoLock);
         }
-        let env = env_builder.open(&env_path).unwrap();
-        let keys = env.create_database(Some(db_name::KEYS)).unwrap();
-        let inv_keys = env.create_database(Some(db_name::INVERSE_KEYS)).unwrap();
-        let edges = env.create_database(Some(db_name::EDGES)).unwrap();
-        let in_edges = env.create_database(Some(db_name::INVERSE_EDGES)).unwrap();
-        let state = env.create_database(Some(db_name::STATE)).unwrap();
-        std::fs::File::create(path.join(env_config::STAMP)).unwrap();
+        let env = env_builder.open(&env_path).expect("Opening env failed");
+        let keys = env
+            .create_database(Some(db_name::KEYS))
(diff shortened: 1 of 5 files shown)


nucliadb_relations/src/storage_system/mod.rs
```diff
@@ -316,7 +272,7 @@ mod tests {
     fn open_and_create() {
         let dir = tempfile::tempdir().unwrap();
         StorageSystem::create(dir.path());
-        StorageSystem::open(dir.path());
+        StorageSystem::create(dir.path());
```
alekece, line 275: You can remove this line :cat:

alekece: Ok, sounds fair. Please add a comment to avoid to remove that line later on (because I am sure I will forget about it and feel tempted to remove it in a future iteration :cat:)

alekece: Plus, can you remove the unnecessary `assert!(true)` just after ?

nucliadb_relations/src/storage_system/mod.rs
```diff
@@ -128,66 +128,22 @@ pub struct StorageSystem {
...
-            .open_database(Some(db_name::INVERSE_EDGES))
-            .unwrap()
-            .unwrap();
-        let state = env.open_database(Some(db_name::STATE)).unwrap().unwrap();
+        let keys = env.create_database(Some(db_name::KEYS)).unwrap();
+        let inv_keys = env.create_database(Some(db_name::INVERSE_KEYS)).unwrap();
+        let edges = env.create_database(Some(db_name::EDGES)).unwrap();
+        let in_edges = env.create_database(Some(db_name::INVERSE_EDGES)).unwrap();
+        let state = env.create_database(Some(db_name::STATE)).unwrap();
+        std::fs::File::create(path.join(env_config::STAMP)).unwrap();
```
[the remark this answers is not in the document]
hermeGarcia: Solved by [538319c](https://github.com/nuclia/nucliadb/pull/365/commits/538319c139d18ded2eda443609f468b807edb50f)

lferran (approved):

PR #371, opened 2022-11-15
filtering invalid facets during indexing

hermeGarcia:
### Description
Invalid facets while indexing will cause the node to panic, now they will be ignored and the error reported.

### How was this PR tested?
Local tests 
Added nucliadb - node e2e test in nucliadb/tests


bloodbare (changes requested):
The error is on the data not to avoid the problem

lferran (approved):

PR #375, opened 2022-11-16
node parallel search

hermeGarcia:
### Description
The locking mechanism used for updating the shard cache should not block during the full search.

### How was this PR tested?
Local test

### DIFF
--- a/nucliadb_node/binding/src/lib.rs
+++ b/nucliadb_node/binding/src/lib.rs
@@ -59,6 +59,7 @@ impl NodeReader {
 
     pub fn get_shard<'p>(&mut self, shard_id: RawProtos, py: Python<'p>) -> PyResult<&'p PyAny> {
         let shard_id = ShardId::decode(&mut Cursor::new(shard_id)).unwrap();
+        self.reader.load_shard(&shard_id);
         if let Some(shard) = self.reader.get_shard(&shard_id) {
             let stats = shard.get_info();
             let shard_pb = ShardPB {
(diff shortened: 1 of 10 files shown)


nucliadb_node/binding/src/lib.rs
```diff
@@ -59,6 +59,7 @@ impl NodeReader {
 
     pub fn get_shard<'p>(&mut self, shard_id: RawProtos, py: Python<'p>) -> PyResult<&'p PyAny> {
         let shard_id = ShardId::decode(&mut Cursor::new(shard_id)).unwrap();
+        self.reader.load_shard(&shard_id);
```
alekece, line 62: I wonder if it is not more ergonomic to nest the `self.reader.load_shard(&shard_id)` call in methods that take a `shard_id` as parameter (such as `search`, `suggest` and so on) because it looks like error-prone and easy to forget that `load_shard` has to be called before-hand.

nucliadb_node/src/reader/grpc_driver.rs
```diff
@@ -35,7 +36,14 @@ impl From<NodeReaderService> for NodeReaderGRPCDriver {
         NodeReaderGRPCDriver(RwLock::new(node))
     }
 }
-
+impl NodeReaderGRPCDriver {
+    async fn shard_loading(&self, id: &ShardId) {
```
alekece, line 40: Can you document this method please ?

nucliadb_node/binding/src/lib.rs
alekece: *UPDATE*: I just saw the other method named `shard_loading`.  You can ignore my comment.

alekece (approved):
Great job !

nucliadb_node/src/reader/grpc_driver.rs
hermeGarcia: Sure, done at [cf87f88](https://github.com/nuclia/nucliadb/pull/375/commits/cf87f88b2ec66bbd265faaff8a884e8702b9d80d).

PR #386, opened 2022-11-21
Display ML-assigned labels in resources list

lferran:
### Description
This PR makes sure the classification metadata created while processing is display on resources list endpoint.

A test is added for search results too, although no new functionality is needed to display them. Only the `&show=extracted` param is required on the search request.

### How was this PR tested?
A couple of backend functional tests were added at nucliadb/tests, one for each endpoint.


nucliadb/nucliadb/reader/api/v1/resource.py
```diff
@@ -80,11 +80,10 @@ async def list_resources(
     driver = await get_driver()
     txn = await driver.begin()
 
-    # Filter parameters for serializer. Those won't be exposed in the API, at least for now
-    # As we want to restrict the exported information to only BASIC
-    show = [ResourceProperties.BASIC]
-    field_types: List[FieldTypeName] = []
-    extracted: List[ExtractedDataTypeName] = []
+    # Filter parameters for serializer
+    show = [ResourceProperties.BASIC, ResourceProperties.EXTRACTED]
```
alekece, line 84: Maybe you can precise the type of the `show` variable as you did for the others ?

alekece (changes requested):

nucliadb/nucliadb/tests/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    etw.field.field = "title"
+    etw.field.field_type = rpb.FieldType.GENERIC
+    bm.extracted_text.append(etw)
+    bm.source = BrokerMessage.MessageSource.WRITER
+    return bm
+
+
+async def inject_message(writer: WriterStub, message):
+    await writer.ProcessMessage([message])  # type: ignore
+    return
```
jotare, line 63: For consistency with other tests, remove duplicate functions in other tests (labels, reindex and search) and use the same everywhere

nucliadb/nucliadb/tests/test_resources.py
```diff
@@ -98,3 +100,26 @@ async def test_resource_crud(
 
     resp = await nucliadb_reader.get(f"/{KB_PREFIX}/{knowledgebox}/resource/{rid}")
     assert resp.status_code == 404
+
+
+@pytest.mark.asyncio
+async def test_list_resources_serializes_labels(
+    nucliadb_reader: AsyncClient,
+    nucliadb_writer: AsyncClient,
+    nucliadb_grpc,
```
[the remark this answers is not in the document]
lferran: yup indeed. However, these are from the python fixtures (defined with a much global scope, to be reusable in different sub-components). I think it should stay like this

nucliadb/nucliadb/tests/utils.py
lferran: yup! good moment to remove duplications 👍🏼

alekece (approved):

PR #399, opened 2022-11-25
Vectors2 delete update

hermeGarcia:
### Description
Delete logic in vectors2 must be updated to handle vectors clusters at prefix level, not only at resource level.
This PR also removes an unnecessary sanity check that was splitting work units after deletion.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_vectors2/src/disk/key_value.rs
+++ b/nucliadb_vectors2/src/disk/key_value.rs
@@ -163,6 +163,16 @@ pub fn get_value<S: Slot>(interface: S, src: &[u8], id: usize) -> &[u8] {
     interface.read_exact(&src[pointer..]).0
 }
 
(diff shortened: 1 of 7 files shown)


bloodbare (approved):

nucliadb_vectors2/src/disk/key_value.rs
```diff
@@ -163,6 +163,16 @@ pub fn get_value<S: Slot>(interface: S, src: &[u8], id: usize) -> &[u8] {
     interface.read_exact(&src[pointer..]).0
 }
 
+// O(1)
+pub fn get_keys<'a, S: Slot + Copy + 'a>(
```
alekece, line 167: Can you please add documentation ?

PR #400, opened 2022-11-25
Fixing filter and factets and empty query

bloodbare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/search.py
+++ b/nucliadb/nucliadb/search/api/v1/search.py
@@ -68,7 +68,7 @@ async def search_knowledgebox(
     request: Request,
     response: Response,
     kbid: str,
-    query: str = Query(default="", min_length=3),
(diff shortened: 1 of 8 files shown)


hermeGarcia (approved):

nucliadb/nucliadb/search/api/v1/search.py
```diff
@@ -169,6 +169,12 @@ async def search(
     audit = get_audit()
     timeit = time()
 
+    if len(item.query) > 0 and len(item.query) < 3:
+        raise HTTPException(
+            status_code=422,
+            detail="Query needs to be bigger than 2 or 0",
```
ebrehault, line 175: that's really a weird message, `Query needs to be bigger than 2 or 0` sounds like bigger than 2 and bigger than 0, maybe `Query needs to be bigger than 2 or empty` would be less confusing

PR #401, opened 2022-11-28
Send cluster members at fix interval

alekece:
### Description
This PR is a little workaround to handle ingest restarting properly. The list of cluster members is sent to the ingest node in case of changes but now also every x times (if any changes happen during that interval).

### How was this PR tested?
We have to check it on stashify.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1732,6 +1732,7 @@ dependencies = [
  "dockertest",
  "env_logger 0.9.3",
  "log",
+ "parse_duration",
  "rand 0.8.5",
  "serde",
  "serde_json",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,15 +47,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


lferran:
@alekece let's say that the ingest container crashes due to a bug. When it restarts, will it pick the right information now? (Even if there haven't been any changes in terms of cluster membership)

alekece:
> @alekece let's say that the ingest container crashes due to a bug. When it restarts, will it pick the right information now? (Even if there haven't been any changes in terms of cluster membership)

Exactly. The behaviour is the cluster manager will wait for fixed period of time (let's say 30 seconds) cluster update. If no update happen, it will send the current list of cluster member to the ingest node. 
So, if for any reason, the ingest node crash and restart, the cluster manager will eventually send the right information (with update or not).

bloodbare (approved):

hermeGarcia (approved):
Great work! Just a couple of minor suggestions.

PR #416, opened 2022-12-01
Store content_type on external file fields when input by the user

lferran:
### Description
Described in title

### How was this PR tested?
Adding an integration test

### DIFF
--- a/nucliadb/nucliadb/one/tests/test_field_external_file.py
+++ b/nucliadb/nucliadb/one/tests/test_field_external_file.py
@@ -52,7 +52,7 @@ async def test_external_file_field(nuclia_jwt_key, nucliadb_api, knowledgebox_on
             json={
                 "slug": "resource1",
                 "title": "Resource 1",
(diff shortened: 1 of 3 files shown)


lferran:
[sc-3156]

nucliadb/nucliadb/one/tests/test_field_external_file.py
```diff
@@ -28,7 +28,9 @@
 TEST_EXTERNAL_FILE_FIELD_PAYLOAD = {
     "file": {
         "uri": EXTERNAL_FILE_URL,
+        "content_type": "video/webm",
```
jotare, line 31: Maybe you should change the extension of `EXTERNAL_FILE_URL` or create a second payload

jotare (approved):

nucliadb/nucliadb/one/tests/test_field_external_file.py
lferran: It doesn't matter much, as it is just test data. 

However, I set it to be different than pdf explicitly, so that is clear that whatever the user inputs will be stored.

PR #421, opened 2022-12-02
Add load score in cluster node state

alekece:
### Description
This PR aims to add load score in cluster node state in order to the future load balancer to perform shard cutover with fresh data.

### How was this PR tested?
We have to check on stashify.

### DIFF
--- a/nucliadb_cluster/src/cluster.rs
+++ b/nucliadb_cluster/src/cluster.rs
@@ -20,6 +20,11 @@ use crate::error::Error;
 
 const NODE_TYPE_KEY: &str = "node_type";
 const LOAD_SCORE_KEY: &str = "load_score";
+
+pub trait Score {
+    fn compute_score(&self) -> f32;
+}
+
 /// The ID that makes the cluster unique.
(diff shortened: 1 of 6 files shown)


nucliadb_cluster/src/cluster.rs
```diff
@@ -18,6 +18,13 @@ use uuid::Uuid;
 
 use crate::error::Error;
 
+const NODE_TYPE_KEY: &str = "node_type";
+const LOAD_SCORE_KEY: &str = "load_score";
+
+pub trait Score {
```
hermeGarcia, line 24: I dont think this trait is a good abstraction, I would expect for a type implementing `Score` to be already a score, not to be in charge of computing a score. Maybe the function should be called `score` or `value`.

hermeGarcia (approved):

nucliadb_cluster/src/cluster.rs
alekece: Done in 1e220ac8f0c0883c27cb301ea528561520c2106b

PR #428, opened 2022-12-08
bigger datapoints + removed redundant delete

hermeGarcia:
### Description
`nucliadb_vectors2` state needs to grow at a slower pace, in order to achieve this the following PR includes:
- Bigger datapoints, one resource will produce exactly one datapoint.
- Removing redundant delete logs. The index should rely on the `sentence_to_delete` and `vectors_to_delete`.

### How was this PR tested?
Local tests.

### DIFF
--- a/vectors_benchmark/src/binaries/1m_stats.rs
+++ b/vectors_benchmark/src/binaries/1m_stats.rs
@@ -39,12 +39,7 @@ fn label_set(batch_id: usize) -> Vec<String> {
         .collect()
 }
 
-fn add_batch(
-    writer: &mut Index,
-    _batch_id: String,
-    elems: Vec<(String, Vec<f32>)>,
-    labels: Vec<String>,
-) {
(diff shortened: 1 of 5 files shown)


vectors_benchmark/src/binaries/1m_stats.rs
```diff
@@ -41,7 +41,7 @@ fn label_set(batch_id: usize) -> Vec<String> {
 
 fn add_batch(
     writer: &mut Index,
-    batch_id: String,
+    _batch_id: String,
```
alekece, line 44: Please remove this parameter 🐱

hermeGarcia: Done at [349c9e9](https://github.com/nuclia/nucliadb/pull/428/commits/349c9e90ef4540f50001078123025ec50dee56e4)

lferran (approved):

PR #432, opened 2022-12-09
Add field classification labels on resource basic

lferran:
### Description
This PR adds a new field on basic metadata which stores the classification labels computed while processing.
For now, only the field level labels are added.

We also clean stale classification labels if individual fields are removed from a resource

### How was this PR tested?
Adding unit and integration tests


bloodbare (approved):
Would be great to document  on code why we need deleted_fields

lferran:
[sc-3198]

PR #433, opened 2022-12-09
temporal marks added to transactions

hermeGarcia:
### Description
This PR uses temporal marks to unify deletes and inserts during a `set_resource`. Thanks to this technique the state size  is considerably smaller.

### How was this PR tested?
Local tests


alekece (approved):
LGTM !

PR #447, opened 2022-12-13
Improve search performance by asynchronously fetch cloud resources

jotare:


bloodbare (changes requested):
Concurrency on fetching data with cache requires a lock on cache management.

Paralelization of sentences and paragraphs is enough, paralelization of labels and text may end up with an explosion of concurrent tasks. Paralelization of the 3 merges may be interesting.

vangheem:
Should we close this?

vangheem:
Closing for now but we can reopen if we think we need to revive!

PR #448, opened 2022-12-14
Add prometheus values in charts

alekece:
### Description
This PR aims to define prometheus environment variables in order to send metrics to prometheus **AND** update the load score of the writer node.

*Note that the `prometheus.url` is not defined in charts in order to keep the repository agnostic.*

### How was this PR tested?
Not tested.


jotare (approved):

charts/nucliadb_writer/values.yaml
```diff
@@ -54,3 +54,7 @@ tracing:
   jaegerAgentTag: 1.34.1
   jaegerCollectorHost: jaeger-collector.observability.svc.cluster.local
   jaegerCollectorGrpcPort: 14250
+
+prometheus:
```
bloodbare, line 58: @ferpizza

PR #450, opened 2022-12-14
Always Ppull latest cluster manager always on ingest and search

lferran:
### Description

### How was this PR tested?


bloodbare (approved):
We should define a versioning for cluster to avoid Versioning but lets fix first

lferran:
> We should define a versioning for cluster to avoid Versioning but lets fix first

noted. Will work on it

PR #452, opened 2022-12-15
Knowledge graph search

jotare:
### Description
Implement knowledge graph search powered by relations index

### How was this PR tested?
Integration tests in both sides (python and rust)


nucliadb_relations/Cargo.toml
```diff
@@ -1,3 +1,5 @@
+# WARNING! This package is deprecated. Look at nucliadb_vectors2/relations instead!
+
```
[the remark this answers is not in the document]
alekece: However, there is no such package in Rust. Could you rename `This package` to `This crate` please ?

nucliadb_node/tests/common/node_reader.rs
```diff
@@ -0,0 +1,87 @@
...
+        .authority(format!("{}:{}", READER_HOST, READER_PORT))
+        .path_and_query("/")
+        .build()
+        .unwrap();
+
+    let mut wait = 100;
+    const MAX_WAIT: u64 = 5000;
+    while let Err(error) = Channel::builder(server_uri.clone()).connect().await {
+        tokio::time::sleep(tokio::time::Duration::from_millis(wait)).await;
+        wait = linear_backoff(wait, 2);
```
alekece, line 75: Could it be better to use exponential backoff instead of linear ? 
And you should take a look of this crate https://crates.io/crates/backoff :cat:

nucliadb_node/tests/common/node_reader.rs
```diff
@@ -0,0 +1,87 @@
...
+        if wait > MAX_WAIT {
+            panic!(
+                "Something went wrong while starting reader gRPC server (too many times): {:?}",
+                error
+            );
+        }
+    }
+
+    let mut stub: HealthClient<Channel> = HealthClient::new(Channel::builder(server_uri).connect().await.unwrap());
```
alekece, line 84: This `unwrap` call might break. Why the `stub` variable is not initialized in the upper while loop ?

nucliadb_node/tests/common/node_writer.rs
```diff
@@ -0,0 +1,148 @@
...
+        panic!(
+            "writer gRPC server not serving, it's {:?}!",
+            result.get_ref().status()
+        );
+    }
+}
+
+
+/// Create a new Node Writer gRPC Client
+pub async fn node_writer_client() -> TestNodeWriter {
```
alekece, line 101: Looks like the codes from `node_writer` and the `node_reader` really similar. Maybe you can factorize it

nucliadb_node/tests/common/node_reader.rs
```diff
@@ -0,0 +1,107 @@
...
+        .build()
+        .unwrap();
+
+    let mut wait = 100;
+    const MAX_WAIT: u64 = 5000;
+    while let Err(error) = Channel::builder(server_uri.clone()).connect().await {
+        tokio::time::sleep(tokio::time::Duration::from_millis(wait)).await;
+        wait = linear_backoff(wait, 2);
+        if wait > MAX_WAIT {
+            panic!(
```
hermeGarcia, line 77: These constructions where you are panicking if a certain condition is fulfilled can be better written using `assert!`. For example in this case:
```rust
assert!(wait <= MAX_WAIT, "Something went wrong while starting reader gRPC server (too many times): {error:?}"); 
```

nucliadb_node/src/services/writer.rs
```diff
@@ -249,9 +249,11 @@ impl ShardWriterService {
         let relation_writer_service = self.relation_writer.clone();
         let relation_resource = resource.clone();
         let relation_task = move || {
+            println!("Relation service starts set_resource");
```
hermeGarcia, line 252: Must remove before merge

nucliadb_node/src/services/reader.rs
```diff
@@ -385,7 +381,9 @@ impl ShardReaderService {
             .flat_map(|relation| {
                 relation
                     .unwrap()
-                    .neighbours
+                    .prefix
+                    .unwrap()
```
alekece, line 385: Can you confirm that this `unwrap` call is safe ? If yes, can you add some comment to explain it

nucliadb_node/src/services/reader.rs
```diff
@@ -485,26 +483,43 @@ impl ShardReaderService {
...
 
+        let relation_request = search_request.relations;
+        let relation_reader_service = self.relation_reader.clone();
+        let relation_task = move || {
+            if !skip_relations {
+                let results = relation_reader_service.search(&relation_request.unwrap());
+                Some(results)
+            } else {
+                None
+            }
```
alekece, line 494: This code is a bit convoluted and can be replaced by the following functional code:
```rust
search_request
   .relations
   .map(|relations| relation_reader_service.search(relation)
```
By doing that, you get rid of the not-useful-anymore `skip_relations` variable, and make the code clearer.

nucliadb_protos/nodereader.proto
```diff
@@ -202,39 +202,88 @@ message VectorSearchResponse {
...
-    // Will filter the search to nodes of type ntype.
-    utils.RelationNode.NodeType ntype = 1;
-    // Additionally the search can be even more specific by 
+message RelationNodeFilter {
+    utils.RelationNode.NodeType node_type = 1;
+
+    // Additionally the search can be even more specific by
+    // providing a subtype. The empty string is a wilcard that 
+    // indicates to not filter by subtype. 
+    string subtype = 2;
```
alekece, line 211: Instead of using an empty string, it would be more idiomatic to use an optional here, right ?

nucliadb_protos/python/nucliadb_protos/nodereader_pb2_grpc.pyi
```diff
@@ -3,10 +3,10 @@
 isort:skip_file
```
alekece, line 3: Is it normal to commit those `*.pyi` files ?

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -18,15 +18,14 @@
...
-use std::collections::{HashSet, LinkedList};
+use std::collections::{HashMap, HashSet, LinkedList};
 
 use super::errors::*;
 use crate::relations::graph_db::*;
 
 // BfsGuide allows the user to modify how the search will be performed.
 // By default a BfsGuide does not interfere in the search.
 pub trait BfsGuide {
-    fn matches(&self, node: Entity) -> bool;
```
alekece, line 29: What is the rational behind removing this method ? Is it some dead code ?

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -76,55 +81,69 @@ where Guide: BfsGuide
 impl<'a, Guide> BfsEngine<'a, Guide>
 where Guide: BfsGuide
 {
-    pub fn search(mut self) -> RResult<Vec<Entity>> {
-        std::mem::take(&mut self.entry_points)
+    pub fn search(mut self) -> RResult<impl Iterator<Item = GCnx>> {
```
alekece, line 84: Nice improvement in term of user interface :cat:

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -76,55 +81,69 @@ where Guide: BfsGuide
...
+            .chain(self.graph.get_inedges(self.txn, node.point)?)
             .flat_map(|a| a.ok().into_iter())
             .filter(|edge| node.depth < self.max_depth || self.guide.free_jump(*edge))
             .filter(|edge| self.guide.edge_allowed(edge.edge()))
             .filter(|edge| self.guide.node_allowed(edge.to()))
             .for_each(|edge| {
-                if !self.visited.contains(&edge.to()) {
+                let is_free_jump = self.guide.free_jump(edge);
+                let can_use_free_jump = same_level.contains_key(&node.point);
+                if !is_free_jump && !can_use_free_jump {
```
alekece, line 112: I don't know how critical is that `expand` method but you can avoid to calculate `can_use_free_jump` variable if `is_free_jump` is true:
```rust
if !self.guide.free_jump(edge) && same_level.contains_key(&node.point) { ... }
```
It's just a small tiny optimization so feel free to use or not.

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -76,55 +81,69 @@ where Guide: BfsGuide
...
                     let node = BfsNode {
                         point: edge.to(),
-                        depth: node.depth + (!self.guide.free_jump(edge) as usize),
+                        depth: node.depth + 1,
```
alekece, line 115: Can you explain that `+ 1` offset ?

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -173,6 +192,7 @@ mod test {
             .unwrap();
         let expected = &nodes;
         let result = bfs.search().unwrap();
+        let result: Vec<_> = result.map(|cnx| cnx.to()).collect();
```
alekece, line 195: It is more idiomatic to put the `Vec<_>` stuff directly after the method call such as:
```rust
let result = result.map(|cnx| cnx.to()).collect::<Vec<_>>();
```

nucliadb_vectors/src/relations/bfs_engine.rs
```diff
@@ -192,6 +212,7 @@ mod test {
...
```
alekece, line 215: Ditto

nucliadb_protos/python/nucliadb_protos/nodereader_pb2_grpc.pyi
jotare: Yes, are part of the type checking for Python. I think we commit them for people using nucliadb_protos package. @bloodbare correct me if I'm wrong

nucliadb_vectors/src/relations/bfs_engine.rs
jotare: I don't know, @hermeGarcia delete it

nucliadb_vectors/src/relations/bfs_engine.rs
jotare: @hermeGarcia :)

nucliadb_vectors/src/relations/bfs_engine.rs
jotare: I think it's increasing the depth of the BFS, if it's not a free jump (i.e., it count as a jump), the next node is at the distance of the current node + 1

nucliadb_vectors/src/relations/bfs_engine.rs
jotare: Fixed

nucliadb_vectors/src/relations/bfs_engine.rs
jotare: I think here a named variables makes the code more readable. From performance point of view, I suppose Rust compiler will write the same code for both options

nucliadb/nucliadb/search/search/merge.py
```diff
@@ -355,17 +397,20 @@ async def merge_results(
     show: List[ResourceProperties],
     field_type_filter: List[FieldTypeName],
     extracted: List[ExtractedDataTypeName],
+    query: SearchRequest,
```
lferran, line 400: NIT: would it make more sense to pass the relations metadata of the query?(ie: `query.relations` for instance)

nucliadb/nucliadb/search/search/merge.py
```diff
@@ -347,6 +353,42 @@ async def merge_paragraph_results(
...
+            elif destination in relations.entities:
+                relations.entities[destination].related_to.append(
+                    DirectionalRelation(
+                        entity=origin,
+                        relation=relation_label,
+                        direction=RelationDirection.IN,
+                    )
+                )
+            else:
+                assert False, "origin or destination must be in the response"
```
lferran, line 387: This will make the search fail with a 500. Is this intented? Have you considered doing a logger.warning and moving on?

nucliadb/nucliadb/tests/test_search.py
```diff
@@ -476,3 +480,251 @@ async def inject_resource_with_a_sentence(knowledgebox, writer):
...
+        json={
+            "title": "My resource",
+            "slug": "myresource",
+            "summary": "Some summary",
+        },
+    )
+    assert resp.status_code == 201
+    rid = resp.json()["uuid"]
+
+    relation_nodes = {
```
lferran, line 507: NIT: would be nice to have this in a fixture to use in future tests

nucliadb/nucliadb/search/search/merge.py
jotare: It should be improved, but the reason is that if it fails, the node is returning bad results (and should never happen). We can log or send an issue to sentry and continue with the request (makes more sense)

nucliadb/nucliadb/tests/test_search.py
jotare: Agree, moving it to a fixture

nucliadb_node/src/services/reader.rs
jotare: Done

nucliadb_protos/nodereader.proto
jotare: Totally agree

nucliadb_node/src/services/reader.rs
jotare: It should be, as the request is a search for prefix, a `None` here would be a bug. I can't write it down or express in a different way to be more explicit about this fact

jotare: I'll change it for an `expect`

nucliadb_vectors/src/relations/bfs_engine.rs
alekece: Could you please add a comment to explain it ? I am sure @hermeGarcia will complete the command later on with further explanation :cat:

jotare: Done

nucliadb/nucliadb/search/search/merge.py
jotare: I'll change it for a `logger.error` and send an error to Sentry

bloodbare (approved):

PR #454, opened 2022-12-16
Update references to Quickwit to NucliaDB

swallez:
### Description

While looking at NucliaDB I found some references to Quickwit that should obviously have been updated to be NucliaDB. This PR fixes that.

There are still two references that this PR doesn't update:

* CODE_STYLE_RUST.md: this should be reworded for NucliaDB  
  https://github.com/nuclia/nucliadb/blob/2371a6707fa42f8ebac9fa741398ef6fd4541019/CODE_STYLE_RUST.md?plain=1#L117-L118
* `nucliadb_cluster/LICENSE-AGPL`: it's a fork, so the copyrights should probably be NucliaDB?  
  https://github.com/nuclia/nucliadb/blob/2371a6707fa42f8ebac9fa741398ef6fd4541019/nucliadb_cluster/LICENSE-AGPL#L1

### How was this PR tested?

No tests needed for docs, and I'm not sure the change in `nucliadb_telemetry/src/sender.rs` need one.


bloodbare:
Thanks @swallez for the PR !

bloodbare (approved):

bloodbare:
There is a bug on registry to download docker images, let us improve the system so it can download images from docker hub instead.

PR #461, opened 2022-12-21
Fix import when nucliadb is in the cloud

lferran:
#### Problem
When importing a resource that was previously exported, we call ingest's gRPC `ProcessMessage` directly. This emulates the ingestion of a resource, but outside of a normal transaction (That is why we are setting a `tid=-1` [here](https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/ingest/service/writer.py#L255)). 

This, in effect, makes that the node sidecar can't download the corresponding index message to push to the node, as it depends on a valid transaction id.

#### Solution
On the shard object, before sending the message to the node, we detect when `tid` is `-1` and will treat the operation as if it was a reindexing operation (which also doesn't rely on any transaction).

### How was this PR tested?
Local tests will test the import step.
Existing tests give us the confidence that we don't break anything.

### DIFF
--- a/nucliadb/nucliadb/ingest/consumer/pull.py
+++ b/nucliadb/nucliadb/ingest/consumer/pull.py
@@ -192,7 +192,7 @@ class PullWorker:
     async def subscription_worker(self, msg: Msg):
         subject = msg.subject
         reply = msg.reply
-        seqid = int(msg.reply.split(".")[5])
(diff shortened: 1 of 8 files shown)


lferran:
[sc-2903]

bloodbare (approved):

PR #462, opened 2022-12-22
delete does not require for the shard to be load

hermeGarcia:
### Description
Deleting a shard required for it to be loaded. This causes problems if the shard is corrupted, this PR fixes the problem.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_node/src/writer/grpc_driver.rs
+++ b/nucliadb_node/src/writer/grpc_driver.rs
@@ -101,7 +101,7 @@ impl NodeWriter for NodeWriterGRPCDriver {
         self.instrument(&request);
 
         info!("gRPC delete_shard {:?}", request);
-        // Deletion does not requiere for the shard
+        // Deletion does not require for the shard
         // to be loaded.
         let shard_id = request.into_inner();
         let mut writer = self.0.write().await;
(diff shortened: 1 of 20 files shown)


nucliadb_node/src/writer/grpc_driver.rs
```diff
@@ -101,23 +101,19 @@ impl NodeWriter for NodeWriterGRPCDriver {
         self.instrument(&request);
 
         info!("gRPC delete_shard {:?}", request);
-
+        // Deletion does not requiere for the shard
```
alekece, line 104: Little typo here `requiere` => `require`

nucliadb_node/src/writer/grpc_driver.rs
```diff
@@ -130,8 +126,9 @@ impl NodeWriter for NodeWriterGRPCDriver {
 
         info!("gRPC delete_shard {:?}", request);
 
+        // Deletion and upgrade do not requiere for the shard
```
alekece, line 129: Ditto

alekece (approved):
Nice catch !

nucliadb_node/src/writer/grpc_driver.rs
hermeGarcia: Ouch true! Thanks for noticing, fixed it at [ad5a4c3](https://github.com/nuclia/nucliadb/pull/462/commits/ad5a4c34b67331b5357922d335e22f3c09505558)

PR #467, opened 2022-12-28
advance seach field

hermeGarcia:
### Description
Node users should be able to take advantage of the query language offered by Tantivy.
This PR allows users to do such thing through the field `advance_query`. Said field should contain a valid Tantivy query, meaning that if it does not follow the expected grammar and error will be returned. Once the `advance_query` is parsed it will be run with the original query as if they where one.

### How was this PR tested?
Local tests.


nucliadb_protos/nodereader.proto
```diff
@@ -54,10 +54,11 @@ message DocumentSearchRequest {
     int32 result_per_page = 8;
 
     Timestamps timestamps = 9;
-    optional noderesources.Resource.ResourceStatus with_status = 16;
-
+    
     bool reload = 10;
     bool only_faceted = 15;
+    optional noderesources.Resource.ResourceStatus with_status = 16;
+    optional string advance_query = 17;
```
lferran, line 61: typo: `advanced_query`

lferran (approved):

PR #470, opened 2022-12-29
Add load score in chichat/members route response

alekece:
### Description
This PR aims to add the load score in the `chitchat/members` route response and also in the gRPC `ListMembers` method.

### How was this PR tested?
Not tested.


nucliadb_protos/writer.proto
```diff
@@ -320,11 +320,14 @@ message Member {
     /// If true, it means self.
     bool is_self = 3;
 
-    /// Writer, Node, Reader
+    /// Io, Ingest, Search, Train.
```
lferran, line 323: Not for this PR probably, but it would be nice to change it to be an enum

lferran (approved):

PR #475, opened 2022-12-30
Delete KB keys in batches while purging

lferran:
### Description
This PR changes the way we delete the KB keys from the main database. Before we were deleting them all at once, while now we will delete them in batches of 1000.

### How was this PR tested?
Added integration and unit tests.
I tested against a local tikv server and a batch size of 1000 keys isn't causing any trouble

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/knowledgebox.py
+++ b/nucliadb/nucliadb/ingest/orm/knowledgebox.py
@@ -18,7 +18,7 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 from datetime import datetime
-from typing import AsyncGenerator, AsyncIterator, Optional, Tuple, Union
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -486,17 +486,22 @@ async def purge(cls, driver: Driver, kbid: str):
         await cls.delete_all_kb_keys(driver, kbid)
 
     @classmethod
-    async def delete_all_kb_keys(cls, driver: Driver, kbid: str):
+    async def delete_all_kb_keys(cls, driver: Driver, kbid: str, chunk_size=1_000):
```
jotare, line 489: I think chunk size should not be a parameter, as it's a detail not important for callers wanting to delete all KB keys

lferran: agree! Changed in my last commit

jotare (approved):

PR #476, opened 2023-01-02
Add positions for sentences on search results (without the page number)

lferran:
### Description
This PR adds positions to the sentence object of search results.

This will not incur on worse performance of the search endpoint yet, because we are not returning the page number.

### How was this PR tested?
Integration backend tests


nucliadb/nucliadb/ingest/orm/brain.py
```diff
@@ -109,19 +110,6 @@ def __init__(self, rid: str):
     def apply_field_text(self, field_key: str, text: str):
         self.brain.texts[field_key].text = text
 
-    def get_paragraph_page_number(
```
bloodbare, line 112: I would leave this utility somewere, as it will be usefull

bloodbare (approved):

PR #482, opened 2023-01-05
Fix setting icon

lferran:
### Description
This PR fixes the logic for setting the icon of resources in two scenarios:
 - For resources with external file fields where content_type is set by the user.
 - For resources with regular file fields where content_type is detected while processing.

Additionally, I moved the functional tests regarding the external file field feature from `one` to `nucliadb`.

### How was this PR tested?
Integration tests


lferran:
[sc-3156]

bloodbare (approved):

PR #485, opened 2023-01-10
Fix purge: use different transactions for reading keys and for deletion

lferran:
### Description
We introduced a bug in the purge command where we were using the same tikv transaction for reading all kb keys and for deleting them. This is not possible with tikv.
We were not catching the bug in tests because they were running with the redis driver instead.

### How was this PR tested?
Added a tikv driver integration test

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/knowledgebox.py
+++ b/nucliadb/nucliadb/ingest/orm/knowledgebox.py
@@ -18,7 +18,7 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 from datetime import datetime
-from typing import Any, AsyncGenerator, AsyncIterator, List, Optional, Tuple, Union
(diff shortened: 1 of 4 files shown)


nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -486,22 +486,23 @@ async def purge(cls, driver: Driver, kbid: str):
         await cls.delete_all_kb_keys(driver, kbid)
 
     @classmethod
-    async def delete_all_kb_keys(cls, driver: Driver, kbid: str):
-        # Delete KB Keys
+    async def delete_all_kb_keys(
+        cls, driver: Driver, kbid: str, chunk_size: int = 1_000
```
lferran, line 490: @jotare I have added back the `chunk_size` parameter because it makes testing way easier.
Otherwise I have to create more than 1000 resources in a test and the test takes too long.

jotare (approved):

PR #486, opened 2023-01-11
Always calculate load score

alekece:
### Description
This PR re-introduces the commits about always calculating the load score even if the prometheus environment variables are not set.

### How was this PR tested?
Copy (partially) shards folder from `node-1` and run locally the node writer.

### DIFF
--- a/nucliadb_node/src/bin/writer.rs
+++ b/nucliadb_node/src/bin/writer.rs
@@ -113,12 +113,6 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let metrics_task = tokio::spawn(async move {
         info!("Start metrics task");
 
-        let mut node_reader = NodeReaderService::new();
-
-        if let Err(e) = node_reader.load_shards() {
-            warn!("Not all the shards have been loaded: {e}");
-        };
-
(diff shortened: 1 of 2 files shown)


hermeGarcia (approved):

nucliadb_node/src/bin/writer.rs
```diff
@@ -108,47 +108,55 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        let report = NodeReport::new(host_key.to_string())?;
-        let mut metrics_publisher = Publisher::new("node_metrics", prometheus_url);
+        let mut node_reader = NodeReaderService::new();
 
-        if let Some((username, password)) =
-            Configuration::get_prometheus_username().zip(Configuration::get_prometheus_password())
-        {
-            metrics_publisher = metrics_publisher.with_credentials(username, password);
-        }
+        if let Err(e) = node_reader.load_shards() {
```
hermeGarcia, line 118: load shards will open all the shards, is there a way to avoid this?

hermeGarcia: Alternative implemented at [2164223](https://github.com/nuclia/nucliadb/pull/486/commits/2164223a5d2ff1aa68d4920b4ebe69424aaecbf2)

PR #490, opened 2023-01-12
Prevent processed resources from disappearing from resources list page

lferran:
### Description
This PR tweaks the way we are indexing the processing status:
 - Facets: the current processing status is always set there
 - Status document tantivy field:
   - Once it becomes PROCESSED it stays like that forever.
   - Once it becomes ERROR, it can only go to PROCESSED.
   
### How was this PR tested?
Integration tests

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/brain.py
+++ b/nucliadb/nucliadb/ingest/orm/brain.py
@@ -326,21 +326,30 @@ class ResourceBrain:
     def set_processing_status(
         self, basic: Basic, previous_status: Optional[Metadata.Status.ValueType]
     ):
-        if previous_status and previous_status == Metadata.Status.PROCESSED:
+        status = basic.metadata.status
+        useful = basic.metadata.useful
+        if previous_status == Metadata.Status.PROCESSED:
             # For resources that have been processed successfully, we force their status to be
             # always PROCESSED so that they don't disappear from the resource list page.
             self.brain.status = PBBrainResource.PROCESSED
-            return
-        status = basic.metadata.status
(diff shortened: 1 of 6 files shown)


nucliadb/nucliadb/ingest/orm/brain.py
```diff
@@ -59,11 +59,10 @@
...
 
-PROCESSING_STATUS_PB_TYPE_TO_NAME_MAP = {
-    PBBrainResource.ERROR: ResourceProcessingStatus.ERROR.name,
-    PBBrainResource.EMPTY: ResourceProcessingStatus.EMPTY.name,
-    PBBrainResource.PROCESSED: ResourceProcessingStatus.PROCESSED.name,
-    PBBrainResource.PENDING: ResourceProcessingStatus.PENDING.name,
+METADATA_STATUS_PB_TYPE_TO_NAME_MAP = {
+    Metadata.Status.ERROR: ResourceProcessingStatus.ERROR.name,
+    Metadata.Status.PROCESSED: ResourceProcessingStatus.PROCESSED.name,
+    Metadata.Status.PENDING: ResourceProcessingStatus.PENDING.name,
```
bloodbare, line 65: Empty is missing ?

lferran: `EMPTY` is not part of the `Metadata.Status` [proto definition](https://github.com/nuclia/nucliadb/blob/main/nucliadb_protos/resources.proto#L263). I added `BLOCKED` and `EXPIRED` though, just in case

bloodbare (approved):

PR #497, opened 2023-01-13
Sort by relevance

jotare:
Add two new parameters to `/search` endpoint to be able to perform sorting onto a subset of a search.

- `sort_limit` limits the size of the search when `sort` is passed
- `sort_order` allows ordering (ascendant or descendant)

On POST requests, all sort options are passed together in a new `SortOptions` model, where the dependency between options is well defined


nucliadb/nucliadb/search/search/query.py
```diff
@@ -48,6 +48,7 @@ async def global_query_to_pb(
     fields: Optional[List[str]] = None,
     sort: Optional[str] = None,
     sort_ord: int = Sort.ASC.value,
+    sort_limit: int = 100,
```
lferran, line 51: Should we cap it to some value no matter what the user input?

jotare: It might be a good idea, but I'm not sure what should be that limit...

nucliadb/nucliadb/search/search/merge.py
```diff
@@ -62,22 +63,65 @@
     ResourceSearchResults,
     Sentence,
     Sentences,
+    SortField,
+    SortOptions,
+    SortOrder,
     TextPosition,
 )
 
+Score = Any
```
lferran, line 72: Maybe it looks ugly, but you can be more specific about the type for score.
Something like:

```python
NodeScore = Tuple[int, int]
TimestampScore = datetime.datetime
TitleScore = str

Score = Union[NodeScore, TimestampScore, TitleScore]
```

nucliadb/nucliadb/search/search/merge.py
```diff
@@ -62,22 +63,65 @@
...
+        score = basic.created.ToDatetime()
+    elif sort.field == SortField.MODIFIED:
+        score = basic.modified.ToDatetime()
+    elif sort.field == SortField.TITLE:
+        score = basic.title
+
+    return score
+
+
+async def document_score(
```
lferran, line 104: is this function useful? same for `paragraph_score`

nucliadb/nucliadb/search/search/query.py
```diff
@@ -67,7 +67,12 @@ async def global_query_to_pb(
 
     # We need to ask for all and cut later
     request.page_number = 0
-    request.result_per_page = page_number * page_size + page_size
+    if sort is not None:
+        # As the index can't sort, we have to do it when merging. To
+        # have consistent results, we must limit them
+        request.result_per_page = sort.limit
```
lferran, line 73: what happens if page_size provided by the user is >= sort.limit?

nucliadb/nucliadb/search/search/query.py
lferran: Some thoughts:

I wonder if the sort_limit should not be exposed as an endpoint parameter... maybe it makes sense for the sort_limit to be somehow compared to the page_size that the user is requesting.

For instance: if the user is requesting 100 results + sorting, we calculate sort.limit as 100 + (100 * 0.5) 

This way, if we enforce a limit to page_size (which we probably should), then we also have an upper bound for sort.limit

nucliadb/nucliadb/search/search/query.py
jotare: They won't have any result. The idea here is that the user is requesting the `sort.limit` most relevant results and then sorting them by some criteria. If you want more, you need to request more.

nucliadb/nucliadb/search/search/merge.py
jotare: Sounds nicer, changed

lferran (approved):

bloodbare (approved):

PR #502, opened 2023-01-17
Various storage fixes

lferran:
### Description
This PR includes a collection of bugfixes found on both local and gcs storages:
 - GCS object's custom metadata was not stored properly according to the docs (needs to be inside the `metadata` key)
 - Local file storage `exists()` was not working properly
 - DownloadManager.update method has been converted to a sync method as it doesn't need I/O
 
### How was this PR tested?
Integration tests


lferran:
[ch-3552]

nucliadb_utils/nucliadb_utils/storages/gcs.py
```diff
@@ -226,7 +226,14 @@ async def start(self, cf: CloudFile) -> CloudFile:
             quote_plus(upload_uri),
         )
         metadata = json.dumps(
-            {"FILENAME": cf.filename, "SIZE": cf.size, "CONTENT_TYPE": cf.content_type}
+            {
+                "contentType": cf.content_type,
```
jotare, line 230: Content type should be duplicated?

lferran: I left it duplicated for bw/compatibility, just in case.

`contentType` is part of the contract with GCS, whereas `metadata.CONTENT_TYPE` is part of the custom metadata we control

lferran: actually, that is inferred from the header, so you are right, we don't need to duplicate it

bloodbare (approved):

PR #505, opened 2023-01-18
optional depth parameter on relations

hermeGarcia:
### Description
Node iterations with complex filters requires ignoring the depth of the search. To provide a clean interface depth must be optional.

### How was this PR tested?
Local tests


alekece (approved):
Great work ! `depth` variable name does not reflect super well that if depth is `Option::None` it would not limit the `bfs` at all

PR #511, opened 2023-01-18
Improve knowledge graph search

jotare:
- Added label to relations
- Added entity type to serialized relations
- Breaking changes on relations API: relations are now closer to protobuffers. A relation is defined by a relation type, a label and two entities (from and to). If not defined, from is the resource

[sc-3672] https://app.shortcut.com/flaps/story/3672/knowledge-graph-return-empty-relations
[sc-3673] https://app.shortcut.com/flaps/story/3673/serialize-entity-type

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/brain.py
+++ b/nucliadb/nucliadb/ingest/orm/brain.py
@@ -49,7 +49,7 @@ from nucliadb_protos.utils_pb2 import (
 from nucliadb.ingest import logger
 from nucliadb.ingest.orm.labels import BASE_TAGS, flat_resource_tags
 from nucliadb.ingest.orm.utils import compute_paragraph_key
-from nucliadb_models.metadata import ResourceProcessingStatus
+from nucliadb_models.metadata import RelationTypePbMap, ResourceProcessingStatus
 
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


nucliadb/nucliadb/ingest/orm/brain.py
```diff
@@ -374,6 +374,7 @@ def set_global_tags(self, basic: Basic, uuid: str, origin: Optional[Origin]):
                         relation=Relation.COLAB,
                         source=relationnodedocument,
                         to=relationnodeuser,
+                        relation_label=RelationTypePbMap[Relation.COLAB].value,
```
bloodbare, line 377: Isnt it redundant ?

jotare: Yes and no. It's a way to label the automatic relations, as they are going to presented in the API, and avoid returning empty strings. If it's not desired, I can remove them.

bloodbare: Yes, the entity relation type already provides this information. We can change in the future if its required

bloodbare (changes requested):

bloodbare (approved):

PR #512, opened 2023-01-19
User query pre-search processing

lferran:
### Description
This PR adds a pre-processing step to the query input by the user.

**IMPORTANT NOTES**
I made the following decisions for these corner cases. Please add comments if you disagree:
 - if user inputs a query such that after post-processing, the query is "", then we just search with the initial query to avoid searching for all (query="").
 
 - If the user is searching by exact match (eg: within double quotes) then we don't pre-process it

### How was this PR tested?
Added unit and integration tests


nucliadb/nucliadb/search/search/query.py
```diff
@@ -188,3 +188,22 @@ async def paragraph_query_to_pb(
     ResourceProcessingStatus.PROCESSED: Resource.ResourceStatus.PROCESSED,
     ResourceProcessingStatus.ERROR: Resource.ResourceStatus.ERROR,
 }
+
+
+def pre_process_query(user_query: str) -> str:
```
jotare, line 193: I think it might be a good idea to use a Strategy design pattern here to have more flexibility switching between preprocessing algorithms

jotare (approved):

nucliadb/nucliadb/search/search/query.py
lferran: I agree that if the business logic becomes more complex, a Strategy pattern would help. However, I don't see the need right now. If it's OK, I'll leave a comment in the code.

PR #518, opened 2023-01-20
Convert load score to float

alekece:
### Description
This PR fixes the issue that the load score in propagated as a string (due to chitchat internal node state management).

### How was this PR tested?
Run ingest node, node writer and cluster manager with debug logs enabled and check that `ClusterMember::load_score` attribute contains a float and not a string anymore.

### DIFF
--- a/nucliadb/nucliadb/ingest/chitchat.py
+++ b/nucliadb/nucliadb/ingest/chitchat.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 import asyncio
 import binascii
 import json
-from typing import List, Optional
+from typing import Any, List, Optional
 
 from nucliadb.ingest import logger
 from nucliadb.ingest.orm.node import ClusterMember, chitchat_update_node
@@ -130,7 +130,7 @@ class ChitchatNucliaDB:
         self.task.cancel()
 
 
-def build_member_from_json(member_serial):
+def build_member_from_json(member_serial: dict[str, Any]):
     try:
         load_score = float(member_serial.get("load_score"))
     except TypeError:
--- a/nucliadb/nucliadb/ingest/chitchat.py
+++ b/nucliadb/nucliadb/ingest/chitchat.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 import asyncio
 import binascii
 import json
-from typing import List, Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 
 from nucliadb.ingest import logger
 from nucliadb.ingest.orm.node import ClusterMember, chitchat_update_node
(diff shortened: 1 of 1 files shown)


nucliadb/nucliadb/ingest/chitchat.py
```diff
@@ -112,7 +112,7 @@ async def socket_reader(
                                 listen_addr=x["address"],
                                 node_type=x["type"],
                                 is_self=x["is_self"],
-                                load_score=x.get("load_score", 0.0),
+                                load_score=float(x.get("load_score", "0")),
```
bloodbare, line 115: Could we try except this operation ?

alekece: Yes sure. In case of failure, log it and default load score to zero ?

bloodbare: yes

nucliadb/nucliadb/ingest/chitchat.py
```diff
@@ -135,3 +128,20 @@ async def socket_reader(
     async def close(self):
         self.chitchat_update_srv.close()
         self.task.cancel()
+
+
+def build_member_from_json(x):
```
bloodbare, line 133: Can you provide proper parameter name, on lambda makes sense but on function should be semantic naming

bloodbare (changes requested):

nucliadb/nucliadb/ingest/chitchat.py
```diff
@@ -135,3 +128,23 @@ async def socket_reader(
     async def close(self):
         self.chitchat_update_srv.close()
         self.task.cancel()
+
+
+def build_member_from_json(json):
```
bloodbare, line 133: json is a bad name, please use a semantic variable

bloodbare (changes requested):

nucliadb/nucliadb/ingest/chitchat.py
alekece: @bloodbare Do you have any suggestion in mind ? From my point of view, `json` is ok so far since we cannot assert anything else about it before validating and perform the actual conversion. Something like `maybe_member_json_repr` can be used here but it adds just noise IMO.

alekece: `member_serial` can be a good candidate BTW

bloodbare: member serial is a good name. json its a bad name because its a common libeary and is not semantic

nucliadb/nucliadb/ingest/chitchat.py
```diff
@@ -135,3 +128,25 @@ async def socket_reader(
     async def close(self):
         self.chitchat_update_srv.close()
         self.task.cancel()
+
+
+def build_member_from_json(member_serial):
```
jotare, line 133: Could you add type annotations for it?

jotare, line 133: Can you add a type annotation for this?

bloodbare (approved):

PR #520, opened 2023-01-23
Include large_computed_metadata on exports

lferran:
### Description
This metadata was not being exported. As a consequence, search results were slightly different in source kb and destination kb after an import.

### How was this PR tested?
Integration tests


alekece (approved):
Looks good !

PR #521, opened 2023-01-23
Increase gcrp max message size

lferran:
### Description
We have a resource with 130MB of size and the current limit is 100MB. 
This is making the node sidecar crash when sending the brain to the node.

Solution: increase the client's max_message_size to 250MB for now.

### How was this PR tested?
No tests.


bloodbare (changes requested):
telemetry should not be released isnt it?

bloodbare (approved):

lferran:
[sc3748]

PR #524, opened 2023-01-23
Prevent paragraph annotations from having no classifications

lferran:
### Description
Add fastapi valiadations on ParagraphAnnotation.classifications field so that we reject empty classification payloads with a HTTP422 error

### How was this PR tested?
Integration tests at the writer component


nucliadb_models/nucliadb_models/metadata.py
```diff
@@ -249,9 +249,23 @@ class TokenSplit(BaseModel):
...
 
 class ParagraphAnnotation(BaseModel):
-    classifications: List[Classification] = Field(Classification, min_items=1)
+    classifications: List[Classification] = []
     key: str
 
+    @root_validator()
+    def _validate_classifications(cls, values):
+        classifications = values.get("classifications") or []
+        if len(classifications) < 1:
```
alekece, line 258: `if len(classifications) == 0` looks more understandable if you want to check emptiness here

nucliadb_models/nucliadb_models/metadata.py
```diff
@@ -249,9 +249,23 @@ class TokenSplit(BaseModel):
...
+            raise ValueError("ensure this value has at least 1 items")
+        seen = set()
+        for cf in classifications:
+            cf_tuple = tuple(cf.dict().values())
+            if cf_tuple in seen:
+                raise ValueError("Paragraph classifications need to be unique")
+            else:
+                seen.add(cf_tuple)
```
alekece, line 266: Can you just use a set here and compare the length of both list and set to check uniqueness ?

lferran: I could do that, but i think it is better to raise the error as soon as there is a duplicate

lferran: After some local test, I think readability is better than the performance we get here. changing it!

alekece (approved):
Looks good, great work :cat:

PR #532, opened 2023-01-26
Clear kb counters cache after processing each message on import

lferran:
## Description
On normal ingest operation, the Ingest cosumer will reset the cache every time a new resource has been processed ([see](https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/ingest/consumer/pull.py#L228))

However, during import we are calling the grpc service ProcessMessage which simply processes the resource. This causes that the counters show all 0 after importing a bunch of resources if the KB overview dashboard page has been visited before.

This PR adds logic to clear the KB counters cache once after resources have been imported.

## How was this PR tested?
Integration tests


lferran:
[ch3360]

bloodbare (approved):

PR #536, opened 2023-01-26
Move one shard replica from one node to another

alekece:
### Description
This PR adds two methods in the node writer gRPC interface:
- **MoveShard**: indicates which shard to move to which node
- **AcceptShard**: indicates to a given node to accept any incoming shard. Note that if the shard already exists on the given node, the shard will be rejected. If you want to override it no matter what, set `override_shard` attribute to `true`.

giphy

### How was this PR tested?
I ran two `node_writer` instances locally using two distinct data folders then configure `Postman` to call respectively `MoveShard` and `AcceptShard` gRPC methods on both instances.

```sh
DATA_PATH=path/to/data/node1 cargo run --bin node_writer&
DATA_PATH=path/to/data/node2 CHITCHAT_PORT=40101 cargo run --bin node_writer&
```


nucliadb_node/src/writer/mod.rs
```diff
@@ -241,4 +247,54 @@ impl NodeWriterService {
             .collect();
         ShardIds { ids }
     }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn move_shard(
```
hermeGarcia, line 252: I dont think this logic should be in `NodeWriterService`, it is `async` and is tightly related with our grpc interface.  Maybe it would be better to move it to `NodeWriterGRPCDriver::move_shard`.

nucliadb_node/src/writer/mod.rs
```diff
@@ -241,4 +247,54 @@ impl NodeWriterService {
...
+            .await
+            .map(Some)
+            .map_err(|e| match e {
+                nucliadb_ftp::Error::IoError(e) => ServiceError::IOErr(e),
+                _ => ServiceError::GenericErr(Box::new(e)),
+            })
+    }
+
+    #[tracing::instrument(skip_all)]
+    pub async fn accept_shard(
```
hermeGarcia, line 276: Same as with `move_shard`.

nucliadb_node/src/writer/mod.rs
```diff
@@ -241,4 +247,54 @@ impl NodeWriterService {
...
+        &mut self,
```
hermeGarcia, line 277: This worries me, we are blocking the reader to accept the shard. I guess because of the `load_shard` call. That is why I think the logic of `accepting` a shard should be at on the grpc driver, before taking the service lock.

hermeGarcia (approved):

PR #541, opened 2023-01-30
Fix serialization of empty paragraph annotations

lferran:
### Description
We no longer accept paragraph annotations with empty classifications. We added the corresponding fastapi validations on the writer component (see [this commit to main](https://github.com/nuclia/nucliadb/commit/c7aa65726c7411c463822d9da12ff9ff5ee97fa4)), so we ensure new data is validated.

However, previously existing paragraph annotations with empty classifications now throw an error at resource serialization.

This PR removes the automatic fastapi validations and adds a specific validation on the writer side, so that already created data does not fail, but still we are validating any new data.

### How was this PR tested?
Integration tests


lferran:
[ch3752]

jotare (commented):
Do you think it might be a good idea to write a "migration" script to remove empty classifications instead? You are moving the validation code inside writer because of existing data, not because it makes more sense.

nucliadb/nucliadb/writer/resource/basic.py
```diff
@@ -188,3 +193,13 @@ def set_processing_info(bm: BrokerMessage, processing_info: ProcessingInfo):
     if processing_info.account_seq is not None:
         bm.basic.last_account_seq = processing_info.account_seq
     bm.basic.queue = bm.basic.QueueType.Value(processing_info.queue.name)
+
+
+def validate_classifications(paragraph: ParagraphAnnotation):
+    classifications = paragraph.classifications
+    if len(classifications) == 0:
+        raise ValidationError("ensure classifications has at least 1 items")
```
bloodbare, line 201: Why not just answer the HTTP exception as its on the writer ?

bloodbare (changes requested):

bloodbare (approved):

nucliadb/nucliadb/writer/resource/basic.py
lferran: Yes, that works too. 
My rationale was that `fastapi.HTTPException` is something that belongs to the "view" logic (the outermost layer of the writer), whereas the `parse_basic_modify` and `parse_basic` are more internal. I thought that having a custom `ValidationError` was making the `validate_classifications` function more reusable.

lferran:
> Do you think it might be a good idea to write a "migration" script to remove empty classifications instead? You are moving the validation code inside writer because of existing data, not because it makes more sense.

The validation code was already in the writer, but as part of the fastapi/pydantic layer. We decided to move it inside the writer views/endpoints so that fastapi doesn't validate that too on the output (while serializing the responses). 

About the migration: yes, that would be ideal. However, I don't think we have the right tools to do a migration at the resource level at the moment.

PR #542, opened 2023-01-30
Pagination broken after 100 results

jotare:
### Description
- Fix sorted search
- New bug appeared due to Rust side (sorting by creation/modification date for all resources don't work)

### How was this PR tested?
- Added and improved sorted search tests


ebrehault (approved):

nucliadb/nucliadb/search/api/v1/search.py
```diff
@@ -221,8 +220,17 @@ async def search(
                 ),
             )
     else:
-        if item.sort is not None and item.sort.limit is None:
-            item.sort.limit = SORTED_RELEVANT_SEARCH_LIMIT
+        if item.sort is None:
```
lferran, line 223: NIT: since the logic for validating the sort options is getting complex, I would move it into a separate function.

jotare: Done

bloodbare (approved):

PR #549, opened 2023-02-01
Create new shards on least loaded nodes

lferran:
### Description
This PR changes the way we choose the nodes when it comes to create a new shard.
We leverage the new `load_score` metric added by the nodes to simply find the least loaded ones.

We no longer need to use the rendervouz algorithm, which it doesn't make sense since the number of nodes changes over time (and the number of shards in the nodes too, due to deletions)

### How was this PR tested?
Unit test + existing tests should cover integration tests


lferran:
[ch3822]

PR #551, opened 2023-02-01
fix versions.json on reader

hermeGarcia:
### Description
Load `Versions` should use `from_deprecated` when the version file does not exits. So the reader can work until the writer creates the file.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_node/src/services/versions.rs
+++ b/nucliadb_node/src/services/versions.rs
@@ -27,6 +27,7 @@ const VECTORS_VERSION: u32 = 1;
 const PARAGRAPHS_VERSION: u32 = 1;
 const RELATIONS_VERSION: u32 = 1;
 const TEXTS_VERSION: u32 = 1;
+const DEPRECATED_CONFIG: &str = "config.json";
 
 #[derive(Serialize, Deserialize)]
 pub struct Versions {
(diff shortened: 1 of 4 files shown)


nucliadb_node/src/services/versions.rs
```diff
@@ -172,10 +172,20 @@ impl Versions {
     }
 
     pub fn load(versions_file: &Path) -> NodeResult<Versions> {
-        let versions_json = std::fs::read_to_string(versions_file)?;
-        let mut versions: Versions = serde_json::from_str(&versions_json)?;
-        versions.fill_gaps();
-        Ok(versions)
+        let deprecated_config = versions_file.parent().map(|v| v.join("config.json"));
```
jotare, line 175: Can you use a constant instead of hardcoding a string?

jotare (changes requested):

nucliadb_node/src/services/versions.rs
jotare: It also appears below

hermeGarcia: Good catch, fix at [ca38165](https://github.com/nuclia/nucliadb/pull/551/commits/ca38165967e7677f5f45a88b4ecc97078cf1546e). Also cleaned a lit bit the struct by defining a function for the duplicated functionality.

jotare (approved):

PR #552, opened 2023-02-01
get_or_create assumes get_kb will return None if no kb

ebrehault:
### Description

### How was this PR tested?


nucliadb_sdk/nucliadb_sdk/utils.py
```diff
@@ -88,6 +88,9 @@ def get_kb(
         headers={"X-NUCLIADB-ROLES": "READER"},
     )
 
+    if response.status_code == 404:
+        return None
```
jotare, line 92: Could you add the missing type annotation for the return type?

jotare (approved):

PR #553, opened 2023-02-01
tantivy order by field

hermeGarcia:
### Description
Ordering by Created and Modified.

### How was this PR tested?
Local tests.


nucliadb_protos/nodereader.proto
```diff
@@ -19,7 +19,11 @@ message OrderBy {
       DESC = 0;
       ASC = 1;
     }
-    string field = 1;
+    enum OrderField {
+        Created = 0;
+        Modified = 1;   
+    }
+    OrderField field = 1;
```
jotare, line 26: @bloodbare are we storing any protobuf with this field? Should we deprecate `field` and add code to be backwards compatible?

hermeGarcia: Done at [5eb00a5](https://github.com/nuclia/nucliadb/pull/553/commits/5eb00a5b82c3fea8e963f699253bdc0b2a4087ec)

PR #557, opened 2023-02-03
Fix streaming downloads

lferran:
### Description
`Content-Lenght` and `Content-Range` response headers were not properly set in all cases.

I fixed it by following these docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

### How was this PR tested?
Unit test
Local tests: I started the reader locally and checked that all the download combinations worked.

### DIFF
--- a/nucliadb/nucliadb/reader/tests/test_reader_file_download.py
+++ b/nucliadb/nucliadb/reader/tests/test_reader_file_download.py
@@ -218,6 +218,8 @@ FILE_SIZE = 10
         (f"bytes=0-{FILE_SIZE}", FILE_SIZE, 0, FILE_SIZE - 1, FILE_SIZE, None),
         # End range < file size
         (f"bytes=0-5", FILE_SIZE, 0, 5, 6, None),
+        # End range > file size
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/reader/tests/test_reader_file_download.py
```diff
@@ -206,17 +206,28 @@ async def _get_message_with_file(test_resource):
...
+        # No end range specified
+        ("bytes=0-", FILE_SIZE, 0, FILE_SIZE - 1, FILE_SIZE, None),
+        # End range == file size
+        (f"bytes=0-{FILE_SIZE}", FILE_SIZE, 0, FILE_SIZE - 1, FILE_SIZE, None),
+        # End range < file size
+        (f"bytes=0-5", FILE_SIZE, 0, 5, 6, None),
+        # End range > file size
+        (f"bytes=0-{FILE_SIZE + 1}", FILE_SIZE, 0, FILE_SIZE - 1, FILE_SIZE, None),
+        # Invalid range
+        ("bytes=something", FILE_SIZE, None, None, None, ValueError),
```
jotare, line 224: What happens in a multipart range request? The range is calculated as `end - start`, but the size of `Range: bytes=0-50, 100-150` is 100, not 150

jotare (approved):

PR #562, opened 2023-02-06
Improvements on node load reporting

lferran:
### Description
This PR includes a few improvements in the way we report the load of the nodes:
 - Changing the meaning of `load_score` reported by the nodes: now it's the total number of paragraphs.
 - Adding a new chitchat metadata named `shard_count`, reporting the total number of shards.
 - Removes the logic to push the load score to prometheus from the node writer.
 - Adds metrics reporting on ingest (master) about load score and shard count for the nodes
 
### How was this PR tested?
Extended integration and unit tests


nucliadb_node/src/bin/writer.rs
```diff
@@ -108,25 +108,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
     });
 
-    let report = NodeReport::new(host_key.to_string())?;
-
     let metrics_task = tokio::spawn(async move {
```
hermeGarcia, line 111: At some point it would be better to add this as static functions on the file instead of closures inside a huge main.

hermeGarcia (approved):
Just a comment!

hermeGarcia (changes requested):
The Python part should be looked at by someone else

nucliadb_node/src/bin/writer.rs
alekece: I agree.

alekece: Could be part of another PR, isn't it ?

hermeGarcia: yes, should be part of another PR

nucliadb_node/src/env.rs
```diff
@@ -225,63 +225,30 @@ pub fn get_sentry_env() -> &'static str {
...
-            None
-        }
-    }
-}
-
 /// Retuns the Promethus push timing, defaulted to 1h if not defined.
-pub fn get_prometheus_push_timing() -> Duration {
-    const DEFAULT_TIMING_PLACEHOLDER: &str = "1h";
-    const DEFAULT_TIMING: Duration = Duration::from_secs(60 * 60);
+pub fn get_metrics_update_interval() -> Duration {
```
bloodbare, line 229: What is this about ?

bloodbare (changes requested):
The cache memory structure is not in this PR ?

nucliadb_node/src/env.rs
alekece: The last part of the optimization is in this PR https://github.com/nuclia/nucliadb/pull/564 
And that free-function is removed

hermeGarcia (approved):

bloodbare (approved):

nucliadb/nucliadb/ingest/chitchat.py
```diff
@@ -133,7 +133,7 @@ async def close(self):
 JsonObject = Dict[str, JsonValue]
 
 
-def build_member_from_json(member_serial: JsonObject):
+def parse_load_score(member_serial: JsonObject) -> float:
```
jotare, line 136: I think it would be nice to refactor this obscure dict into a pydantic model. I've created a new ticket to do it: https://app.shortcut.com/flaps/story/3993

nucliadb/nucliadb/ingest/tests/servicer/test_list_members.py
```diff
@@ -34,3 +34,5 @@ async def test_list_members(grpc_servicer: IngestFixture):
 
     for member in response.members:
         assert member.type == Member.Type.IO
+        assert isinstance(member.load_score, float)
+        assert isinstance(member.shard_count, int)
```
jotare, line 38: When it could be a different type?

nucliadb/nucliadb/ingest/chitchat.py
alekece: Agree

nucliadb/nucliadb/ingest/tests/servicer/test_list_members.py
alekece: What do you mean ?

jotare: It's testing that load_score and shard_count are of types float and int. I was wondering if they can be of a different type. As the functions to get the value always return the correct value and type

lferran:
[ch3822]

PR #563, opened 2023-02-06
garbage collection implemented

hermeGarcia:
### Description
This PR defines a garbage collection strategy for the vectors index.
The writer process and the merge thread are synced through an atomic flag so the garbage collector does not erase a datapoint that is not garbage, but being created. 
Including a fix on the service interface to avoid calling`unwrap`.

### How was this PR tested?
Local tests


nucliadb_vectors/src/data_point_provider/merge_worker.rs
```diff
@@ -23,19 +23,20 @@ use nucliadb_core::fs_state;
...
 
 use super::merger::{MergeQuery, MergeRequest};
-use super::{State, VectorR};
+use super::work_flag::WorkFlag;
+use super::State;
 use crate::data_point::{DataPoint, DpId};
+use crate::VectorR;
 
-pub struct Worker(PathBuf);
+pub struct Worker(PathBuf, WorkFlag);
```
alekece: It could be create to turn that tuple struct to classic one and naming its fields for clarity purpose.

alekece (approved):

bloodbare (approved):

nucliadb_vectors/src/data_point_provider/merge_worker.rs
hermeGarcia: Done at [83d676f](https://github.com/nuclia/nucliadb/pull/563/commits/83d676fa207252d6791cc50184c5914ca8e4adbc)

PR #564, opened 2023-02-06
Optimize metrics calculation

alekece:
### Description
This PR aims to avoid go through all the list of shards to perform node load calculation.
To do so, the node metadata is calculated at startup then update dynamically on resource/shard update.

### How was this PR tested?
Ran locally.


nucliadb_node/src/writer/mod.rs
```diff
@@ -123,6 +140,10 @@ impl NodeWriterService {
             info!("Deleting {:?}", shard_path);
             std::fs::remove_dir_all(shard_path)?;
         }
+        self.metadata
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
```
hermeGarcia, line 145: Ditto

nucliadb_node/src/writer/mod.rs
```diff
@@ -213,10 +247,19 @@ impl NodeWriterService {
...
+
         let Some(shard) = self.get_mut_shard(shard_id) else {
             return Ok(None);
         };
+
         shard.remove_resource(resource)?;
+
+        metadata
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
```
hermeGarcia, line 260: Ditto

hermeGarcia (changes requested):

nucliadb_node/src/writer/mod.rs
```diff
@@ -113,6 +126,10 @@ impl NodeWriterService {
             relation_service: new_shard.relation_version() as i32,
         };
         self.cache.insert(shard_id, new_shard);
+        self.metadata
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
```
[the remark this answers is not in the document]
alekece: Done

hermeGarcia (commented):
Dont you think it would be better if the sending of metrics goes to the writer's grpc driver? I dont think the bindings should be affected by this. Also, there we can take advantage of the `async` environment

hermeGarcia (approved):

PR #569, opened 2023-02-08
Full text search with accents

lferran:
### Description
Fix fulltext search being tolerant to accents


bloodbare (changes requested):
Bloqued until proper discussion

bloodbare:
After discussion with @borinot we decided to pospone this functionallity as it works on the paragraph fuzzy search. Full text search will be accent sensitive

PR #571, opened 2023-02-08
Add metadata and label to relation serialization

lferran:
### Description
We need to serialize the metadata of the relation aswell, so that the front-end can display the apropriate entity.

### How was this PR tested?
Integration tests


lferran:
[ch3965]

hermeGarcia (approved):

PR #575, opened 2023-02-08
Simd vector similarity

hermeGarcia:
### Description
Simd cosine similarity + inlining of function

### How was this PR tested?
Local tests + performance analysis with `1m_stats`.

### DIFF
--- a/nucliadb_vectors/src/data_types/vector.rs
+++ b/nucliadb_vectors/src/data_types/vector.rs
@@ -41,8 +41,8 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
         .fold(encode_length(vec![], vec), encode_unit)
 }
 
-#[allow(unused)]
 #[inline]
+#[allow(unused)]
 pub fn simd_consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
     use wide::f32x8;
     const N: usize = 8;
@@ -87,8 +87,8 @@ pub fn simd_consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
     sum / (f32::sqrt(dem_x) * f32::sqrt(dem_y))
 }
 
-#[allow(unused)]
 #[inline]
+#[allow(unused)]
 pub fn consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
     let mut buff_x = [0; 8];
     let mut buff_y = [0; 8];
--- a/nucliadb_vectors/src/data_types/vector.rs
+++ b/nucliadb_vectors/src/data_types/vector.rs
@@ -43,7 +43,7 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
 
 #[inline]
 #[allow(unused)]
-pub fn simd_consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
+pub fn simd_cosine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
     use wide::f32x8;
     const N: usize = 8;
     let mut buff_x = [0; 8];
@@ -152,11 +152,11 @@ mod test {
         let v1_r = encode_vector(&v1);
         assert_eq!(
             naive_cosine_similatiry(&v0, &v1),
-            simd_consine_similarity(&v0_r, &v1_r)
+            simd_cosine_similarity(&v0_r, &v1_r)
         );
         assert_eq!(
             naive_cosine_similatiry(&v0, &v0),
-            simd_consine_similarity(&v0_r, &v0_r)
+            simd_cosine_similarity(&v0_r, &v0_r)
         );
     }
 }
--- a/nucliadb_vectors/src/data_types/vector.rs
+++ b/nucliadb_vectors/src/data_types/vector.rs
@@ -41,6 +41,52 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
         .fold(encode_length(vec![], vec), encode_unit)
 }
 
+#[allow(unused)]
+pub fn simd_consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
+    use wide::f32x8;
+    const N: usize = 8;
+    let mut buff_x = [0; 8];
+    let mut buff_y = [0; 8];
+    x.read_exact(&mut buff_x).unwrap();
+    y.read_exact(&mut buff_y).unwrap();
+    let len_x = Len::from_le_bytes(buff_x);
+    let len_y = Len::from_le_bytes(buff_y);
+    assert_eq!(len_x, len_y);
+    let len = len_x;
+    let mut buff_x = [0; 4];
+    let mut buff_y = [0; 4];
+    let mut sum = 0.0;
+    let mut dem_x = 0.0;
+    let mut dem_y = 0.0;
+    let mut i = 0;
+    let mut x_simd_buffer = [0.0; N];
+    let mut y_simd_buffer = [0.0; N];
+    let mut free_in_buffer = 0;
+    while i < len {
+        i += 1;
+        x.read_exact(&mut buff_x).unwrap();
+        y.read_exact(&mut buff_y).unwrap();
+        x_simd_buffer[free_in_buffer] = Unit::from_le_bytes(buff_x);
+        y_simd_buffer[free_in_buffer] = Unit::from_le_bytes(buff_y);
+        free_in_buffer += 1;
+        if free_in_buffer == N || (i == len && free_in_buffer > 0) {
+            let partial_simd_x = f32x8::new(x_simd_buffer);
+            let partial_simd_y = f32x8::new(y_simd_buffer);
+            let partial_dot_product = partial_simd_x * partial_simd_y;
+            let partial_power_x = partial_simd_x * partial_simd_x;
+            let partial_power_y = partial_simd_y * partial_simd_y;
+            dem_x += partial_power_x.reduce_add();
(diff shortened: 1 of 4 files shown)


nucliadb_vectors/src/data_types/vector.rs
```diff
@@ -41,6 +41,54 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
         .fold(encode_length(vec![], vec), encode_unit)
 }
 
+#[inline]
+#[allow(unused)]
+pub fn simd_consine_similarity(mut x: &[u8], mut y: &[u8]) -> Dist {
```
fulmicoton, line 46: cosine :D

nucliadb_vectors/src/data_types/vector.rs
```diff
@@ -41,6 +41,54 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
...
+        x_simd_buffer[free_in_buffer] = Unit::from_le_bytes(buff_x);
+        y_simd_buffer[free_in_buffer] = Unit::from_le_bytes(buff_y);
+        free_in_buffer += 1;
+        if free_in_buffer == N || (i == len && free_in_buffer > 0) {
+            let partial_simd_x = f32x8::new(x_simd_buffer);
+            let partial_simd_y = f32x8::new(y_simd_buffer);
+            let partial_dot_product = partial_simd_x * partial_simd_y;
+            let partial_power_x = partial_simd_x * partial_simd_x;
+            let partial_power_y = partial_simd_y * partial_simd_y;
+            dem_x += partial_power_x.reduce_add();
```
fulmicoton, line 79: you don't want to reduce here. I think it can be done out of the loop.

nucliadb_vectors/src/data_types/vector.rs
```diff
@@ -41,6 +41,54 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
...
+    let mut sum = 0.0;
+    let mut dem_x = 0.0;
+    let mut dem_y = 0.0;
+    let mut i = 0;
+    let mut x_simd_buffer = [0.0; N];
+    let mut y_simd_buffer = [0.0; N];
+    let mut free_in_buffer = 0;
+    while i < len {
+        i += 1;
+        x.read_exact(&mut buff_x).unwrap();
```
fulmicoton, line 68: I'd avoid using read_exact here... The compiler may or may not be smart enough to convert that into a simd unaligned load.

nucliadb_vectors/src/data_types/vector.rs
```diff
@@ -41,6 +41,54 @@ pub fn encode_vector(vec: &[Unit]) -> Vec<u8> {
...
+    const N: usize = 8;
+    let mut buff_x = [0; 8];
+    let mut buff_y = [0; 8];
+    x.read_exact(&mut buff_x).unwrap();
+    y.read_exact(&mut buff_y).unwrap();
+    let len_x = Len::from_le_bytes(buff_x);
+    let len_y = Len::from_le_bytes(buff_y);
+    assert_eq!(len_x, len_y);
+    let len = len_x;
+    let mut buff_x = [0; 4];
```
fulmicoton, line 57: is it 4 lane or 8 lane?

hermeGarcia: In that case is 4 bytes, is the buffer used to read the f32 values

nucliadb_vectors/src/data_types/vector.rs
hermeGarcia: Yes I had a version with that out of the loop, but the performance stayed the same, I will do that version again and update the code!

nucliadb_vectors/src/data_types/vector.rs
hermeGarcia: Im not sure I follow you on this.. Here im only reading the f32 value, that I have as bytes on x, then I add the f32 to the simd buffer.

nucliadb_vectors/src/data_types/vector.rs
hermeGarcia: fixed at [cd2f1a4](https://github.com/nuclia/nucliadb/pull/575/commits/cd2f1a4d6f80a2509115d44296278d443e9f2aec)

nucliadb_vectors/src/data_types/vector.rs
fulmicoton:
```
/// `left` & `right` do not need to be aligned.
/// This requires nightly due to portable SIMD,
/// and a cpu with AVX2 or equivalent to run.
/// f32 are assumed to be using native endianess
/// (most likely little endian in your case.)
///
/// AVX 512 should be 2x faster.
pub fn dot_product_simd(left: &[u8], right: &[u8]) -> f32 {
    const DIM: usize = 768;
    const NUM_LANES: usize = 8;
    assert_eq!(left.len(), DIM * 4);
    assert_eq!(right.len(), DIM * 4);
    let mut left_ptr = left.as_ptr() as *const f32x8;
    let mut right_ptr = right.as_ptr() as *const f32x8;
    let mut acc: f32x8 = Default::default();
    unsafe { //< ok thanks to the assert above.
        for i in 0..DIM / NUM_LANES {
            let mut left: f32x8 = std::ptr::read_unaligned(left_ptr);
            let right: f32x8 = std::ptr::read_unaligned(right_ptr);
            acc.add_assign(left.mul(right));
            left_ptr = left_ptr.offset(1 as isize);
            right_ptr = right_ptr.offset(1 as isize);
        }
    }
    acc.as_array().iter().copied().sum()
}
```

fulmicoton: (You need to add the sqrt root stuff or normalized your vector though)

hermeGarcia: I see, thank you so much! I am trying to adapt this to `nucliadb_vectors` where apart of adding  the sqrt and all that the dimension of the vectors is user defined..

fulmicoton: for dot product I get a x4 perf improvment with this.

fulmicoton: here is an implem for cosine...

```rust
pub fn cosine_simd(left: &[u8], right: &[u8]) -> f32 {
    const DIM: usize = 768;
    const NUM_LANES: usize = 8;
    assert_eq!(left.len(), DIM * 4);
    assert_eq!(right.len(), DIM * 4);
    let mut left_ptr = left.as_ptr() as *const f32x8;
    let mut right_ptr = right.as_ptr() as *const f32x8;
    let mut acc_left_norm: f32x8 = Default::default();
    let mut acc_right_norm: f32x8 = Default::default();
    let mut acc_dot_product: f32x8 = Default::default();
    unsafe { //< ok thanks to the assert above.
        for i in 0..DIM / NUM_LANES {
            let mut left: f32x8 = std::ptr::read_unaligned(left_ptr);
            let right: f32x8 = std::ptr::read_unaligned(right_ptr);
            acc_left_norm.add_assign(left.mul(left));
            acc_right_norm.add_assign(right.mul(right));
            acc_dot_product.add_assign(left.mul(right));
            left_ptr = left_ptr.offset(1 as isize);
            right_ptr = right_ptr.offset(1 as isize);
        }
    }
    let norm_left = acc_left_norm.reduce_sum().sqrt();
    let norm_right = acc_right_norm.reduce_sum().sqrt();
    acc_dot_product.reduce_sum() / (norm_left * norm_right)
}
```

fulmicoton:
```
scalar_dot              time:   [564.91 ns 574.43 ns 585.44 ns]
                        change: [-19.913% -16.021% -12.075%] (p = 0.00 < 0.05)
                        Performance has improved.

simd_dot                time:   [74.888 ns 76.182 ns 77.759 ns]
                        change: [+1.3111% +3.2754% +5.5679%] (p = 0.00 < 0.05)
                        Performance has regressed.

scalar_cosine           time:   [601.11 ns 602.38 ns 604.07 ns]
                        change: [+0.3208% +0.7946% +1.2116%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

simd_cosine             time:   [87.091 ns 87.313 ns 87.598 ns]
                        change: [-34.935% -34.503% -34.119%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

```

hermeGarcia: Thank you so much for the help! I am trying to figure out how would this behave in the case where a lane can not be filled, for examen if the dimension of the vectors is 5.

hermeGarcia: I added the following to the code: [398587e](https://github.com/nuclia/nucliadb/pull/575/commits/398587e4f9fd18cc1789112742e2b8a7308bb7d8)
Now I am seeing a better perfomance!

fulmicoton: can you share the bench? (and make sure you compile with target-cpu=native...)

hermeGarcia: I use [this binary](https://github.com/nuclia/nucliadb/blob/main/vectors_benchmark/src/binaries/1m_stats.rs) for testing the performance.

javitonino:
Superseded by https://github.com/nuclia/nucliadb/pull/2179

PR #583, opened 2023-02-14
Fixes INGEST-95 INGEST-96 when the knowledgebox does not exist anymore

sunbit:
### Description

### How was this PR tested?


nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -305,7 +300,7 @@ async def process(
             )
         elif self.audit is None:
             logger.warning("No audit defined")
-        elif audit_type is None:
+        elif audit_type is None and txn_result is not None:
```
lferran, line 303: what about the case when both are None?

sunbit: Not possible, because audit_type will be none if txn_result is none, as is derived from it

lferran (approved):

PR #589, opened 2023-02-14
Sent suggest audit events

sunbit:
Include client type info in search and suggest events Add the jaeger trace id to audit

### Description

### How was this PR tested?


nucliadb/nucliadb/search/api/v1/suggest.py
```diff
@@ -179,4 +183,14 @@ async def suggest_knowledgebox(
...
     if debug:
         search_results.shards = queried_shards
+
+    if audit is not None:
+        await audit.suggest(
+            kbid,
+            x_nucliadb_user,
+            x_ndb_client.to_proto(),
+            x_forwarded_for,
+            timeit - time(),
```
lferran, line 193: this should be
```python
time() - timeit
```

nucliadb/nucliadb/tests/test_search.py
```diff
@@ -952,3 +958,76 @@ async def test_only_search_calls_audit(nucliadb_reader, knowledgebox):
...
+
+    # Test suggest sends audit
+    resp = await nucliadb_reader.get(
+        f"/kb/{kbid}/suggest?query=", headers={"x-ndb-client": "chrome_extension"}
+    )
+    assert resp.status_code == 200
+
+    auditreq = await get_audit_messages(psub)
+
+    assert auditreq.kbid == kbid
```
lferran, line 1027: To be more confident, you could have assertions for all attributes of `auditreq`

nucliadb_models/nucliadb_models/search.py
```diff
@@ -31,10 +32,10 @@
...
 from nucliadb_models.resource import ExtractedDataTypeName, Resource
 
-if TYPE_CHECKING:
-    SortValue = OrderBy.OrderType.V
-else:
-    SortValue = int
+# if TYPE_CHECKING:
+#     ClientTypeValue = ClientType.V
+# else:
+#     ClientTypeValue = int
```
lferran, line 38: does this need to be uncommented back?

nucliadb_utils/nucliadb_utils/audit/basic.py
```diff
@@ -41,12 +41,23 @@ async def search(
...
     ):
         logger.debug(f"SEARCH {kbid} {user} {origin} ''{search}'' {timeit} {resources}")
 
+    async def suggest(
+        self,
+        kbid: str,
+        user: str,
+        client: int,
+        origin: str,
+        timeit: float,
```
lferran, line 58: what about `duration` as a better name?

nucliadb_utils/nucliadb_utils/audit/stream.py
```diff
@@ -175,12 +180,34 @@ async def search(
...
+
+    async def suggest(
+        self,
+        kbid: str,
+        user: str,
+        client_type: int,
+        origin: str,
+        timeit: float,
+    ):
+        # Search is a base64 encoded search
```
lferran, line 202: this comment does not belong here right?

lferran (changes requested):

nucliadb/nucliadb/search/api/v1/suggest.py
sunbit: wow good catch. There's at least one other place where this is wrong, not from this PR; i'll fix it

nucliadb_models/nucliadb_models/search.py
sunbit: i'll remove it,  i could't find a way to make that work with enums, it was just breaking in one point of another, so i typ ignored the two offending lines in the stream.py file

nucliadb_utils/nucliadb_utils/audit/basic.py
sunbit: Agree, let me check which consequences may have this change on the receiving side of this events

nucliadb_utils/nucliadb_utils/audit/stream.py
sunbit: yup, copy paste

nucliadb/nucliadb/search/api/v1/suggest.py
sunbit: Fixed

nucliadb_models/nucliadb_models/search.py
sunbit: Fixed

nucliadb_utils/nucliadb_utils/audit/stream.py
sunbit: Fixed

nucliadb_utils/nucliadb_utils/audit/basic.py
sunbit: We have some places that make use of this field, so i'll leave it like this for now.

nucliadb/nucliadb/tests/test_search.py
sunbit: from the POV of search, this are enough IMO, I agree that we may have unit tests on the `nucliadb_utils` for this that we don't right now, and we really need this to get rolling.

lferran (approved):

PR #590, opened 2023-02-14
Pagination doesn't work in resource search

jotare:
Story details: 
[sc-4011] https://app.shortcut.com/flaps/story/4011
[sc-4023] https://app.shortcut.com/flaps/story/4023

Added same pagination strategy we are doing in KB search

Added more exhaustive tests for pagination

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/resource.py
+++ b/nucliadb/nucliadb/search/api/v1/resource.py
@@ -84,9 +84,6 @@ async def search(
     range_creation_end: Optional[datetime] = None,
     range_modification_start: Optional[datetime] = None,
     range_modification_end: Optional[datetime] = None,
-    features: List[SearchOptions] = [
-        SearchOptions.PARAGRAPH,
-    ],
     reload: bool = Query(False),
     highlight: bool = Query(False),
     split: bool = Query(False),
@@ -99,12 +96,6 @@ async def search(
     debug: bool = Query(False),
     shards: List[str] = Query(default=[]),
 ) -> ResourceSearchResults:
-    if not all(map(lambda feature: feature == SearchOptions.PARAGRAPH, features)):
-        raise HTTPException(
-            status_code=422,
-            detail=f"Invalid search option. Only {SearchOptions.PARAGRAPH} is valid",
-        )
-
     if not rid:
         rid = await get_resource_uuid_by_slug(kbid, rslug, service_name=SERVICE_NAME)  # type: ignore
         if rid is None:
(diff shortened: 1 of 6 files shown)


nucliadb/nucliadb/search/search/query.py
```diff
@@ -178,6 +178,11 @@ async def paragraph_query_to_pb(
     request = ParagraphSearchRequest()
     request.reload = reload
     request.with_duplicates = with_duplicates
+
+    # We need to ask for all and cut later
+    request.page_number = 0
+    request.result_per_page = page_number * page_size + page_size
+
     if SearchOptions.PARAGRAPH in features:
```
lferran, line 186: Does it make sense to query the resource search and not have PARAGRAPH as requested feature?

nucliadb/nucliadb/search/api/v1/resource.py
```diff
@@ -86,8 +86,6 @@ async def search(
     range_modification_end: Optional[datetime] = None,
     features: List[SearchOptions] = [
         SearchOptions.PARAGRAPH,
-        SearchOptions.VECTOR,
```
lferran, line 89: Like this, if the user requests vectors or relations we will simply ignore it.
Have you considered returning a 4xx instead?

nucliadb/nucliadb/search/search/query.py
jotare: Yes, you can search by dates for example (although it's not implemented either (I'll add it now))

nucliadb/nucliadb/search/api/v1/resource.py
jotare: I've just removed the option, as we don't want to provide it

nucliadb/nucliadb/search/api/v1/resource.py
```diff
@@ -101,6 +99,12 @@ async def search(
     debug: bool = Query(False),
     shards: List[str] = Query(default=[]),
 ) -> ResourceSearchResults:
+    if not all(map(lambda feature: feature == SearchOptions.PARAGRAPH, features)):
+        raise HTTPException(
+            status_code=422,
+            detail=f"Invalid search option. Only {SearchOptions.PARAGRAPH} is valid",
+        )
```
lferran, line 106: If there is only one option, why providing the parameter at all?
I am in for removing the param and returning paragraphs always. Then we don't need this extra logic.
If we happen to offer more features in the future we can add it back easily.
WDYT?

jotare: Seems fine to me!

lferran (approved):

PR #591, opened 2023-02-14
Tracing and double lock fix

hermeGarcia:
### Description
Adding more logs on vector's merging and unnecessary lock during merge.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_vectors/src/data_point_provider/merge_worker.rs
+++ b/nucliadb_vectors/src/data_point_provider/merge_worker.rs
@@ -20,14 +20,14 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
-use nucliadb_core::fs_state;
-use tracing::*;
-
 use super::merger::{MergeQuery, MergeRequest};
 use super::work_flag::MergerWriterSync;
 use super::State;
 use crate::data_point::{DataPoint, DpId};
 use crate::VectorR;
+use nucliadb_core::fs_state;
+use std::sync::MutexGuard;
+use tracing::*;
 
 const SLEEP_TIME: Duration = Duration::from_millis(100);
 pub struct Worker {
@@ -47,7 +47,9 @@ impl Worker {
         })
     }
     fn merge_report<It>(&self, old: It, new: DpId) -> String
-    where It: Iterator<Item = DpId> {
+    where
+        It: Iterator<Item = DpId>,
+    {
         use std::fmt::Write;
         let mut msg = String::new();
         for (id, dp_id) in old.enumerate() {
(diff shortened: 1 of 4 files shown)


bloodbare (changes requested):
uauu looking forward to review tomorrow, @vangheem

nucliadb_vectors/src/data_point_provider/merge_worker.rs
```diff
@@ -64,16 +65,22 @@ impl Worker {
             tracing::info!("Could not request merge: {}", e);
         }
     }
-    fn work(&self) -> VectorR<()> {
-        while self.work_flag.try_to_start_working().is_err() {}
```
vangheem, line 68: hmmm, so it was just running forever here? This might have caused constant cpu spikes checking for lock then?

vangheem (changes requested):
I worry with locks that we don't have code that makes sure it is always released when the code blocks finish.

I know rust doesn't have try, finally behavior; however, there seem to be ways to emulate it that might be good.

For example, on line 87, we return and do not release the `work_flag`

nucliadb_vectors/src/data_point_provider/merge_worker.rs
hermeGarcia: `work_flag` is a mutex defined on the standard library. It has move semantics, so when it goes out of scope or the variable is not live anymore is dropped (therefore releasing the lock). You can look at the documentation [here](https://doc.rust-lang.org/src/std/sync/mutex.rs.html#178)

hermeGarcia: Is not running forever because is trying to acquire the mutex not blocking. This will stall in the case the garbage collection is running at that time, anyhow this line was completely useless because the lock is being dropped as soon as the while ends. You can see an example of this in the [following's code MIR](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021).

vangheem: Okay, I didn't mean forever that way--just until the lock is available it was previously running over and over until it was available. `while self.work_flag.try_to_start_working().is_err() {}`

In any case, you fixed it :)

vangheem: re locks, that makes more sense. And explicitly dropping them out of memory in the code block is an optimization because they would be unlocked automatically anyways when function finished executing then

hermeGarcia: ouh I see, yeah that was too consuming even though it is a separate thread. I though your concern was that two lines after that I try the same thing in the proper way and that may be causing a deadlock (Lucky for me Rust  came to the rescue in this case..).

vangheem (approved):

bloodbare (approved):

PR #594, opened 2023-02-15
Management of entities on KB via graph index

jotare:
### Description
- [x] Add new entry to maindb to store deleted entitiesgroups per KB
- [x] Entities API merge results from stored entities on maindb and indexed entities on node
- [x] Modify `EntitiesGroup` protobuffer to hold an enum status instead of a bool indicating if its merged
- [x] Index NERs from metadata and split metadata to relations index
- [x] Completely substitute and delete curator

### Do in the next iteration?
- [ ] Update entities API to support PATCH
- [ ] Add POST for entitiesgroup creation (instead of set operation)

### To do in a next iteration
- Proper indexing of SYNONYM relations (and removing of these relations when they change)

### How was this PR tested?
- Adding a new set of tests


nucliadb_protos/utils.proto
```diff
@@ -47,16 +47,17 @@ message RelationNode {
     string subtype = 6;
 }
 
-message JoinGraphCnx {
+message JoinGraphEdge {
```
lferran, line 50: Is this pb stored in maindb? if so, renaming could bring us problems with already created data?

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,261 @@
...
+                entities.setdefault(
+                    name, Entity(value=name, status=Entity.DiffStatus.NORMAL)
+                )
+
+        eg = EntitiesGroup(entities=entities, title=title, color=color, custom=custom)
+        return eg
+
+    async def index_entities(self, group: str, entities: EntitiesGroup):
+        # TODO properly indexing of SYNONYM relations
+        nodes = {
```
lferran, line 248: If I am not wrong, the `nodes` term is being used twice in this function with different meaning: one is for nodes of the graph and the one below for nodes of nucliadb.

What about `graph_nodes` to disambiguate?

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,261 @@
...
+        await self.txn.delete(entities_key)
+
+        deleted_groups_key = KB_DELETED_ENTITIES_GROUPS.format(kbid=self.kbid)
+        payload = await self.txn.get(deleted_groups_key)
+        deg = DeletedEntitiesGroups()
+        if payload is not None:
+            deg.ParseFromString(payload)
+        if group not in deg.entities_groups:
+            deg.entities_groups.append(group)
+        await self.txn.set(deleted_groups_key, deg.SerializeToString())
```
lferran, line 116: We should update the relations index upon deletions too, right?

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,261 @@
...
+
+    async def index_entities(self, group: str, entities: EntitiesGroup):
+        # TODO properly indexing of SYNONYM relations
+        nodes = {
+            i: RelationNode(
+                value=entity.value,
+                ntype=RelationNode.NodeType.ENTITY,
+                subtype=group,
+            )
+            for i, (name, entity) in enumerate(entities.entities.items())
```
lferran, line 254: shouldn't you filter out those that have been deleted?

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,261 @@
...
+        if stored_entities is None:
+            new_entities = entities
+        else:
+            merged = {}
+
+            for key, entity in stored_entities.entities.items():
+                if key not in entities.entities:
+                    entity.status = Entity.DiffStatus.DELETED
+                else:
+                    entity.status = entities.entities[key].status
```
lferran, line 81: mh, so we are delegating the responsability to decide the status (MERGED vs NORMAL) to the caller of the endpoint? I wonder if it would be simpler to do it here by comparing if the new one is equal to what we have stored.

nucliadb/nucliadb/ingest/maindb/local.py
```diff
@@ -73,7 +73,8 @@ async def save(self, key: str, value: bytes):
 
     async def remove(self, key: str):
         key = self.compute_path(key)
-        os.remove(key)
+        if os.path.exists(key):
```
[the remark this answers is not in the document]
jotare: Yes, I'll rename it to `path`

nucliadb/nucliadb/ingest/maindb/keys.py
```diff
@@ -0,0 +1,23 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+KB_ENTITIES = "/kbs/{kbid}/entities"
+KB_ENTITIES_GROUP = "/kbs/{kbid}/entities/{id}"
+KB_DELETED_ENTITIES_GROUPS = "/kbs/{kbid}/deletedentities"
```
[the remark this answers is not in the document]
jotare: It seems reasonable but some the first two keys are already in use and would need a migration. I can rename the variables without problem.

About "entity groups" it's not accurate, as they are groups of entities (many entities in each group)

nucliadb_protos/utils.proto
jotare: AFAIK it's not stored, in fact, I think it haven't been used anywhere yet. But correct me if I'm wrong @bloodbare

nucliadb/nucliadb/ingest/orm/entities.py
jotare: Completely agree

nucliadb/nucliadb/ingest/orm/entities.py
jotare: Mm... Good catch! I'm not sure about it. 

If the entities group is from the system, it definitely shouldn't be deleted from the index and we'll provide a view to the user. But for entities groups created by the user, might have sense to remove them from the index. @bloodbare what do you think?

If we decide to remove them, we have to be sure to only delete user-defined entities groups.

nucliadb/nucliadb/ingest/orm/entities.py
jotare: Merged in the entities context means two relations mean the same. Maybe it's not the best word to describe it. For example, I can have "house cat" and "domestic cat". In the UI I choose "domestic cat" to be the same as "house cat" and it'll set the following:
```
{
    "house cat": {"value": "house cat", "represents": ["domestic cat"], "merged": false},
    "domestic cat": {"value": "domestic cat", "represents": [], "merged": true},
}
```

nucliadb_node/tests/test_search_relations.rs
```diff
@@ -435,6 +435,34 @@ async fn test_search_relations_prefixed() -> Result<(), Box<dyn std::error::Erro
...
 
+    // --------------------------------------------------------------
+    // Test: prefixed search with node filters and empty query
+    // --------------------------------------------------------------
+
+    let response = reader
+        .relation_search(RelationSearchRequest {
+            shard_id: shard_id.clone(),
+            prefix: Some(RelationPrefixSearchRequest {
+                prefix: "".to_string(),
```
alekece, line 446: You can use `String::default` or `String::new` here

nucliadb_node/tests/test_search_relations.rs
```diff
@@ -435,6 +435,34 @@ async fn test_search_relations_prefixed() -> Result<(), Box<dyn std::error::Erro
...
+                node_filters: vec![RelationNodeFilter {
+                    node_type: NodeType::Entity as i32,
+                    node_subtype: Some("animal".to_string()),
+                }],
+            }),
+            ..Default::default()
+        })
+        .await?;
+
+    let expected = HashSet::from_iter(vec!["Cat".to_string()]);
```
alekece, line 456:
```suggestion
    let expected = HashSet::from_iter(["Cat".to_string()]);
```

alekece (commented):
Great work :cat: I just left some suggestions on the Rust part (and left a Python guys validating this PR, since almost all the changes are Python-related)

nucliadb/nucliadb/reader/api/v1/services.py
```diff
@@ -70,14 +70,10 @@ async def get_entities(
...
-                group,
-                preserving_proto_field_name=True,
-                including_default_value_fields=True,
-            )
-            if "" in group_dict["entities"]:
-                del group_dict["entities"][""]
-            response.groups[key] = EntitiesGroup(**group_dict)
+            entities_group = EntitiesGroup.from_message(group)
+            if "" in entities_group.entities:
+                del entities_group.entities[""]
```
lferran, line 75: Why do we need this? Can you leave a comment on the code for the future reader?

nucliadb/nucliadb/reader/api/v1/services.py
```diff
@@ -113,8 +110,14 @@ async def get_entity(request: Request, kbid: str, group: str) -> EntitiesGroup:
             )
         )
```
lferran, line 111: I guess you missed removing the second assignation to `response`. Also, do we need to filter the "" key from the entities key like above?

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,275 @@
...
+
+    async def mark_entities_group_as_deleted(self, group: str):
+        deleted_groups_key = KB_DELETED_ENTITIES_GROUPS.format(kbid=self.kbid)
+        payload = await self.txn.get(deleted_groups_key)
+        deg = DeletedEntitiesGroups()
+        if payload is not None:
+            deg.ParseFromString(payload)
+        if group not in deg.entities_groups:
+            deg.entities_groups.append(group)
+        await self.txn.set(deleted_groups_key, deg.SerializeToString())
```
lferran, line 221: Small optimization: the `txn.set` can be done inside the if clause right above.

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,275 @@
...
+    async def unmark_entities_group_as_deleted(self, group: str):
+        key = KB_DELETED_ENTITIES_GROUPS.format(kbid=self.kbid)
+        payload = await self.txn.get(key)
+        if payload is None:
+            return
+        deg = DeletedEntitiesGroups()
+        deg.ParseFromString(payload)
+        if group in deg.entities_groups:
+            deg.entities_groups.remove(group)
+        await self.txn.set(key, deg.SerializeToString())
```
lferran, line 232: same as above

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -366,45 +364,19 @@ async def del_labelset(self, id: str):
...
-        payload = await self.get_entitiesgroup_inner(group)
-        if payload is None:
-            eg = entities
-        else:
-            eg = EntitiesGroup()
-            eg.ParseFromString(payload)
-            eg.MergeFrom(entities)
-        entities_key = KB_ENTITIES_GROUP.format(kbid=self.kbid, id=group)
-        await self.txn.set(entities_key, eg.SerializeToString())
+        return await self.entities_manager.set_entities(group, entities)
```
lferran, line 373: I couldn't find where you do the txn.rollback() in case there is an exception. 
I guess it should be done on the writer view?

nucliadb/nucliadb/ingest/orm/entities.py
lferran: Was there a decision on this front?

jotare:
> Great work @jotare 👏🏼 I just have a few more comments.
> 
> Also, I know it wasn't there before, but maybe it's a good time to add exception handlings (with txn.abort()) in the ingest/service/writer.py grpc endpoints?

I think it will be better to do it in another PR to avoid this to get more complex

lferran (approved):

nucliadb/nucliadb/reader/api/v1/services.py
lferran: Not too important anyway... for me we can move on

nucliadb/nucliadb/ingest/orm/entities.py
lferran: We can always add this later. For me, we can move on with this PR

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -415,9 +415,16 @@ async def GetEntities(  # type: ignore
...
         entities_manager = EntitiesManager(kbobj, txn)
-        await entities_manager.get_entities(response)
-        response.kb.uuid = kbobj.kbid
-        response.status = GetEntitiesResponse.Status.OK
+        try:
+            await entities_manager.get_entities(response)
+        except Exception as e:
+            if SENTRY:
+                capture_exception(e)
+            traceback.print_exc()
```
lferran, line 422: logger.error instead of a print?

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -434,13 +441,22 @@ async def GetEntitiesGroup(  # type: ignore
...
         entities_manager = EntitiesManager(kbobj, txn)
-        entities_group = await entities_manager.get_entities_group(request.group)
-        if entities_group is None:
-            response.status = GetEntitiesGroupResponse.Status.ENTITIES_GROUP_NOT_FOUND
+        try:
+            entities_group = await entities_manager.get_entities_group(request.group)
+        except Exception as e:
+            if SENTRY:
+                capture_exception(e)
+            traceback.print_exc()
```
lferran, line 448: same

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -98,7 +99,13 @@ async def del_entities(self, group: str):
     async def get_entities_group_inner(self, group: str) -> EntitiesGroup:
         stored = await self.get_stored_entities_group(group)
         indexed = await self.get_indexed_entities_group(group)
-        if stored is not None and indexed is not None:
+        if stored is None and indexed is None:
+            # If an entitiesgroup appears without stored or indexed entities,
+            # most probably the node is reporting a node subtype with no nodes
+            # or a wrongwentitiesgroup is being searched
+            logger.warning("Suspicious entities group without entities: '{group}'")
```
lferran, line 106: you are missing the f'string here to be able to format it

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -0,0 +1,283 @@
...
+        entities_key = KB_ENTITIES.format(kbid=self.kbid)
+        async for key in self.txn.keys(entities_key, count=-1):
+            group = key.split("/")[-1]
+            if exclude_deleted and group in deleted_groups:
+                continue
+            yield group
+            visited_groups.add(group)
+
+        # indexed groups
+        node, shard_id = await self.kb.arbitrary_kb_node()
```
bloodbare, line 192: You need to iterate over all logical shards.

bloodbare (changes requested):

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -122,22 +129,50 @@ async def get_stored_entities_group(self, group: str) -> Optional[EntitiesGroup]
...
+            )
+            ops.append(node.reader.RelationSearch(request))
+
+        try:
+            results = await asyncio.wait_for(
+                asyncio.gather(*ops, return_exceptions=True),  # type: ignore
+                timeout=settings.relation_search_timeout,
+            )
+        except asyncio.TimeoutError as exc:
+            capture_exception(exc)
```
lferran, line 162: if SENTRY:

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -122,22 +129,50 @@ async def get_stored_entities_group(self, group: str) -> Optional[EntitiesGroup]
...
+            raise NodeError("Node unavailable for relation search") from exc
+
+        for result in results:
+            if isinstance(result, Exception):
+                capture_exception(result)
```
lferran, line 167: if SENTRY:

nucliadb/nucliadb/search/search/shards.py
```diff
@@ -35,6 +35,8 @@
 
 def query_shard(node: Node, shard: str, query: SearchRequest) -> SearchResponse:
     query.shard = shard
+    if query.relations:
```
bloodbare, line 38: Why we need this?

bloodbare (changes requested):

bloodbare (approved):

PR #596, opened 2023-02-15
Sidecar: add timeout on writer operations

lferran:
### Description

### How was this PR tested?


lferran:
> What happens if a setresource is > 3 min ?

@bloodbare in that case, the writer grpc client at the sidecar will timeout and the sidecar will retry it later on.

3 min was my initia proposal, but do you think the timeout should be bigger? Like.. 10 Minutes? At max it should be set to the nats timeout.

Right now I feel we don't have enough data to set an approrpiate value.

Which makes me think... we could add metrics on node operations.

vangheem:
We can start with no timeout(maybe with `-1` default setting) and make sure to get metrics on it before we enforce anything.

bloodbare:
Agreee with nathan , also we have metrics on jaegger

bloodbare:
As the bug has been fixed, we will pospone this PR until node refactor.

PR #598, opened 2023-02-16
Ingest disable metrics logs

lferran:
### Description
Prevents flooding stdout with the http access logs from prometheus collecting metrics

### How was this PR tested?
Tested it by starting ingest locally

### DIFF
--- a/nucliadb/nucliadb/ingest/metrics/service.py
+++ b/nucliadb/nucliadb/ingest/metrics/service.py
@@ -17,6 +17,8 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import logging
+
 from aiohttp import web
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/ingest/metrics/service.py
```diff
@@ -29,6 +31,10 @@
     PROMETHEUS = False
 
 
+# Do not flood stdout with /metrics access logs
+logging.getLogger("aiohttp.access").setLevel(logging.WARNING)
```
jotare, line 35: There are other endpoint access that will be hidden by this change on the log level?

lferran: AFAIK Ingest does not expose any http service other than the used for metrics collection, so we are safe.

jotare (approved):

PR #601, opened 2023-02-17
Create new shards on least loaded nodes

lferran:
### Description
This PR changes the way we choose the nodes where replicas for new shards are allocated.
We leverage the new  `shard_count` and `load_score` metrics (propagated by the nodes via chitchat) to simply find the least loaded ones.

We no longer need to use the [rendezvous algorithm](https://github.com/nuclia/nucliadb/blob/main/nucliadb_utils/nucliadb_utils/clandestined.py#L46), which is not useful when the number of nodes changes over time or when there can be arbitrary deletions of shards from nodes (due to KB deletions)

### How was this PR tested?
Unit tests for cluster and node
Existing tests should cover integration tests


lferran:
[ch3822]

nucliadb/nucliadb/ingest/orm/__init__.py
```diff
@@ -45,38 +44,44 @@ def __init__(self):
...
+        total_nodes = len(NODES)
+        if total_nodes < target_replicas:
+            raise NodeClusterSmall(
+                f"Not enough nodes. Total: {total_nodes}, Required: {target_replicas}"
+            )
 
-            return nodes[: settings.node_replicas]
-        else:
-            raise NodeClusterNotFound()
+        available_nodes = [
```
jotare, line 60: NIT: If all filtering conditions are met, we'll iterate the list of nodes 4 times + 1 sorting. Might be good to iterate less times and filters easier to understand (filtering by x[0] or x[1] is more difficult to read than node_id or node.shard_count)

lferran: The main reason to iterate the nodes multiple times was to be able to provide a different error message depending on the reason for not finding enough nodes.

I agree with the filtering being not too readable though. will change that

jotare (approved):

PR #606, opened 2023-02-17
Properly handle ClientDisconnect exceptions

jotare:
- ClientDisconnect exceptions are now managed in all our FastAPIs (reader, writer, search and train)
- Also removed old hack from versioned FastAPI not needed now


nucliadb_utils/nucliadb_utils/fastapi/exception_handlers.py
```diff
@@ -0,0 +1,43 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+from fastapi.responses import JSONResponse
+from sentry_sdk import capture_exception
+from starlette.requests import ClientDisconnect, Request
+
+
+async def global_exception_handler(request: Request, exc: Exception):
+    capture_exception(exc)
```
lferran, line 26: If I am not wrong, reporting to sentry should be opted-in

lferran (approved):

PR #609, opened 2023-02-21
metadata segment added to vectors schema

hermeGarcia:
### Description
Vectors need a `metadata` block. In this PR that space is taken into account in the schema.

### How was this PR tested?
Local tests + ensuring backwards compatibility.


nucliadb_protos/utils.proto
```diff
@@ -104,3 +104,8 @@ message UserVectorSet {
 message UserVectorsList {
     repeated string vectors = 1;
 }
+
+enum VectorSimilarity {
+    Cosine = 0;
+    Dot = 1;
```
jotare, line 110: To maintain a consistent style, use upper case in enums

nucliadb/nucliadb/ingest/tests/servicer/test_knowledgebox_servicer.py
```diff
@@ -89,6 +89,31 @@ async def test_create_knowledgebox(
...
+    pb.config.title = "My Title"
+    pb.config.similarity = utils_pb2.VectorSimilarity.Dot
+    result = await stub.NewKnowledgeBox(pb)  # type: ignore
+    assert result.status == knowledgebox_pb2.KnowledgeBoxResponseStatus.OK
+
+    kbid = knowledgebox_pb2.KnowledgeBoxID(slug="test-dot")
+    resp = await stub.GetKnowledgeBox(kbid)  # type: ignore
+    assert resp.config.similarity == utils_pb2.VectorSimilarity.Dot
+
+    # Check that by default we get cosine similarity
```
jotare, line 106: Split in two tests, as there are two things under tests (set specific similarity and show the default)

nucliadb_models/nucliadb_models/vectors.py
```diff
@@ -31,8 +34,38 @@ class UserVectorWrapper(BaseModel):
...
+            raise ValueError("Unknown similarity")
+
+    @classmethod
+    def from_pb(cls, message: PBVectorSimilarity.ValueType):
+        if message == PBVectorSimilarity.Cosine:
+            return cls("cosine")
+        elif message == PBVectorSimilarity.Dot:
+            return cls("dot")
+        else:
+            raise ValueError("Unknown similarity")
```
jotare, line 56: Use a map instead of two functions blown with if statements

nucliadb_models/nucliadb_models/resource.py
```diff
@@ -90,9 +95,18 @@ def id_check(cls, v: str) -> str:
                 raise ValueError("No uppercase ID")
             if char in "&@ /\\ ":
                 raise ValueError("Invalid chars")
-
         return v
 
+    @classmethod
+    def from_message(cls: Type[_T], message: PBKnowledgeBoxConfig) -> _T:
+        as_dict = MessageToDict(
```
jotare, line 102: Why are you using a different syntax than Vectorset.from_message?

nucliadb_node/binding/CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Changelog
 
+## 0.5.10
```
jotare, line 3: Isn't that a new feature? Why are you incrementing the third number instead of the second?

jotare (changes requested):

nucliadb_node/binding/CHANGELOG.md
lferran: 👍🏼 I thought we didn't really care about semantic versioning at this point

nucliadb_protos/utils.proto
lferran: done

nucliadb_protos/noderesources.proto
```diff
@@ -75,11 +75,8 @@ message ShardList {
 message EmptyResponse {}
 message EmptyQuery {}
 
-message VectorSentence {
-    repeated float vector = 1;
-}
 
-message ParagraphPosition {
+message Position {
```
jotare, line 79: Is this a breaking change. Are we storing a protobuf with this message anywhere?

nucliadb_protos/nodewriter.proto
```diff
@@ -54,9 +55,19 @@ message AcceptShardRequest {
     bool override_shard = 3;
 }
 
+message NewShardRequest {
+    utils.VectorSimilarity similarity = 1;
+    string kbid = 2;
```
jotare, line 60: NIT: why not `kbid = 1`?

nucliadb_vectors/src/data_point/mod.rs
```diff
@@ -53,6 +56,21 @@ impl DeleteLog for NoDLog {
...
 
+#[derive(Default, Debug, Clone, Copy, Serialize, Deserialize)]
+pub enum Similarity {
+    Dot,
+    #[default]
+    Cosine,
+}
+impl Similarity {
+    pub fn compute(&self, x: &[u8], y: &[u8]) -> f32 {
+        match self {
```
jotare, line 67: DOUBT: why are you using an enum and branching over all possible variations instead of having a trait and use concrete implementation for each similarity?

nucliadb_vectors/src/data_point/node.rs
```diff
@@ -149,19 +180,23 @@ mod tests {
     }
     const NO_LABELS: [&[u8]; 0] = [];
     const LABELS: [&[u8]; 3] = [b"L1", b"L2", b"L3"];
+    const NO_METADATA: Option<&[u8]> = None;
```
jotare, line 183: A variable named `NO_METADATA` makes me think about something *without* metadata (no metadata). Not sure about what it really means.

jotare (commented):
Nice job! Python side is okay for me. I let @alekece finish reviewing the Rust side

nucliadb_vectors/src/data_point/node.rs
hermeGarcia: good, is supposed to make you think about that. Is for testing, I use it for the vectors I want to index in the text without metadata.

nucliadb_protos/nodewriter.proto
hermeGarcia: I dont know, just the order I used. Is that relevant?

nucliadb_protos/noderesources.proto
lferran: AFAIK the message type name isn't part of the serialized binary in protobuf, so we should be fine:
https://protobuf.dev/programming-guides/encoding/#structure

According to the docs, we could even change the field name as long as we don't change the field type and the numeric field id.

nucliadb_protos/nodewriter.proto
jotare: No, it's just ordering

nucliadb_vectors/src/data_point/mod.rs
jotare: Is dynamic dispatch slower than a `match` statement for all possible `Similarity` variants?

hermeGarcia: Very much so. Dynamic dispatch requires for the compiler to include a table during runtime that figures out which method are you calling at each given time, and optimising that is actually harder for a compiler. On the other hand, a match expression can be very easy to optimised and actually is very cheap. 
Trait objects are very useful for adding ergonomics or writing code where the types that need to be handle are not a close set (because of user define types for example), but for code where the variants are clear I would say is almost always a good idea to use enums.

jotare: Mm... That's interesting!

bloodbare (changes requested):
similarity should be on the shard object

jotare (approved):

bloodbare (approved):

PR #616, opened 2023-02-23
Handle predict errors on search

lferran:
### Description
Will return HTTP206 partial response if semantic search is requested but no vectors can be converted due to predict api errors

### How was this PR tested?
Integration tests

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/search.py
+++ b/nucliadb/nucliadb/search/api/v1/search.py
@@ -373,7 +373,6 @@ async def search(
     response.status_code = 206 if incomplete_results else 200
     if check_missing_vectors(item, pb_query):
         response.status_code = 206
-        response.reason = "Vector not detected in query"
 
     if audit is not None and do_audit:
         await audit.search(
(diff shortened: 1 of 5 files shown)


nucliadb/nucliadb/search/api/v1/search.py
```diff
@@ -371,6 +372,9 @@ async def search(
     await abort_transaction()
 
     response.status_code = 206 if incomplete_results else 200
+    if check_missing_vectors(item, pb_query):
```
vangheem, line 375: hmm, okay since we don't have a great way of handling this, we're just assuming empty vector list translates to triton and then we do a 206

What do you think about pulling vectors into being a separate step here and not embedding into `global_query_to_pb`?

That way we can specifically handle the triton service issues instead of having to handle the error through a proxy object and config.

So maybe we'd only handle `SendToPredictError` in this function then?

vangheem (commented):
I have a comment about where we're handling predict errors and how we're assembling the query.

I just wonder about the layers we swallow and bubble errors.

Another option might be to have `global_query_to_pb` return a tuple `tuple[SomeResultType, SearchRequest]` so you then are just translating that SearchRequest response error type to a 206

nucliadb/nucliadb/search/api/v1/search.py
lferran: yes, thanks for the suggestions! Looks cleaner this way

bloodbare (changes requested):
Reviewed some changes to not split the global to pb function

bloodbare (approved):

PR #621, opened 2023-02-25
Automate build and versioning

vangheem:
### Description
The purpose of this PR is to remove the manual process parts of releases. When a new commit is merged, we simply want to make a new release without needing to think about dependency issues.

### details
- release new package for every merge to main
- pin all nucliadb dependencies for every package to prevent any inconsistencies
- auto increment version with build numbers
- no longer maintain in-code changelog
- move utils lru dependency so multi platform builds are no longer necessary

### Extraneous
- node bindings are NOT auto released
- telemetry NOT auto released

### How was this PR tested?
Manually. I can not test the actual uploads and it's possible there will be some thrashing as we get this under control.


.github/workflows/pypi_release.yml
```diff
@@ -0,0 +1,43 @@
+name: PyPI Releases
+
+on:
```
bloodbare, line 3: Should it be done only when any of the related jobs is done and sucessfully done ?

bloodbare (changes requested):
Great Job!

bloodbare (approved):

PR #623, opened 2023-02-27
Ingest: handle empty vectors on field_vectors

lferran:
### Description
Processing may send some field_vectors in the protobuf messages that point to a storage file that is empty or simply does not exist.

The goal of this PR is to handle this case.

### How was this PR tested?


lferran:
closing, as this has been fixed on the learning side

PR #624, opened 2023-02-27
fix prefix deletion

hermeGarcia:
### Description
When several prefixes are added to the deleted trie and some are prefixes between them, the time order is lost. In these cases the index may return vectors that are deleted.
The current fix does not require reindexing since the order is still there, it just needs to be interpreted correctly. This will start to happen as soon as this PR is merge. 
### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_vectors/src/data_types/dtrie_ram.rs
+++ b/nucliadb_vectors/src/data_types/dtrie_ram.rs
@@ -19,49 +19,41 @@
 //
 
 use std::collections::HashMap;
+use std::time::SystemTime;
 
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Serialize, Deserialize)]
-pub struct DTrie<Prop> {
-    value: Option<Prop>,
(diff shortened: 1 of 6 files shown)


nucliadb_vectors/src/data_types/dtrie_ram.rs
```diff
@@ -71,127 +63,123 @@ impl<Prop> DTrie<Prop> {
...
-                } else {
-                    self.crnt_trie
-                        .value
-                        .as_ref()
-                        .map(|v| (self.crnt_key.clone(), v))
-                }
-            }
-        }
+    pub fn prune(&mut self, time: SystemTime) {
+        self.inner_prune(time);
```
alekece, line 70: Is it still useful to have these both `inner_prune` and `prune` methods ?

hermeGarcia: `inner_prune` returns a bool that will not make any sense to the caller, is used by `inner_prune` to trim the trie.

alekece (approved):

PR #631, opened 2023-02-28
rename colaborators to collaborators

vangheem:
### Description

### How was this PR tested?


nucliadb_protos/resources.proto
```diff
@@ -84,7 +84,7 @@ message Origin {
     google.protobuf.Timestamp modified = 5;
     map<string, string> metadata = 6;
     repeated string tags = 7;
-    repeated string colaborators = 8;
+    repeated string collaborators = 8;
```
bloodbare, line 87: AFAIK this will break actual messages. Should be deprecated ? Lets discuss

bloodbare (changes requested):

nucliadb_protos/resources.proto
vangheem: Yes, this will break things. This reminds me that it'd probably be a good idea to setup protobuf linting to prevent us from introducing breaking changes.

vangheem: I re-added the field.

lferran (approved):

nucliadb_models/nucliadb_models/metadata.py
```diff
@@ -336,13 +337,13 @@ class Source(Enum):
...
-                message,
-                preserving_proto_field_name=True,
-                including_default_value_fields=True,
-            )
+        data = MessageToDict(
+            message,
+            preserving_proto_field_name=True,
+            including_default_value_fields=True,
         )
+        data["collaborators"] = data.pop("colaborators", [])
```
bloodbare, line 345: Maybe a comment for future devs to understand why ?

lferran (approved):

bloodbare (approved):

PR #639, opened 2023-03-03
Utils test coverage and testing updates

vangheem:
### Description
Some updates to our test setup and changes to how we structure nucliadb_utils tests

### updates
- remove global test coverage badge and replace with coverage separated by flags
- increase utils test coverage
- no cover sprinkled in where it didn't make sense to add coverage
- split integration and unit tests
- put tests into match test files that mirror test structure so it is easy for where where to write tests
- upgrade memorylru for corrected apple silicon support

## Proposals from this PR

- start separating unit from integration tests in other packages
- mirror module structure for tests so it is easy to know where to find and write new tests
- focus on getting every component to > 80% coverage. Once there, get everything > 90%


bloodbare (commented):
Lovely, maybe we should merge after monday promotion

bloodbare (approved):

PR #640, opened 2023-03-03
Hermegarcia/sc 2783

hermeGarcia:
### Description
On node shutdown, make chitchat aware of termination. Now the writer process will shutdown when a `SIGTERM`, `SIGQUIT` or `ctrl-c` operation occurs, but before shutting down the cluster will be notified.

### How was this PR tested?
Local tests.

### DIFF
--- a/nucliadb_node/src/bin/writer.rs
+++ b/nucliadb_node/src/bin/writer.rs
@@ -30,6 +30,7 @@ use nucliadb_cluster::{node, Key, Node, NodeHandle, NodeType};
 use nucliadb_core::protos::node_writer_server::NodeWriterServer;
 use nucliadb_core::protos::GetShardRequest;
 use nucliadb_core::tracing::*;
+use nucliadb_core::NodeResult;
 use nucliadb_node::env;
 use nucliadb_node::reader::NodeReaderService;
 use nucliadb_node::telemetry::init_telemetry;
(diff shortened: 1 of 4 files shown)


nucliadb_node/src/bin/writer.rs
```diff
@@ -80,28 +83,33 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+    nucliadb_telemetry::sync::start_telemetry_loop();
+    tokio::spawn(start_grpc_service(grpc_driver));
+    tokio::spawn(monitor_cluster(cluster_watcher));
+    tokio::spawn(watch_node_update(update_handle, receiver, node_metadata));
 
     info!("Bootstrap complete in: {:?}", start_bootstrap.elapsed());
     eprintln!("Running");
 
-    tokio::try_join!(writer_task, monitor_task, update_task)?;
-    telemetry_handle.terminate_telemetry().await;
```
vangheem, line 99: Is this doing cleanup for us? Should we still make sure all outstanding metrics/spans are flushed before exiting?

vangheem (approved):
Thank you!

nucliadb_node/src/bin/writer.rs
hermeGarcia: I do not see anything on the `try_join` documentation that suggest a cleanup in case of error in one of the tasks. Right now that line was just waiting forever. 
The tasks spawned are terminated when their runtime is dropped. Maybe adding a mechanism for ensuring the other tasks dont finish abruptly should be a matter of other PR?

PR #643, opened 2023-03-04
Fix bug where metadata was not handled correctly on grpc start_span_client

sunbit:
### Description
Not sure what is the intention of the metadata fiddling on the start_span_client, but anyhow it was broken when trying to parse a OrderedDict out of the metadata of the request.

This fixes it, but maybe we need to make sure what is the purpose of the code manipulating the metadata. We didn't hit that before, as we were not using metadata on any requests on servicers that are instrumented with nucliadb_telemetry.

### How was this PR tested?
Adde tests providing fake metadata to ensure code handling metadata is executed


bloodbare (changes requested):
VERSION change required

bloodbare (approved):

PR #645, opened 2023-03-06
Add more performance tests for read operations

jotare:
### Description
- Add performance tests for KB and fields read operations
- KB read endpoints use ingest ORM instead of gRPC API

### DIFF
--- a/nucliadb/nucliadb/reader/api/v1/knowledgebox.py
+++ b/nucliadb/nucliadb/reader/api/v1/knowledgebox.py
@@ -20,14 +20,10 @@
 from fastapi import HTTPException
 from fastapi_versioning import version  # type: ignore
 from google.protobuf.json_format import MessageToDict
-from nucliadb_protos.knowledgebox_pb2 import (
-    KnowledgeBox,
-    KnowledgeBoxID,
-    KnowledgeBoxPrefix,
-    KnowledgeBoxResponseStatus,
-)
 from starlette.requests import Request
 
+from nucliadb.ingest.orm.knowledgebox import KnowledgeBox
+from nucliadb.ingest.utils import get_driver
 from nucliadb.reader.api.v1.router import KB_PREFIX, KBS_PREFIX, api
 from nucliadb_models.resource import (
     KnowledgeBoxList,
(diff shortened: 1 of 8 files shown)


nucliadb/nucliadb/reader/api/v1/knowledgebox.py
```diff
@@ -50,14 +45,13 @@
...
+    txn = await driver.begin()
     response = KnowledgeBoxList()
-    async for kb_id in ingest.ListKnowledgeBox(KnowledgeBoxPrefix(prefix=prefix)):  # type: ignore
-        if kb_id.slug == "":
-            slug = None
-        else:
-            slug = kb_id.slug
-        response.kbs.append(KnowledgeBoxObjSummary(slug=slug, uuid=kb_id.uuid))
+    async for slug in KnowledgeBox.get_kbs(txn, prefix):
+        uuid = await KnowledgeBox.get_kb_uuid(txn, slug)
```
lferran, line 52: Is there a way to get slug and uuid at the same time from maindb to avoid having this nested lookup?

jotare: I don't think so. We'd need to do two lookups in any case I can think...

bloodbare: getKbs needs to answer the kbid also

bloodbare (changes requested):

nucliadb/nucliadb/reader/api/v1/knowledgebox.py
jotare: Done

lferran: Maybe it's not part of the initial scope of the PR anyway, but we could use the scan function to iterate key, value pairs (slug and uuid in our case), just like it is done in the docs: https://github.com/tikv/client-py/

jotare: I like it! But we should implement this for every transaction type (Redis and local). Do you want to do it in this PR or open a ticket?

lferran: for me it's out of the scope. We are good for now and if we ever see this is a problem, we can fix it by using scan

lferran (approved):

vangheem (approved):

bloodbare (approved):

PR #646, opened 2023-03-07
Do not store file passwords on maindb

lferran:
### Description
This PR prevents storing the passwords of password-protected files in the database. 
The passwords will only be send to processing so that we can download, open and process them.

### How was this PR tested?
Unit tests


nucliadb/nucliadb/writer/tests/unit/test_field.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
vangheem, line 1: should this file be in the `unit/resources/test_field` so we can mirror the full mod path?

nucliadb/nucliadb/writer/tests/unit/test_upload.py
```diff
@@ -0,0 +1,77 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
vangheem, line 1: and this one in `unit/api/v1/test_upload.py`?

vangheem (approved):

PR #647, opened 2023-03-07
Synonyms api

lferran:
### Description
This PR adds a new set of services for a Knowledgebox that allow to manage synonyms to use in search.

**DISCLAIMER**: It does NOT include the logic to search by synonyms. This will come in another PR

### How was this PR tested?
Unit and integration tests


lferran:
[ch4358]

nucliadb_protos/python/nucliadb_protos/audit_pb2.pyi
```diff
@@ -3,28 +3,21 @@
...
 import nucliadb_protos.nodereader_pb2
 import nucliadb_protos.resources_pb2
-import sys
 import typing
-
-if sys.version_info >= (3, 10):
-    import typing as typing_extensions
-else:
-    import typing_extensions
+import typing_extensions
```
jotare, line 14: We keep having issues with protos versions. Why are all changed again?

nucliadb/nucliadb/ingest/maindb/driver.py
```diff
@@ -85,3 +86,16 @@ async def keys(
         self, match: str, count: int = DEFAULT_SCAN_LIMIT, include_start: bool = True
     ) -> AsyncGenerator[str, None]:
         raise NotImplementedError()
+
+    @asynccontextmanager
+    async def managed_transaction(self) -> AsyncGenerator[Transaction, None]:
```
jotare, line 91: As an idea for renaming, maybe `transaction` is good enough. The clients will write code like this:
```
with driver.transaction() as txn:
    ...
```

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -435,6 +435,17 @@ async def del_widgets(self, widget: str):
...
+        pbsyn = await Synonyms(self.txn, self.kbid).get()
+        if pbsyn is not None:
+            synonyms.CopyFrom(pbsyn)
+
+    async def set_synonyms(self, synonyms: PBSynonyms):
+        await Synonyms(self.txn, self.kbid).set(synonyms)
+
+    async def delete_synonyms(self):
+        await Synonyms(self.txn, self.kbid).clear()
+
```
jotare, line 448: If every call will create a `Synonyms` instance and call a method, why not let the clients use Synonyms directly as a part of the ORM?

If you find it useful, maybe a `get_synonyms_klass` method following the idea of other classes on the ORM.

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -525,6 +528,74 @@ async def DelWidgets(self, request: DetWidgetsRequest, context=None) -> OpStatus
...
+                response.status = OpStatusWriter.Status.OK
+                return response
+            except Exception as e:
+                if SENTRY:
+                    capture_exception(e)
+                logger.exception(f"Errors setting synonyms")
+                response.status = OpStatusWriter.Status.ERROR
+                return response
+
+    async def DeleteSynonyms(  # type: ignore
```
jotare, line 576: Consider using `DelSynonyms` following the pattern of other functions inside the class

nucliadb/nucliadb/ingest/maindb/driver.py
```diff
@@ -85,3 +86,16 @@ async def keys(
...
+        """
+        Use to make sure transaction is always aborted
+        """
+        txn: Optional[Transaction] = None
+        try:
+            txn = await self.begin()
+            yield txn
+        finally:
+            if txn is not None:
+                await txn.abort()
```
jotare, line 101: If the trasaction has been commited, will it fail when trying to abort on TiKV?

nucliadb_protos/knowledgebox.proto
```diff
@@ -166,3 +166,13 @@ message VectorSet {
...
 }
+
+
+message TermSynonyms {
+    repeated string synonyms = 1;
+}
+
+
+message Synonyms {
+    map <string, TermSynonyms> synonyms = 1;
```
jotare, line 177: Maybe rename to `terms` instead of `synonyms`? I'm thinking about the client code doing:
```
for synonyms in synonyms_pb:
    for synonym in synonyms:
        ...
```

nucliadb/nucliadb/ingest/maindb/driver.py
lferran: good catch! It would not have failed, but we would have an unnecessary `await txn.abort()`

nucliadb_protos/python/nucliadb_protos/audit_pb2.pyi
lferran: I am at protobuff 4.22, seems like it's the most up to date

jotare (approved):

PR #658, opened 2023-03-13
Fix constant rebuild issue (Rust part)

alekece:
### Description
This PR aims to fix a problem we face for ages:
Any time we run a `cargo build` , `cargo test` or `cargo run --bin` at the root project, some workspace members are recompiled no matter what.
The problem is due to an invalid path in the build script in `nucliadb_protos/rust`.


jotare (approved):

vangheem (changes requested):
This will break python proto support.

Maybe we can find another way to do this that works better for both python and rust here.

vangheem (approved):
lgtm

PR #659, opened 2023-03-14
Ease imports for nucliadb datasets and fix sdk test

lferran:
### Description

### How was this PR tested?


nucliadb_sdk/nucliadb_sdk/tests/test_search_resource.py
```diff
@@ -269,7 +269,7 @@ def test_search_resource(knowledgebox: KnowledgeBox):
     resources = knowledgebox.search(
         vector=vector_q, vectorset="all-MiniLM-L6-v2", min_score=0.70
     )
-    assert len(resources.sentences.results) == 11
+    assert len(resources.sentences.results) == 20
```
alekece, line 272: Why the number of results change all of a sudden ?

lferran: Because of this fix from @hermeGarcia https://github.com/nuclia/nucliadb/commit/53bba67bd6a47e22bdfd8ba994f92ca4f1af0f38

nucliadb_dataset/nucliadb_dataset/__init__.py
```diff
@@ -37,3 +43,6 @@ class DatasetType(str, Enum):
 class ExportType(str, Enum):
     DATASETS = "DATASETS"
     FILESYSTEM = "FILESYSTEM"
+
+
+__all__ = ("NucliaDBDataset", "NucliaCloudDataset", "Task", "download_all_partitions")
```
vangheem, line 48: If we use `__all__`, we need to include other imports here.
```suggestion
__all__ = ("NucliaDBDataset", "NucliaCloudDataset", "Task", "download_all_partitions", "NUCLIA_GLOBAL", "CLIENT_ID", "DatasetType", "ExportType")
```

What I like to do is make sure that the package module root only ever imports from other places and maybe internal code never imports from it. This prevents circular import issues.

vangheem (changes requested):

nucliadb_sdk/nucliadb_sdk/tests/test_search_resource.py
alekece: Make sense :cat:

nucliadb_dataset/nucliadb_dataset/__init__.py
lferran: good catch. thanks! fixed it

vangheem (approved):

PR #662, opened 2023-03-14
Raise 419 conflict error on  duplicate resource slugs

lferran:
### Description
When creating resources, make sure writer API returns HTTP419 Conflict error response if attempting to create two resources in the same KB with the same slug.

### How was this PR tested?
Integration tests


vangheem (approved):
I know it's a pattern but I wonder if it's worth the indirection to have function named `resource_slug_exists` in 2 places like this.

For now, it's probably better to stick to consistency

PR #664, opened 2023-03-15
Add shutdown delay on I/O nodes

alekece:
### Description
This PR adds a shutdown delay on I/O nodes (a.k.a node reader and node writer) to fix 5xx errors during node restart.
When an I/O node receive termination signal, it will notify that the gRPC server will not serve anymore then sleep the main process for a period of time and then really shutdown.
For the node writer, it will also close the chitchat TCP/IP connection after receiving the termination signal and before sleeping in order to propagate the information over the network.

### How was this PR tested?
Ran I/O nodes locally, sent termination signals and performed some gRPC calls.

### DIFF
--- a/charts/nucliadb_node/values.yaml
+++ b/charts/nucliadb_node/values.yaml
@@ -45,7 +45,7 @@ storage:
 
 running:
   sentry_url:
-  shutdown_delay: "15s"
+  shutdown_delay: "5s"
 
 tracing:
   enabled: false
--- a/charts/nucliadb_node/values.yaml
+++ b/charts/nucliadb_node/values.yaml
@@ -45,7 +45,7 @@ storage:
 
 running:
   sentry_url:
-  shutdown_delay: "5s"
+  shutdown_delay: "15s"
 
 tracing:
   enabled: false
--- a/charts/nucliadb_node/values.yaml
+++ b/charts/nucliadb_node/values.yaml
@@ -45,6 +45,7 @@ storage:
 
 running:
   sentry_url:
+  shutdown_delay: "5s"
 
 tracing:
   enabled: false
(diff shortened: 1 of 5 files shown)


nucliadb_node/src/env.rs
```diff
@@ -258,3 +258,29 @@ pub fn get_cluster_liveliness_interval_update() -> Duration {
...
+
+                DEFAULT_SHUTDOWN_DELAY
+            }
+        }
+        Err(_) => {
+            warn!("{SHUTDOWN_DELAY_KEY} not defined. Defaulting to {SHUTDOWN_DELAY_PLACEHOLDER}");
+
+            DEFAULT_SHUTDOWN_DELAY
+        }
+    }
```
hermeGarcia, line 285: If we do not care about the error being return here I would have gone with: 
```rust
    let Ok(value) = env::var(SHUTDOWN_DELAY_KEY) else {
        warn!("{SHUTDOWN_DELAY_KEY} not defined. Defaulting to {SHUTDOWN_DELAY_PLACEHOLDER}");
        return DEFAULT_SHUTDOWN_DELAY
    };

    let Ok(duration) = parse_duration::parse(&value) else {
        error!(
            "{SHUTDOWN_DELAY_KEY} defined incorrectly. Defaulting to \
             {SHUTDOWN_DELAY_PLACEHOLDER}"
        );
        return DEFAULT_SHUTDOWN_DELAY
    };
    duration
```
Seems cleaner.

charts/nucliadb_node/values.yaml
```diff
@@ -45,6 +45,7 @@ storage:
 
 running:
   sentry_url:
+  shutdown_delay: "5s"
```
hermeGarcia, line 48: Are you sure the delay is 5s? I thought we were told is 15s.

hermeGarcia (changes requested):
Everything looks good! Just double check the shutdown delay.

hermeGarcia (approved):

vangheem (approved):

PR #668, opened 2023-03-16
Improve nucliadb node test performance

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/nucliadb_node/pull.py
+++ b/nucliadb_node/nucliadb_node/pull.py
@@ -70,12 +70,19 @@ class Worker:
     async def finalize(self):
         if self.gc_task:
             self.gc_task.cancel()
+
         for subscription in self.subscriptions:
             try:
                 await subscription.drain()
             except nats.errors.ConnectionClosedError:
                 pass
         self.subscriptions = []
+
+        try:
+            await self.nc.close()
+        except (RuntimeError, AttributeError):  # pragma: no cover
+            pass
+
         transaction_utility = get_transaction()
         if transaction_utility:
             await transaction_utility.finalize()
--- a/nucliadb_node/nucliadb_node/pull.py
+++ b/nucliadb_node/nucliadb_node/pull.py
@@ -81,6 +81,8 @@ class Worker:
         try:
             await self.nc.close()
         except (RuntimeError, AttributeError):  # pragma: no cover
+            # RuntimeError: can be thrown if event loop is closed
+            # AttributeError: can be thrown by nats-py when handling shutdown
             pass
 
         transaction_utility = get_transaction()
--- a/nucliadb_node/nucliadb_node/tests/fixtures.py
+++ b/nucliadb_node/nucliadb_node/tests/fixtures.py
@@ -21,7 +21,7 @@ import asyncio
 import os
 import tempfile
 import time
-from typing import AsyncIterable
+from typing import AsyncIterable, Iterable
 
 import docker  # type: ignore
 import pytest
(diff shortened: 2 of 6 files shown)


nucliadb_protos/python/nucliadb_protos/nodesidecar_pb2_grpc.pyi
```diff
@@ -12,6 +12,7 @@ from nucliadb_protos.noderesources_pb2 import (
     IndexMetadata as IndexMetadata,
     IndexParagraph as IndexParagraph,
     IndexParagraphs as IndexParagraphs,
+    NodeMetadata as NodeMetadata,
```
lferran, line 15: does it belong to this PR?

nucliadb_protos/python/nucliadb_protos/nodesidecar_pb2.pyi
```diff
@@ -18,6 +18,7 @@ from nucliadb_protos.noderesources_pb2 import (
     IndexMetadata as IndexMetadata,
     IndexParagraph as IndexParagraph,
     IndexParagraphs as IndexParagraphs,
+    NodeMetadata as NodeMetadata,
```
lferran, line 21: same: does this belong here?

nucliadb_node/nucliadb_node/tests/fixtures.py
```diff
@@ -214,16 +229,17 @@ def data_path():
 
 
 @pytest.fixture(scope="function")
-def shadow_folder(data_path):
+def shadow_folder(data_path) -> Iterable[str]:
```
lferran, line 232: NIT: typing would be simpler if we simply return the string

nucliadb_node/nucliadb_node/pull.py
```diff
@@ -70,12 +70,19 @@ def __init__(
...
         for subscription in self.subscriptions:
             try:
                 await subscription.drain()
             except nats.errors.ConnectionClosedError:
                 pass
         self.subscriptions = []
+
+        try:
+            await self.nc.close()
+        except (RuntimeError, AttributeError):  # pragma: no cover
```
lferran, line 83: Can you comment why we could have `RuntimeError` and `AttributeError`

nucliadb_protos/python/nucliadb_protos/nodesidecar_pb2_grpc.pyi
jotare: No, it's missing from management of entities. I probably forgot to rebuild protos after one of the last rebases

nucliadb_protos/python/nucliadb_protos/nodesidecar_pb2.pyi
jotare: Same as above

lferran (approved):
LGTM! Thanks for the cleanup!
Just a few comments

nucliadb_node/nucliadb_node/tests/fixtures.py
jotare: Yes, but as it's an async generator, mypy complains if you annotate only with `str`

lferran: What I meant is that you can change the `yield` to a return and then the annotation just a `str`

nucliadb_node/nucliadb_node/pull.py
jotare: They were handled on NATS utility. RuntimeError can occurr if the event loop is closed. I'm not sure about the AttributeError though. I'll try to know it and write down a comment

PR #674, opened 2023-03-20
Generic sentry integration

vangheem:
### Description
I would like to get rid of all instances of `if SENTRY`.

I was looking at adding `pragma: no cover` for these but realized we really should abstract this out at this point.

I feel like we should not be directly importing sentry_sdk anywhere in our code.

First we need to merge this and get a new release. Then another PR to update all instances of `if SENTRY`

### How was this PR tested?
unit tests


nucliadb_telemetry/nucliadb_telemetry/errors.py
```diff
@@ -0,0 +1,86 @@
...
+
+def setup_error_handling(version: str) -> None:
+    settings = ErrorHandlingSettings()
+
+    if settings.sentry_url:
+        enabled_integrations: list[Any] = []
+
+        if settings.logging_integration:
+            sentry_logging = LoggingIntegration(
+                level=logging.CRITICAL, event_level=logging.CRITICAL
```
lferran, line 76: Should level be a parameter? If I am not wrong, for some components we are receiving logger.ERRORs on sentry right now.

lferran (approved):
LGTM 👍🏼

nucliadb_telemetry/nucliadb_telemetry/errors.py
vangheem: This was a straight copy of the current implementation. I think I'll leave for now and wait until we actually need something else.

PR #679, opened 2023-03-20
Clean up sentry integration

vangheem:
### Description
The purpose of this PR is to clean up sentry integration.

Right now, sentry integration is managed with `if SENTRY` in many places.

Moreover, even just having the `sentry_sdk` import everywhere means it'll be more difficult if we ever decide we want to use something else for capturing errors.

[ch-4411]

### How was this PR tested?


lferran (approved):
lgtm! 👍🏼

PR #682, opened 2023-03-21
Fix ranged stream downloads downloads

lferran:
### Description
There was a case where we would set incorrect response headers, triggering some errors on downloads

### How was this PR tested?
Unit tests and local tests (starting the reader locally and trying out all different range combinations)


lferran:
[sc-4257]
[sc-3862]

vangheem (approved):
off by 1 error! :)

PR #683, opened 2023-03-21
remove prefix harcoded limit

hermeGarcia:
### Description
Prefix search had a hardcoded limit, this pr removes it.
Additionally we now use [`DocSetCollector`](https://docs.rs/tantivy/latest/tantivy/collector/struct.DocSetCollector.html#impl-Collector-for-DocSetCollector) for the prefix search instead of [`TopDocs`](https://docs.rs/tantivy/latest/tantivy/collector/struct.TopDocs.html).

### How was this PR tested?
Local tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1977,7 +1977,7 @@ dependencies = [
 
 [[package]]
 name = "nucliadb_node_binding"
-version = "0.7.1"
+version = "0.7.2"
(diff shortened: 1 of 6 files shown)


jotare (changes requested):
Bump a new version for `nucliadb_bindings`

jotare (approved):

PR #694, opened 2023-03-21
add metric utilities

vangheem:
### Description
The purpose of this PR is to add utilities for adding metrics to our components.

The main utility class is an `Observer`. It is often important to couple together a histogram with a counter. Counter is for rates of actions and histogram is for timings of those actions.

Example usage:

```
from nucliadb_telemetry import metrics

SEARCH_OBSERVER = metrics.Observer("search", labels={"type": "default"})

...

async def search(...):
    with SEARCH_OBSERVER() as metric:
        ...
        metric.labels["type"] = "special"
```

### How was this PR tested?
unit tests


nucliadb_telemetry/nucliadb_telemetry/metrics.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+__all__ = (
+    "Observer",
+    "ObserverRecorder",
+    "OK",
+    "ERROR",
+    "Counter",
+    "Histogram",
+    "Gauge",
```
lferran, line 137: These are implementation-specific (prometheus) and should not be exposed

lferran (approved):
Very useful ! LGTM except one minor comment

PR #697, opened 2023-03-22
add health checks for nodes in ingest and search

vangheem:
### Description
If no nodes are detected, the container should be considered unhealthy.

[ch-4624]

### How was this PR tested?
unit and manual

### DIFF
--- a/nucliadb/nucliadb/ingest/consumer/pull.py
+++ b/nucliadb/nucliadb/ingest/consumer/pull.py
@@ -193,7 +193,7 @@ class PullWorker:
             await self.nc.close()
 
     async def subscription_worker(self, msg: Msg):
-        if len(NODES) == 0:
+        while len(NODES) == 0:
             logger.warning(
                 "Waiting for nodes to be discovered through chitchat before processing any messages"
             )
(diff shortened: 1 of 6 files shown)


bloodbare (changes requested):
Shouldnt be also to restart all pod in case one container fails?

nucliadb/nucliadb/ingest/consumer/pull.py
```diff
@@ -192,6 +193,12 @@ async def finalize(self):
             await self.nc.close()
 
     async def subscription_worker(self, msg: Msg):
+        while len(NODES) == 0:
```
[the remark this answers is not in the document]
vangheem: for example, if there is actually a lag on startup to get the nodes, I wouldn't want a constant exit and never being able to start up

nucliadb/nucliadb/ingest/consumer/pull.py
```diff
@@ -192,6 +193,15 @@ async def finalize(self):
             await self.nc.close()
 
     async def subscription_worker(self, msg: Msg):
+        if len(NODES) == 0:
```
jotare, line 196: Could you add a comment here. It's not obvious that we are waiting to get node metadata for chitchat with a 1s sleep.

Furthemore, why adding an extra ack_wait (10s) and wait only 1s? The remaining 9s wouldn't be enough?

vangheem: yeah, this is silly. It should be a `while` and wait for it to become available.

The remaining 9s should be enough, however, the `in_progress` call resets the AckWait setting so it's like a new message until we're ready to work on it.

jotare (approved):

PR #700, opened 2023-03-23
Sidecar indexed stream

lferran:
### Description
This PR adds a logic to the node sidecar to publish indexed operations informations to a nats jetstream.
This will be used from ingest to know when a resource has been indexed in all nodes to:
  - Record audit/accounting stats
  - Return the http response to a user that requested the `x-synchronous: true` header
Both will be done in further PRs.

Notes on partitioning:
I have chosen to use the partition as part of the stream subject so that a particular ingest instance is able to subscribe to all changes for its own partition. Otherwise we would need to have 1 consumer for every ingest - sidecar pair.

### How was this PR tested?
Unit and integration tests


nucliadb_node/nucliadb_node/pull.py
```diff
@@ -291,3 +307,95 @@ async def subscribe(self):
...
+        self,
+        stream: str,
+        subject: str,
+        queue: str,
+        servers: List[str],
+        auth: Optional[str],
+    ):
+        self.stream = stream
+        self.subject = subject
+        self.queue = queue
```
vangheem, line 323: `self.queue` is not used. Maybe we should not have it as a property here?

nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -125,6 +125,10 @@ class IndexingSettings(BaseSettings):
     index_jetstream_auth: Optional[str] = None
     index_local: bool = False
 
+    indexed_jetstream_target: str = "indexed.{partition}"
+    indexed_jetstream_group: str = "indexed-{partition}"
```
vangheem, line 129: Since we're not consumer from this component, maybe we shouldn't have this setting?

nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -125,6 +125,10 @@ class IndexingSettings(BaseSettings):
     index_jetstream_auth: Optional[str] = None
     index_local: bool = False
 
+    indexed_jetstream_target: str = "indexed.{partition}"
```
vangheem, line 128: Shouldn't these settings be on the node? Maybe this whole settings class should be moved there?

vangheem (approved):
looks good, just a few comments.

you're choice on what to do

nucliadb_utils/nucliadb_utils/settings.py
lferran: yes I agree with you. To not make the dependency between node and utils worse, I will isolate the settings in the node

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -777,9 +777,12 @@ async def ReIndex(self, request: IndexResource, context=None) -> IndexStatus:  #
                 await kbobj.set_resource_shard_id(request.rid, shard.sharduuid)
 
             if shard is not None:
-                logger.info("Calling shard.add_resource")
-                counter = await shard.add_resource(brain.brain, 0, uuid.uuid4().hex)
-                logger.info("Finished shard.add_resource")
+                counter = await shard.add_resource(
+                    brain.brain,
+                    0,
+                    partition="reindex",
```
lferran, line 783: @bloodbare any objections with this? So if we ever do a reindex, we could subscribe to this partition to check when it has finished.

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -271,7 +271,7 @@ async def ProcessMessage(  # type: ignore
         async for message in request_stream:
             try:
                 await self.proc.process(
-                    message, -1, partition=0, transaction_check=False
+                    message, -1, partition="process", transaction_check=False
```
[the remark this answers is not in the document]
bloodbare: YOu know which partition you are, you are the partition, look inside yourself and you will find out.

nucliadb/nucliadb/ingest/service/writer.py
bloodbare: Feel the partition Luke

nucliadb_utils/nucliadb_utils/settings.py
lferran: done

nucliadb/nucliadb/ingest/service/writer.py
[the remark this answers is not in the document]
lferran: The thing is:
- Each ingest will be assigned to a set of partitions
- For each partition, it starts a pull worker (and each pull worker has its own processor)
- However, there is a single writer GRPC service, and its pull worker does not have partition set.

What seems more correct to me is to calculate the partition of every processed message based on the kbid and the resource. WDYT?

PR #703, opened 2023-03-23
fix paragraph lengths

vangheem:
### Description
off by 1 error

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/utils.py
+++ b/nucliadb/nucliadb/ingest/orm/utils.py
@@ -79,7 +79,9 @@ def set_title(writer: BrokerMessage, toprocess: PushPayload, title: str):
     etw.body.text = title
     writer.extracted_text.append(etw)
     fmw = FieldComputedMetadataWrapper()
-    paragraph = Paragraph(start=0, end=len(title), kind=Paragraph.TypeParagraph.TITLE)
+    paragraph = Paragraph(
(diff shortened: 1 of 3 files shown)


nucliadb_client/nucliadb_client/resource.py
```diff
@@ -166,15 +166,15 @@ def add_text(
             fcmw.metadata.split_metadata[split].paragraphs.append(
                 Paragraph(
                     start=0,
-                    end=len(text),
+                    end=len(text) - 1,
```
lferran, line 169: this one IMO is no big deal, as we are not using it that much and we should probably move client outside of the nucliadb repo

nucliadb/nucliadb/ingest/orm/utils.py
```diff
@@ -79,7 +79,9 @@ def set_title(writer: BrokerMessage, toprocess: PushPayload, title: str):
     etw.body.text = title
     writer.extracted_text.append(etw)
     fmw = FieldComputedMetadataWrapper()
-    paragraph = Paragraph(start=0, end=len(title), kind=Paragraph.TypeParagraph.TITLE)
+    paragraph = Paragraph(
+        start=0, end=len(title) - 1, kind=Paragraph.TypeParagraph.TITLE
```
lferran, line 83: ouch

vangheem:
I think the database behavior is correct and in other places on the platform it's off by 1

PR #705, opened 2023-03-23
Add more metrics

vangheem:
### Description
- search
- ingest consumer
- merge results
- tikv ops

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/search/search/shards.py
+++ b/nucliadb/nucliadb/search/search/shards.py
@@ -41,7 +41,7 @@ node_observer = metrics.Observer("node_client", labels={"type": ""})
 async def query_shard(node: Node, shard: str, query: SearchRequest) -> SearchResponse:
     query.shard = shard
     with node_observer({"type": "search"}):
-        return await node.reader.Search(query)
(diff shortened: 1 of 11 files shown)


nucliadb/nucliadb/search/search/shards.py
```diff
@@ -33,35 +33,47 @@
...
 
+node_observer = metrics.Observer("node_client", labels={"type": ""})
 
-def query_shard(node: Node, shard: str, query: SearchRequest) -> SearchResponse:
+
+async def query_shard(node: Node, shard: str, query: SearchRequest) -> SearchResponse:
     query.shard = shard
-    return node.reader.Search(query)
+    with node_observer({"type": "search"}):
+        return await node.reader.Search(query)  # type: ignore
```
lferran, line 44: Did this never work before as a coroutine? I wonder if we are using the method at all then

vangheem: Yup because it was still be awaited and the `Search` method still returned an awaitable result. Changing where we await here just allows us to observe the call.

lferran (approved):

PR #713, opened 2023-03-24
New Entities API

jotare:
### Description
- Deprecate POST to set entities groups
- Create two new endpoints POST and PATCH to create and update entities groups
- Add new methods on ingest to support this operations

### How was this PR tested?
- Unit tests at writer
- Integration tests at nucliadb

### Doubts and thoughts
- Return an error when creating an already existent entities group force us to perform at least a search per logical shard. Is it worth? Unit test added on writer but ingest doesn't implement it now
- How we should deal with entitiy deletions in relation to the index? Should we unindex entities? We'll need a method for this
- Indexing should be done of diffs, not all updates on an entities group


nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -90,6 +99,101 @@ async def set_entities(request: Request, kbid: str, group: str, item: EntitiesGr
...
+
+    pbrequest.delete.extend(item.delete)
+
+    set_info_on_span({"nuclia.kbid": kbid})
+
+    status: UpdateEntitiesGroupResponse = await ingest.UpdateEntitiesGroup(pbrequest)  # type: ignore
+    if status.status == UpdateEntitiesGroupResponse.Status.OK:
+        return
+    elif status.status == UpdateEntitiesGroupResponse.Status.KB_NOT_FOUND:
+        raise HTTPException(status_code=404, detail="Knowledge Box does not exist")
```
vangheem, line 188: Can we not raise exceptions for handled errors? Can you just return the http response?

Also, I think all our responses should be json

nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -90,6 +99,101 @@ async def set_entities(request: Request, kbid: str, group: str, item: EntitiesGr
...
+        entitypb.deleted = False
+        entitypb.represents.extend(entity.represents)
+
+    set_info_on_span({"nuclia.kbid": kbid})
+
+    status: NewEntitiesGroupResponse = await ingest.NewEntitiesGroup(pbrequest)  # type: ignore
+    if status.status == NewEntitiesGroupResponse.Status.OK:
+        return
+    elif status.status == NewEntitiesGroupResponse.Status.KB_NOT_FOUND:
+        raise HTTPException(status_code=404, detail="Knowledge Box does not exist")
```
vangheem, line 138: See other comment on raise http exceptions.

I think this will also product sentry errors for us.

vangheem (changes requested):
looks great!

nucliadb/nucliadb/writer/api/v1/services.py
jotare: Yes!
We can create a `Error` response model in `nucliadb_models`. Which fields do you think we need?

nucliadb/nucliadb/writer/api/v1/services.py
jotare: Same as above

nucliadb/nucliadb/writer/api/v1/services.py
jotare: I agree on changing exception raising with returning a response. I'll change it in this two endpoints and we can create a ticket to work on all other endpoints

nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -28,19 +28,28 @@
...
 from nucliadb.writer.resource.vectors import create_vectorset  # type: ignore
-from nucliadb_models.entities import EntitiesGroup
+from nucliadb_models.entities import (
+    CreateEntitiesGroupPayload,
+    EntitiesGroup,
+    UpdateEntitiesGroupPayload,
+)
 from nucliadb_models.labels import LabelSet
 from nucliadb_models.resource import NucliaDBRoles
+from nucliadb_models.responses import HTTPConflict, HTTPNotFound
```
vangheem, line 52: Should these just be in the `nucliadb` package? `nucliadb_models` is more for communicating types to external parties but we probably don't want or need it to have a `fastapi` dependency right?

vangheem (changes requested):

nucliadb/nucliadb/writer/api/v1/services.py
jotare: Moved to `nucliadb/models`

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -65,22 +65,30 @@ def __init__(
         self.txn = txn
         self.kbid = self.kb.kbid
 
+    async def create_entities_group(self, group: str, entities: EntitiesGroup):
+        # XXX: It's useful enough to check existence to return
+        # if await self.entities_group_exists(group):
+        #     raise AlreadyExists(f"Entities group {group} already exists")
+
```
lferran, line 72: If we decide to not do this check, I would remove the comment and all the related logic (try catch, etc). We can always add it back if we change our minds

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -502,25 +533,54 @@ async def SetEntities(self, request: SetEntitiesRequest, context=None) -> OpStat
...
-                entities_manager = EntitiesManager(kbobj, txn)
-                await entities_manager.del_entities(request.group)
-                await txn.commit(resource=False)
-                response.status = OpStatusWriter.Status.OK
+                await entities_manager.delete_entities_group(request.group)
             except Exception as e:
                 errors.capture_exception(e)
                 logger.error("Error in ingest gRPC servicer", exc_info=True)
                 response.status = OpStatusWriter.Status.ERROR
                 await txn.abort()
```
lferran, line 578: NIT: no need to abort if you are using the transaction context manager. It's still ok to be explicit though

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -502,25 +533,54 @@ async def SetEntities(self, request: SetEntitiesRequest, context=None) -> OpStat
...
+
     async def DelEntities(self, request: DelEntitiesRequest, context=None) -> OpStatusWriter:  # type: ignore
-        txn = await self.proc.driver.begin()
-        kbobj = await self.proc.get_kb_obj(txn, request.kb)
         response = OpStatusWriter()
-        if kbobj is not None:
+        async with self.proc.driver.transaction() as txn:
+            kbobj = await self.proc.get_kb_obj(txn, request.kb)
+            if kbobj is None:
+                await txn.abort()
```
lferran, line 567: NIT: no need to abort if you are using the transaction context manager. It's still ok to be explicit though

nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -90,6 +100,100 @@ async def set_entities(request: Request, kbid: str, group: str, item: EntitiesGr
         )
 
 
+@api.post(
+    f"/{KB_PREFIX}/{{kbid}}/entitiesgroups",
```
lferran, line 104: If you are adding new services, they need to add them to the virtual services yamls to be reachable in the hosted version of nucliadb

nucliadb_protos/writer.proto
```diff
@@ -151,12 +151,46 @@ message GetLabelsRequest {
...
     knowledgebox.KnowledgeBoxID kb = 1;
     string group = 2;
     knowledgebox.EntitiesGroup entities = 3;
 }
 
+message UpdateEntitiesGroupRequest {
+    knowledgebox.KnowledgeBoxID kb = 1;
+    string group = 2;
+    map<string, knowledgebox.Entity> add = 3;
+    map<string, knowledgebox.Entity> update = 4;
```
lferran, line 181: Can you add a comment indicating what is the expected value of the map's key?

nucliadb/nucliadb/writer/api/v1/services.py
jotare: If I'm right, they are already covered by this: https://github.com/nuclia/nucliadb/blob/bb88f30df81b095ffe227c4356ac723d3a07fca2/charts/nucliadb_writer/templates/writer.vs.yaml#L18

nucliadb_protos/writer.proto
jotare: They are the entity ids. Same as with EntitiesGroup entities map. I'll add a comment though

nucliadb/nucliadb/ingest/orm/entities.py
jotare: I'll uncomment it and enable the feature

nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -90,6 +100,100 @@ async def set_entities(request: Request, kbid: str, group: str, item: EntitiesGr
...
+
+    status: UpdateEntitiesGroupResponse = await ingest.UpdateEntitiesGroup(pbrequest)  # type: ignore
+    if status.status == UpdateEntitiesGroupResponse.Status.OK:
+        return
+    elif status.status == UpdateEntitiesGroupResponse.Status.KB_NOT_FOUND:
+        return HTTPNotFound(detail="Knowledge Box does not exist")
+    elif status.status == UpdateEntitiesGroupResponse.Status.ENTITIES_GROUP_NOT_FOUND:
+        return HTTPNotFound(detail="Entities group does not exist")
+    elif status.status == UpdateEntitiesGroupResponse.Status.ERROR:
+        raise HTTPException(
```
vangheem, line 192: Won't this cause sentry to trigger? Don't we want this to simply be a return instead of raise here?

nucliadb/nucliadb/writer/api/v1/services.py
```diff
@@ -90,6 +100,100 @@ async def set_entities(request: Request, kbid: str, group: str, item: EntitiesGr
...
+    if status.status == NewEntitiesGroupResponse.Status.OK:
+        return
+    elif status.status == NewEntitiesGroupResponse.Status.KB_NOT_FOUND:
+        return HTTPNotFound(detail="Knowledge Box does not exist")
+    elif status.status == NewEntitiesGroupResponse.Status.ALREADY_EXISTS:
+        return HTTPConflict(
+            detail=f"Entities group {item.group} already exists in this Knowledge box",
+        )
+    elif status.status == NewEntitiesGroupResponse.Status.ERROR:
+        raise HTTPException(
```
vangheem, line 145: here too

vangheem (changes requested):

vangheem (approved):

lferran (approved):

PR #714, opened 2023-03-25
improve naming and comments on resource shard management

vangheem:
### Description
When reviewing this code, I think this naming would have helped me understand what was going on more quickly.

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/abc.py
+++ b/nucliadb/nucliadb/ingest/orm/abc.py
@@ -56,11 +56,13 @@ class ShardCounter(BaseModel):
 
 class AbstractShard(metaclass=ABCMeta):
     @abstractmethod
-    def __init__(self, sharduuid: str, shard: PBShard, node: Optional[Any] = None):
+    def __init__(
+        self, sharduuid: str, shard: PBShard, node: Optional[Any] = None
+    ):  # pragma: no cover
         pass
 
     @abstractmethod
-    async def delete_resource(self, uuid: str, txid: int, partition: str):
+    async def delete_resource(self, uuid: str, txid: int, partition: str):  # pragma: no cover
         pass
 
     @abstractmethod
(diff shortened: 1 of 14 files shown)


nucliadb/nucliadb/ingest/orm/abc.py
```diff
@@ -50,23 +50,25 @@
 
 class ShardCounter(BaseModel):
     shard: str
-    fields: int
+    resources: int
```
lferran, line 53: mh I am not sure about this one.
Resources can have many fields mapped to it, so I think it makes sense to shard based on the number of fields

nucliadb/nucliadb/ingest/settings.py
```diff
@@ -88,7 +88,7 @@ class Settings(BaseSettings):
     sidecar_port_map: Dict[int, int] = {}
 
     # Node limits
-    max_node_fields: int = 200000
+    max_shard_fields: int = 200000  # max number of fields to target per shard
     max_node_shards: int = 600
```
lferran, line 92: Maybe that should be named `max_node_replicas` then?

From ingest point of view (and in my mind) a shard is a logic shard. A replica shard is the container unit from the node's point of view

lferran (approved):

nucliadb/nucliadb/ingest/settings.py
vangheem: it probably shouldn't be "max" either then right? It's the number that we always use. So just "shard_replicas"?

lferran: Just to clarify: `max_node_shards` here does not mean the number of replicas for a given shard.
It means the maximum number of shard replicas that a node can have

vangheem: yeah, I'm dumb. I see now. I think it makes sense

PR #725, opened 2023-03-28
Add metric to node sidecar subscriber

vangheem:


nucliadb_node/nucliadb_node/pull.py
```diff
@@ -52,6 +52,8 @@
     get_transaction,
 )
 
+consumer_observer = metrics.Observer("message_processor")
```
lferran, line 55: What about `worker_observer` or `subscriber_observer`? I know it is a consumer, but just to be consistent with naming. Otherwise, I'm ok with changing to `consumer` everywhere else in the module

vangheem: yeah, we use pull, worker, subscriber and now consumer... hmmm too many names

And I guess I'm just used to "consumer" so that why I introduced it here. I'll change.

lferran (approved):

PR #726, opened 2023-03-29
Small improvements

lferran:
### Description
- Ingest worker:
  -  Sentry event only if seqid is much older than the one stored

- Sidecar worker:
  - Cleanup storage only for set resource messages: we are not storing anything for deletion messages. this way we will avoid an unnecessary call to GCP that is resulting in a 404 error that is ignored on exception handling.
  
  - Log when indexed messages are published to stream

### How was this PR tested?
Unit tests


vangheem (approved):

nucliadb_node/nucliadb_node/pull.py
```diff
@@ -109,6 +108,7 @@ async def closed_cb(self):
         logger.info("Connection is closed on NATS")
 
     async def initialize(self):
+        self.storage = await get_storage(service_name=SERVICE_NAME)
```
jotare, line 111: IMO object attributes should be all defined in the constructor

vangheem: But it's an async func. We probably should add type definition for it at the top of the class though

PR #727, opened 2023-03-29
Fix indexed entity deletion

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/entities.py
+++ b/nucliadb/nucliadb/ingest/orm/entities.py
@@ -140,14 +140,19 @@ class EntitiesManager:
         await self.index_entities_group(group, entitiesgroup)
 
     async def delete_entities(self, group: str, delete: List[str]):
-        entities_group = await self.get_stored_entities_group(group)
-        if entities_group is None:
+        stored = await self.get_stored_entities_group(group)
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -140,14 +140,19 @@ async def set_entities_group_force(self, group: str, entitiesgroup: EntitiesGrou
...
+        if stored is None and indexed is None:
             return
+
+        stored = stored or EntitiesGroup()
+        indexed = indexed or EntitiesGroup()
         for name in delete:
-            if name in entities_group.entities:
-                entity = entities_group.entities[name]
+            if name in stored.entities or name in indexed.entities:
+                entity = stored.entities[name]
```
lferran, line 153: this could raise a KeyError if `name` is in `indexed` but not in `stored`?

lferran (approved):
LGTM! Only one comment

nucliadb/nucliadb/ingest/orm/entities.py
jotare: I don't think so, as they are protobufs, `stored.entities[name]` returns an empty `Entity` if the key doesn't exist. The conditional won't fail either.

PR #730, opened 2023-03-29
Node metrics, feature flagging and startup/shutdown improvements

vangheem:
### Description

- add node sidecar metrics for fs change events
- implement initial feature flagging system
- improve startup/shutdown handling
- improve prometheus integration across project

### How was this PR tested?


nucliadb_node/nucliadb_node/app.py
```diff
@@ -73,63 +60,47 @@ async def main() -> App:
...
+        reader.close,
+    ]
 
-    for task in to_cancel:
-        task.cancel()
+    if get_ff().enabled("nucliadb_node_sync_fs", default=False):
+        from nucliadb_node import write_sync
 
-    loop.run_until_complete(tasks.gather(*to_cancel, return_exceptions=True))
+        finalizers.append(write_sync.start())
```
lferran, line 98: Is a write_sync.start() a finalizer on purpose to sync any left unsynced changes?

nucliadb_node/nucliadb_node/tests/integration/test_indexing.py
```diff
@@ -26,36 +26,36 @@
...
 
-    pbshard: Optional[Shard] = await sidecar.reader.get_shard(sipb)
+    pbshard: Optional[Shard] = await reader.get_shard(sipb)
     if pbshard is not None and pbshard.resources > 0:
         processed = True
     else:
         processed = False
 
-    while processed is False:
+    while processed is False:  # pragma: no cover
```
lferran, line 56: Why do we need the pragma no cover here? Should we skip coverage for the whole tests folder?

nucliadb_node/nucliadb_node/app.py
vangheem: it could be eventually but for us and right now, it's only just to tell the coroutine to exit.

lferran (approved):

PR #745, opened 2023-03-31
Do not start pulling until there are some nodes on the cluster

bloodbare:
### Description

### How was this PR tested?


vangheem (commented):
we can remove the other place this is happening then I think. Better here!

vangheem (approved):

PR #749, opened 2023-03-31
Do not delete index data

vangheem:
### Description
Do not delete index data anymore.

This is so we can potentially use index data as an event source for replaying data.

### How was this PR tested?
unit/integration

### DIFF
--- a/nucliadb_node/nucliadb_node/pull.py
+++ b/nucliadb_node/nucliadb_node/pull.py
@@ -210,6 +210,7 @@ class Worker:
 
     async def set_resource(self, pb: IndexMessage) -> Optional[OpStatus]:
         brain: Resource = await self.storage.get_indexing(pb)
+        brain.shard_id = brain.resource.shard_id = pb.shard
         is_shadow_shard = self.ssm.exists(pb.shard)
         logger.info(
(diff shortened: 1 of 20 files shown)


bloodbare (approved):

nucliadb_node/nucliadb_node/tests/integration/test_indexing.py
```diff
@@ -118,18 +119,17 @@ async def test_indexing_shadow_shard(data_path, worker, shadow_shard: str):
     await asyncio.sleep(1)
 
     # Check that indexing messages have been deleted from storage
```
lferran, line 121: I think the comment can be removed: the assert is self-explanatory

lferran (approved):
lgtm! 👍🏼

nucliadb_node/nucliadb_node/pull.py
lferran, line 1: Should we fail hard in this case now??
https://github.com/nuclia/nucliadb/pull/749/files#diff-1090907991692180c21b752f642c8dd99cbe2c2dbc8fcbdc22ac68d299fb47b4L274

Also it would be nice to change the generic `KeyError` to something more explicit like `IndexingNotFound`

PR #758, opened 2023-04-03
add grpc metrics to telemetry

vangheem:
### Description

### How was this PR tested?


vangheem:
[ch-4850]

bloodbare (commented):
Can we keep the old names ? this library is used a lot on all the platform.

nucliadb_telemetry/nucliadb_telemetry/grpc_metrics.py
```diff
@@ -0,0 +1,99 @@
...
+
+class StreamUnaryClientInterceptor(aio.StreamUnaryClientInterceptor):
+    async def intercept_stream_unary(
+        self, continuation, client_call_details: ClientCallDetails, request_iterator
+    ):
+        metric = grpc_client_observer(labels={"method": client_call_details.method})
+        call = await continuation(client_call_details, request_iterator)
+        call.add_done_callback(functools.partial(finish_metric_grpc, metric))
+
+        return call
```
bloodbare, line 95: We also have UnaryStream GRPC services

vangheem: Okay, after thinking it over again, I figure we should just observe these as normal and if any additional custom instrumentation needs to be done, that can be done on a case-by-case basic.

PR #767, opened 2023-04-05
prometheus metrics

hermeGarcia:
### Description
This PR integrates prometheus into the node.

### How was this PR tested?
Local tests


nucliadb_core/src/context.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[cfg(log_metrics)]
+fn create_metrics() -> Arc<dyn Metrics> {
+    Arc::new(metrics::ConsoleLogMetrics)
+}
+
+#[cfg(not(any(prometheus_metrics, log_metrics)))]
+fn create_metrics() -> Arc<dyn Metrics> {
+    println!("NONE");
```
vangheem, line 41: Do we want this print here?

nucliadb_node/src/bin/reader.rs
```diff
@@ -47,17 +47,19 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     }
 
     let grpc_driver = NodeReaderGRPCDriver::from(node_reader_service);
-
-    tokio::spawn(start_grpc_service(grpc_driver));
+    let metrics_addr = SocketAddr::from(([127, 0, 0, 1], 3000));
```
vangheem, line 50: I'm pretty sure this will need to be `0.0.0.0` I think in order for it to be scrapable

vangheem (changes requested):
It's too bad it's so verbose every time we want to add a metric. I wish there was a way to just wrap a function or do the good ol python context manager approach :)

I have a few comments and I think you'll need to change the addr you bind to

vangheem (approved):

nucliadb_node/src/services/writer.rs
```diff
@@ -198,14 +197,13 @@ impl ShardWriterService {
         let result = ShardWriterService::initialize(id, path, metadata, tsc, psc, vsc, rsc);
         let metrics = context::get_metrics();
         let took = time.elapsed().map(|i| i.as_secs_f64()).unwrap_or(f64::NAN);
-        let metric = request_time::RequestTimeKey::shard(NAME.to_string());
+        let metric = request_time::RequestTimeKey::shard("writer/new".to_string());
```
hermeGarcia, line 200: Dont you thing is better for clarity to use a constant than to have the actual hardcoded value around the code? If is efficiency what your are concern about, the compiler ends up removing replacing constants for the value either way.

nucliadb_core/src/context.rs
hermeGarcia: ouch, thought I did remove it

Dockerfile.node
```diff
@@ -27,14 +27,15 @@ FROM eu.gcr.io/stashify-218417/basenode:latest as builder
 
 ARG CARGO_FEATURES=release-feature-set
 ARG CARGO_PROFILE=release
+ENV RUSTFLAGS="--cfg prometheus_metrics"
```
jotare, line 30: You should propagate this changes to Dockerfile.local_node, the version of Dockerfile.node that don't recall on our private basenode

charts/nucliadb_node/templates/node.sm.yaml
```diff
@@ -17,7 +17,16 @@ spec:
     matchLabels:
       app: node
   endpoints:
+  # node sidecar metrics
   - targetPort: {{ .Values.serving.metricsPort }}
     interval: 10s
     path: /metrics
+  # reader
+  - targetPort: 3031
```
jotare, line 25: Should we parametrize these too?

nucliadb_node/Makefile
```diff
@@ -25,3 +25,8 @@ test:
 .PHONY: test-cov
 test-cov:
 	pytest -rfE --cov=nucliadb_node --cov-config=../.coveragerc -s --tb=native -v --cov-report term-missing:skip-covered --cov-report xml nucliadb_node
+
+.PHONY: build
+build:
+	export RUSTFLAGS="--cfg=prometheus_metrics"
```
jotare, line 31: `export` doesn't take effect here, as commands are executed in different subshells. Use target-specific variables instead to make this env var local to the target:
https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html

```make
build: export RUSTFLAGS="--cfg=prometheus_metrics"
build:
        cargo build
```

nucliadb_node/src/bin/writer.rs
```diff
@@ -105,6 +106,7 @@ async fn main() -> NodeResult<()> {
     ));
     let update_task = tokio::spawn(update_node_state(update_handle, update_receiver));
     let monitor_task = tokio::spawn(monitor_cluster(cluster_watcher));
+    let metrics_task = tokio::spawn(run_http_metrics_server(3032));
```
jotare, line 109: Can we add this port number to settings?

nucliadb_node/src/services/reader.rs
```diff
@@ -93,11 +95,14 @@ impl ShardReaderService {
     }
     #[tracing::instrument(skip_all)]
     pub fn get_info(&self, request: &GetShardRequest) -> NodeResult<Shard> {
+        let span = tracing::Span::current();
```
jotare, line 98: If time we measure for `request_time` is always in the scope of a function, we might want to create a simple macro to decorate it and automatically handle time and metric. Similar to the observer decorator in the Python side

PR #768, opened 2023-04-06
Allow easy switch between maindb drivers

jotare:
### Description
Improve maindb fixtures and create a generic `maindb_driver` fixture using `redis_driver` as default. To change driver in a single file, simply override `maindb_driver` fixture:
```python
@pytest.fixture()
async def maindb_fixture(tikv_driver):
    yield tikv_driver
```
and every test using maindb will use TiKV

### How was this PR tested?
Validating all tests keep running correctly with the same driver

### DIFF
--- a/nucliadb/nucliadb/ingest/tests/integration/orm/test_orm_knowledgebox.py
+++ b/nucliadb/nucliadb/ingest/tests/integration/orm/test_orm_knowledgebox.py
@@ -49,35 +49,11 @@ def test_chunker():
     assert iterations == 0
 
 
-@pytest.fixture(scope="function")
-def tikv_driver_configured(tikv_driver):
-    from nucliadb.ingest.settings import settings
-    from nucliadb_utils.store import MAIN
-
-    prev_driver = settings.driver
-    settings.driver = "tikv"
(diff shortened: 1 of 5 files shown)


vangheem (approved):
very nice!

nucliadb/nucliadb/ingest/tests/integration/orm/test_orm_knowledgebox.py
```diff
@@ -49,35 +49,11 @@ def test_chunker():
...
-
-@pytest.fixture(scope="function")
-async def tikv_txn(tikv_driver):
-    txn = await tikv_driver.begin()
-    yield txn
-    await txn.abort()
-
-
 @pytest.mark.asyncio
 async def test_knowledgebox_delete_all_kb_keys(
```
lferran, line 53: to make sure this test is passing with the tikv driver as before, shouldn't you add a local fixture overwriting `main_driver`?

jotare: No, because this test is using `tikv_driver` directy. We can refactor it to use `get_driver` instead though

PR #769, opened 2023-04-06
mark processing errors on resources

vangheem:
### How was this PR tested?
unit and integration

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/processor.py
+++ b/nucliadb/nucliadb/ingest/orm/processor.py
@@ -491,19 +491,16 @@ class Processor:
             return
         txn = None
         try:
-            resource.txn = txn = await self.driver.begin()
-            resource.basic.metadata.status = PBMetadata.Status.ERROR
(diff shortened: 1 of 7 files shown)


nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -467,6 +467,44 @@ async def txn(
...
+                "Unable to mark resource as error, shard is None. "
+                "This should not happen so you did something special to get here."
+            )
+            return
+        if resource is None or resource.basic is None:
+            logger.info(
+                f"Skip when resource does not even have basic metadata: {resource}"
+            )
+            return
+        txn = None
```
lferran, line 492: FYW you can also use the `with self.driver.transaction() as txn:` context manager that will handle aborting for you

lferran (approved):

nucliadb/nucliadb/ingest/orm/processor.py
vangheem: Good point. Let me fix. It's a much better API!

PR #774, opened 2023-04-06
Revert "upgrade telemetry"

vangheem:
some weirdness, let's revert and see


vangheem:
bypass to merge and unblock

PR #777, opened 2023-04-10
fix tikv transaction leaks

vangheem:
### Description
- disambiguate `get_transaction` from `get_transaction_utility`
- remove duplicate `get_transaction` implementations
- add lock for acquiring transaction
- auto abort transaction on task cleanup -- this might be controversial!

### How was this PR tested?
unit and manual


bloodbare (approved):
Good catch

PR #779, opened 2023-04-10
structured logs support

vangheem:
### Description
Add structured logs support to nucliadb_telemetry

Structured logs allow us to add metadata into our logs that can more easily be queried in google cloud's interface.

### How was this PR tested?
unit


bloodbare (approved):

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+        if hasattr(record, "stack_info"):
+            data["stack_info"] = record.stack_info
+        else:  # pragma: no cover
+            data["stack_info"] = None
+
+    def extra_from_record(self, record):
+        return {
+            attr_name: record.__dict__[attr_name]
+            for attr_name in record.__dict__
+            if attr_name not in _BUILTIN_ATTRS
```
lferran, line 122: shouldn't this be `if attr_name in _BUILTIN_ATTRS` instead?

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+    def format(self, record: logging.LogRecord) -> str:
+        recordcopy = copy(record)
+        (
+            client_addr,
+            method,
+            full_path,
+            http_version,
+            status_code,
+        ) = recordcopy.args  # type: ignore[misc]
+        request_line = "%s %s HTTP/%s" % (method, full_path, http_version)
```
jotare, line 136: Now you're changing log format, I think it could be interesting to log request datetime

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+    root_logger = logging.getLogger()
+    access_logger = logging.getLogger(_ACCESS_LOGGER_NAME)
+    if not settings.dev:
+        root_logger.addHandler(json_handler)
+        access_logger.handlers = []
+        access_logger.addHandler(access_handler)
+    else:
+        root_logger.addHandler(stream_handler)
+
+    root_logger.setLevel(getattr(logging, settings.log_level.value))
```
jotare, line 181: Settings `LogLevel`s are coupled to logging log level attributes and it's not explicit. A change on the enum names can break this code.

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+    stream_handler = logging.StreamHandler()
+
+    root_logger = logging.getLogger()
+    access_logger = logging.getLogger(_ACCESS_LOGGER_NAME)
+    if not settings.dev:
+        root_logger.addHandler(json_handler)
+        access_logger.handlers = []
+        access_logger.addHandler(access_handler)
+    else:
+        root_logger.addHandler(stream_handler)
```
lferran, line 179: I am not sure, but will access logs not show up if `settings.dev = True`?

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+
+    root_logger.setLevel(getattr(logging, settings.log_level.value))
+
+    for logger_name, level in settings.logger_levels.items():
+        log = logging.getLogger(logger_name)
+        if logger_name != _ACCESS_LOGGER_NAME:
+            if not settings.dev:
+                log.addHandler(json_handler)
+            else:
+                root_logger.addHandler(stream_handler)
```
lferran, line 189: For when `dev = True` I am not sure this will work. If I am not wrong, will only see logs sent to the root logger in the stdout

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+    "processName",
+    "relativeCreated",
+    "stack_info",
+    "thread",
+    "threadName",
+)
+
+_OUTPUT_PROPERTIES = {
+    "filename": "filename",
+    "funcName": "funcName",
```
lferran, line 61: why only this one is in camelcase?

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -0,0 +1,192 @@
...
+            data["exc_info"] = self.formatException(record.exc_info)
+
+        for property, output_property in _OUTPUT_PROPERTIES.items():
+            data[output_property] = getattr(record, property, None)
+
+        data["logging.googleapis.com/sourceLocation"] = {
+            "file": record.pathname,
+            "line": record.lineno,
+            "function": record.funcName,
+        }
```
lferran, line 103: Why do we need this?
Isn't it already covered by `filename`, `lineno` and `funcName`?

nucliadb_telemetry/nucliadb_telemetry/logs.py
vangheem: nah, this is to gather properties that are NOT part of the record.

nucliadb_telemetry/nucliadb_telemetry/logs.py
vangheem: The request time is already logged as part of another field in the json log. Do we really need it here as well?

nucliadb_telemetry/nucliadb_telemetry/logs.py
vangheem: Because these are the standard property names. The only one I changed was `levelname` -> `severity` for gcp integration. Let me change so it's less confusing.

nucliadb_telemetry/nucliadb_telemetry/logs.py
vangheem: alright, tweaked, lmkwyt

nucliadb_telemetry/nucliadb_telemetry/logs.py
vangheem: all logs are propagated down to root by default

lferran (approved):

PR #780, opened 2023-04-10
more consistent script naming

vangheem:
### Description
For me, it doesn't make sense to have scripts that are named with "-metrics".

Let's just have 1 entrypoint and 1 way to deploy ndb.

Additionally, some cleanup to not do telemetry setup at module/global scope.

### How was this PR tested?

### DIFF
--- a/.github/workflows/nucliadb_reader.yml
+++ b/.github/workflows/nucliadb_reader.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Extract docs
         run: |
           mkdir -p /tmp/openapi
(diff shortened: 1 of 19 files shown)


bloodbare (approved):

lferran (approved):
LGTM!
Now that we are at it though, I wonder if the `nucliadb-` prefix is needed at all.

vangheem:
@lferran I like the prefixing. It namespaces us. Maybe we could do `ndb` instead of `nucliadb`?

Otherwise, these scripts get installed into the current env and `ingest`, `reader`, `writer`, etc feels odd then

PR #781, opened 2023-04-10
remove unused settings and remove duplicates

vangheem:
### Description

### How was this PR tested?


bloodbare (approved):

nucliadb_node/nucliadb_node/settings.py
```diff
@@ -36,13 +37,7 @@ class Settings(BaseSettings):
...
 
 
-class IndexingSettings(BaseSettings):
-    index_jetstream_target: Optional[str] = "node.{node}"
-    index_jetstream_group: Optional[str] = "node-{node}"
-    index_jetstream_stream: Optional[str] = "node"
-    index_jetstream_servers: List[str] = []
-    index_jetstream_auth: Optional[str] = None
-
+class IndexingSettings(utils_settings.IndexingSettings):
```
lferran, line 40: With this change the dependency to utils is a bit bigger at the node.
If we are not planning on getting rid of it any time soon, then I think it's OK

nucliadb/nucliadb/ingest/settings.py
```diff
@@ -39,9 +39,7 @@ def _missing_(cls, value):
                 return member
 
 
-class Settings(BaseSettings):
-    grpc_port: int = 8030
-
+class DriverSettings(BaseSettings):
     # allowing defaults is not ideal
     # TODO: implement settings without defaults
```
lferran, line 44: out of curiosity: why it is not ideal?

nucliadb_node/nucliadb_node/settings.py
vangheem: I thought any change would be moving to rust right? So for any python changes, the utils package is our tooling package to use for this sort of thing.

We need something like that for rust though probably

nucliadb/nucliadb/ingest/settings.py
vangheem: IMO, things that require configuration like this should be explicit and hard fail.

For example, right now, with default redis, you also need to configure the redis to use but a user could leave this setting unfilled and wonder what is wrong when they get a bunch of redis errors.

lferran (approved):

PR #785, opened 2023-04-11
Guess content type from filenames on uploads

lferran:
### Description

### How was this PR tested?


nucliadb/nucliadb/writer/api/v1/upload.py
```diff
@@ -754,3 +755,12 @@ def maybe_b64decode(some_string: str) -> str:
     except ValueError:
         # not b64encoded
         return some_string
+
+
+def guess_content_type(filename: str) -> str:
+    default_ct = "application/octet-stream"
```
jotare, line 761: NIT: I think `_ct` suffix doesn't bring value here. As it's a short function with a clear name, it's easy to understand `default` as default content type

nucliadb/nucliadb/writer/api/v1/upload.py
```diff
@@ -754,3 +755,12 @@ def maybe_b64decode(some_string: str) -> str:
...
+    try:
+        guessed_ct, _ = mimetypes.guess_type(filename)
```
jotare, line 763: When it can raise a `TypeError`?

lferran: It happens if for some reason filename is None. It should not happen though and it should be catched by mypy if we have such a situation. So I will remove it 👍🏼

jotare (approved):

PR #790, opened 2023-04-12
Fix purge command

lferran:
### Description

### How was this PR tested?


vangheem (changes requested):
hmmm, I don't think this is right.

I believe the `config_nucliadb` command was meant for only standalone mode.

at the bottom of `config_nucliadb` it also forces the usage of local node and for it to not work in cluster mode

We have 2 purge commands. Maybe this is only to fix the purge command that was meant to be used for standalone mode so we don't need this `standalone` option?

Did you mean to fix the ingest/purge.py?

PR #793, opened 2023-04-12
force reindex when labels change

vangheem:
### Description
Currently, when labels change, they do not cause all paragraphs to be updated. This forces reindex when a label is applied to a resource.

This did not previously happen with the `search` endpoint because it would utilize the fulltext index for a resource. Now, that the frontend is using the `find` endpoint for everything, we are searching by paragraph and not the resource at all.

### How was this PR tested?
integration test


bloodbare (approved):

vangheem:
[sc-4962]

PR #801, opened 2023-04-14
Fix token classifier NER computation 

jotare:
### Description
Fix token classifier NER computation

### How was this PR tested?
Unit tests

[sc-4995]


vangheem (approved):
I think the variable renaming is fine but it is a bit hard as a reviewer to see where the fix actually is!

PR #803, opened 2023-04-14
remove write sync metrics since they can not work properly

vangheem:
This approach that attempted to monitor is in a way that did not touch anything else did not work at all with the large number of files we have on a node.

We need to rethink what we're going to do here.


jotare (approved):
Why it didn't work?

vangheem:
Too many files kept borking it out :( We'll need to observe this in a different way.

PR #804, opened 2023-04-17
Add option to download files inline

lferran:
### How was this PR tested?
integration test


vangheem (approved):

jotare (commented):
How do we have less coverage?

PR #807, opened 2023-04-17
WIP: implement seeking to paragraph position support

vangheem:
### Description

### How was this PR tested?


nucliadb/nucliadb/search/search/merge.py
```diff
@@ -289,8 +295,13 @@ async def merge_vectors_results(
             index_int = int(index)
         except ValueError:
             index_int = -1
-        text = await get_text_sentence(
```
lferran, line 292: Is the new `get_paragraph_text` also compatible with what `get_text_sentence` was doing? I see there is some sketchy logic here: https://github.com/nuclia/nucliadb/blob/6d3a47ac28baeeae160d4e62655da4c0e8279abe/nucliadb/nucliadb/search/search/fetch.py#L150-L152

nucliadb/nucliadb/search/lifecycle.py
```diff
@@ -62,6 +63,8 @@ async def initialize() -> None:
     set_utility(Utility.NODES, NodesManager(driver=driver, cache=cache))
     await start_chitchat(SERVICE_NAME)
 
+    await paragraphs.initialize_cache()
```
lferran, line 66: where are you finalizing it?

nucliadb/nucliadb/search/search/paragraphs.py
```diff
@@ -0,0 +1,301 @@
...
+                key = await self.queue.get()
+                key = key + "_hits"
+                val = await self.redis.get(key)
+                if val is None:
+                    CACHE_OPS.inc({"type": "miss"})
+                else:
+                    CACHE_OPS.inc({"type": "hit"})
+                    CACHE_HIT_DISTRIBUTION.observe(int(val))
+                await self.redis.incr(key, 1)
+                await self.redis.expire(key, 60 * 10)
```
lferran, line 93: mh, maybe this value be bigger? since we don't have that much traffic yet...

lferran (commented):
looks great 👏🏼 only a few comments

nucliadb/nucliadb/search/search/paragraphs.py
vangheem: it's tough to balance because we don't want something too large and then we always get hits.

nucliadb/nucliadb/search/search/merge.py
vangheem: okay, for now, this is reverted just in case we don't understand this correctly.

PR #819, opened 2023-04-19
Fix problems with PATCH entitiesgroup

jotare:
### Description
Fix both issues

### How was this PR tested?
Integration tests

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/entities.py
+++ b/nucliadb/nucliadb/ingest/orm/entities.py
@@ -44,7 +44,11 @@ from nucliadb.ingest.maindb.keys import (
     KB_ENTITIES,
     KB_ENTITIES_GROUP,
 )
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -107,9 +107,12 @@ async def update_entities(self, group: str, entities: Dict[str, Entity]):
         intact. Use `delete_entities` to delete them instead.
 
         """
+        if not self.entities_group_exists(group):
+            raise KeyError(f"Entities group '{group}' doesn't exist")
```
lferran, line 111: Can you raise a more explicit exception? Otherwise we could be covering up a hidden KeyError bug when handling this one

lferran (approved):

PR #820, opened 2023-04-19
Improve ingest consumer logs

lferran:
### Description
Improve logging so that it is easier to debug ingest problems

### How was this PR tested?
No additional tests needed


lferran:
@vangheem partition is one of the things we need to add when we do the structured logs at ingest

jotare (approved):

vangheem (changes requested):

vangheem (approved):

PR #826, opened 2023-04-19
Refactor dummy nuclia engines

lferran:
### Description
Refactor test logic out of the engines to their own dummy class implementations.

### How was this PR tested?
- Existing tests should cover most of the changes
- Added full coverage of both engines with unit tests

### DIFF
--- a/nucliadb/nucliadb/config.py
+++ b/nucliadb/nucliadb/config.py
@@ -117,11 +117,11 @@ def config_nucliadb(nucliadb_args: Settings):
 
     if nucliadb_args.nua_api_key:
         nuclia_settings.nuclia_service_account = nucliadb_args.nua_api_key
-        nuclia_settings.disable_send_to_process = False
         ingest_settings.pull_time = 60
     else:
         ingest_settings.pull_time = 0
-        nuclia_settings.disable_send_to_process = True
+        nuclia_settings.dummy_processing = True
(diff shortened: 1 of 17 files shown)


vangheem (approved):
This is a nice cleanup -- it's good to not mix test code

I wonder, should we inject this dummy predict engine in rather than making it an explicit switch. Then, move the `DummyPredictEngine` class into a test module?

injection could either through monkey patching or dynamically loading a predict util factory from config value.

Would be nice to be able to remove all test code/config from the codebase if possible though.

lferran:
> This is a nice cleanup -- it's good to not mix test code
> 
> I wonder, should we inject this dummy predict engine in rather than making it an explicit switch. Then, move the `DummyPredictEngine` class into a test module?
> 
> injection could either through monkey patching or dynamically loading a predict util factory from config value.
> 
> Would be nice to be able to remove all test code/config from the codebase if possible though.

I totally agree!

However, the dummy implementation is being used from the sdk tests (that use the nucliadb docker image), see [here](https://github.com/nuclia/nucliadb/blob/ae20fe313b2eb22ba7b92e79279bf63a49e90868/nucliadb_sdk/nucliadb_sdk/tests/fixtures.py#L41)... So either we remove those tests or we are tied to having it in the codebase :/

PR #833, opened 2023-04-20
WIP: node scaling proposal: Index Coordinator

vangheem:
Alternative proposal after more thought has gone into our scaling designs and current implementation.


jotare (commented):
I really like how your proposal increase our code coverage :metal:

proposals/README.md
```diff
@@ -0,0 +1,11 @@
+# NIPs(NucliaDB Improvements Proposals)
```
jotare, line 1: Maybe you can do a PR with this file so you don't need to add it in each proposal

proposals/records/template.md
```diff
@@ -0,0 +1,23 @@
+# [Proposal Title]
```
jotare, line 1: Same for template

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+```mermaid
+flowchart TB
+    IndexNodeCoordinator-- CopyShardRequest -->SourceIndexNode
+    SourceIndexNode-- CopyShard -->DestinationIndexNode
+    IndexNodeCoordinator-- ValidateShard -->DestinationIndexNode
+```
+
+When an IndexNode needs it's replicas increased:
+
+- All indexing to shard needs to be buffered
```
lferran, line 88: do you think we can use shadow shards for buffering?

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+## Rollout plan
+
+### Phase 1: Implement side-by-side
+
+Phase 1 will focus on an approach with limited impact on the existing implementation.
+
+- Existing IndexNodes and architecture will stay in place
+- New IndexCoordinator component provided
```
lferran, line 144: Is the IndexCoordinator stateless? or would need to persist which IndexNodes it is managing somehow?

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+- New IndexNodes deployed seperately from existing nodes
+- Validate indexing, search approach, feature flag for test accounts
+
+### Phase 2: Cutover
+
+Once phase 1 is validated, cut over new scheme.
```
lferran, line 150: How do you imagine the cut over? Would it be kind of a migration script that assings old IndexNode data to IndexCoordinators?

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+
+The final phase will be putting the parts together for being able to rebalance shards.
+
+This will allow us to scale if we're read heavy on certain shards as well as scale down underutilized shards.
+
+## Key metrics
+
+- Coordinator search proxy performance
+- Index time
+- Replica write contention
```
lferran, line 168: What does this mean?

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+- Write lag
```
lferran, line 169: how do these two differ?

proposals/records/1-index-scaling.md
```diff
@@ -0,0 +1,183 @@
...
+
+This proposal focuses on moving the shard replica management, indexing and search load balancing
+to a seperate Coordinator layer.
+
+This proposal differs from https://github.com/nuclia/nucliadb/pull/722 in that it keeps all the
+same core concepts of the database engine in place while still allowing us to scale.
+
+
+*Goals*:
+- Allow us to dynamically scale reads/writes to real load, not shard size or count
```
lferran, line 11: What metric do you think we could use as real load?

proposals/records/1-index-scaling.md
vangheem: but we still need to finish playing all the data onto the shards that are being worked on so I'm not sure I see how shadow shard will help here.

proposals/records/1-index-scaling.md
vangheem: It would be persistent. It would be managing this state and also storing statistics on shards/node activity.

proposals/records/1-index-scaling.md
vangheem: In phase 1, we should index all the shards to the new coordinator nodes so this will just be a cutover. Everything should  already be validated as working.

proposals/records/1-index-scaling.md
vangheem: Updated. Is it more clear now?

proposals/records/1-index-scaling.md
vangheem: Added some points. Does that help?

PR #848, opened 2023-04-25
Refactor ingest processing txn logic

lferran:
### Description
- Fixed reindex logic (it was checking only the last message)
- Fixed code smell where `apply_resource` was optionally returning a tuple or None. Better handled via raising an exception.
- Encapsulate index resource logic in its own method (for better readability)

### How was this PR tested?
Existing tests should cover it

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/processor.py
+++ b/nucliadb/nucliadb/ingest/orm/processor.py
@@ -348,9 +348,6 @@ class Processor:
             partition, seqid, message.multiid, message.kbid, message.uuid
         )
 
-    def generate_index(self, resource: Resource, messages: List[BrokerMessage]):
-        pass
-
     async def txn(
         self, messages: List[BrokerMessage], seqid: int, partition: str
     ) -> Optional[TxnResult]:
@@ -372,9 +369,10 @@ class Processor:
         counter = None
         created = False
         shard: Optional[Shard] = None
-
+        reindex: bool = False
         try:
             for message in messages:
+                reindex = reindex or message.reindex
                 if resource is not None:
                     assert resource.uuid == message.uuid
                 result = await self.apply_resource(message, kb, resource)
@@ -388,11 +386,11 @@ class Processor:
             if resource:
                 await resource.compute_global_text()
                 await resource.compute_global_tags(resource.indexer)
-                if message.reindex:
+                if reindex:
                     # when reindexing, let's just generate full new index message
                     resource.replace_indexer(await resource.generate_index_message())
 
-            if resource and resource.modified:
+            if resource and (resource.modified or reindex):
                 shard_id = await kb.get_resource_shard_id(uuid)
                 node_klass = get_node_klass()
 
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -372,64 +371,38 @@ async def txn(
...
-                        similarity = await kb.get_similarity()
-                        shard = await node_klass.create_shard_by_kbid(
-                            txn, kbid, similarity=similarity
-                        )
-                else:
-                    raise AttributeError("Shard is not available")
+            if resource is None:
+                raise NoResourceError()
+
+            # TODO: in case of reindex = True, shouldn't we do that after generate_index_message?
```
vangheem, line 390: I think that should be right? I guess it was done this way to ensure we didn't miss anything.

nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -372,64 +371,38 @@ async def txn(
...
+            await resource.compute_global_text()
+            await resource.compute_global_tags(resource.indexer)
+            if reindex:
+                # When reindexing, let's just generate full new index message
+                resource.replace_indexer(await resource.generate_index_message())
+
+            if resource.modified or reindex:
+                # TODO: Shouldn't we index the resource after it has been commited?
```
vangheem, line 398: I agree. Technically we could have things in the index that failed to write.

nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -372,64 +371,38 @@ async def txn(
...
+                counter, shard = await self.index_resource(
+                    resource, txn, kb, uuid, seqid, partition
+                )
 
                 await txn.commit(partition, seqid)
 
+                # TODO: Look if it's really needed to commit the slug every time
```
vangheem, line 405: And what happens if slug commit fails but the other succeeded? This is weird here.

nucliadb/nucliadb/ingest/orm/processor.py
```diff
@@ -471,6 +448,47 @@ async def txn(
...
+                )
+            await kb.set_resource_shard_id(uuid, shard.sharduuid)
+
+        if shard is None:
+            raise AttributeError("Shard not available")
+
+        counter = await shard.add_resource(
+            resource.indexer.brain, seqid, partition=partition, kb=kb.kbid
+        )
+        if counter is not None and counter.fields > settings.max_shard_fields:
```
vangheem, line 485: While looking at this, I just want to mention that I wonder if we should change how we create new shards.

`add_resource` should not also being calling `GetCount` and returning a counter -- seems very odd

This could be on a separate asyncio queue reader so we're not doing this request during writes

vangheem (approved):
I think it's okay. I have some additional comments but everything should be okay.

lferran:
ok! I will address all these changes in a following PR (I don't want to add too much refactor at once)

PR #849, opened 2023-04-25
fix total count in pagraphs

hermeGarcia:
### Description
Currently `ParagraphSearchRequest` uses the field total as the number of matches that are returned.
This PR changes it to being the total number of matches found in the shard.

### How was this PR tested?
Local tests


lferran (approved):
Python side LGTM 👍🏼

PR #861, opened 2023-04-27
fix list resources in sdk

vangheem:
No tests here and the implementation response type was incorrect


nucliadb_sdk/nucliadb_sdk/client.py
```diff
@@ -222,11 +222,11 @@ async def async_del_resource(self, id: str):
         else:
             raise HTTPError(f"Status code {response.status_code}: {response.text}")
 
-    def list_resources(self):
+    def list_resources(self, page: int = 1):
```
jotare, line 225: Shouldn't be `page=0` the default?

jotare (approved):

nucliadb_sdk/nucliadb_sdk/client.py
vangheem: yes! Sorry. I'm used to 1-based paging.

PR #864, opened 2023-04-27
SDK v2

vangheem:
First step for v2 SDK implementation. This version focussed on HTTP API consistency and verbosity over facades over the APIs. Longer term, we can think about helpers to make some of the endpoint usage less verbose(however, helpers should be at argument level instead of sdk API level)

This implementation focuses on:
- automagical, make it easier for us to maintain, utilize our typing -- we don't want to have to deal with too much work around maintaining an SDK
- being consistent with API (if you read one, you can understand the other)
- lightly deprecates only API and updates all tests to use new API instead of old one
- DOES NOT attempt to maintain tests for both APIs
- consider this alpha version of API

What is NOT done:
- put deprecation warnings on old APIs
- type support -- this is more difficult that you would expect
- helpers to make the endpoints less verbose -- on purpose, let's wait here before we design something that we regret
- all the endpoints are NOT currently supported. If we like this approach, let's add support for everything.

Bottom line, I think this gives us a reasonable place to start. The end result is verbose but at least it is aligned with our http api. If our HTTP API is too verbose and difficult to understand, I don't think we should try and have the sdk make up for it.

### DIFF
--- a/nucliadb_sdk/nucliadb_sdk/v2/sdk.py
+++ b/nucliadb_sdk/nucliadb_sdk/v2/sdk.py
@@ -1,0 +1,119 @@
+import enum
+from typing import Any, Optional, Type, Union
+
+import httpx
+from pydantic import BaseModel
+
+from nucliadb_models.resource import Resource, ResourceList
+from nucliadb_sdk.v2 import exceptions
+
+
+class Region(enum.Enum):
+    EUROPE1 = "europe-1"
+    ON_PREM = "on-prem"
+
+
+def _request_builder(
+    path_template: str,
+    method: str,
+    path_params: dict[str, str],
+    request_type: Optional[Type[BaseModel]],
+    response_type: Optional[Type[BaseModel]],
+):
+    def _func(self: "NucliaSDK", content: Optional[Any] = None, **kwargs):
+        path_data = {}
+        for param in path_params:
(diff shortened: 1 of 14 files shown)


jotare (commented):
Nice work! I like how you have simplified things!

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -0,0 +1,290 @@
...
+        opts: dict[str, Any] = {}
+        if data is not None:
+            opts["data"] = data
+        if query_params is not None:
+            opts["params"] = query_params
+        response: httpx.Response = getattr(self.session, method.lower())(url, **opts)
+
+        if response.status_code < 300:
+            return response
+        elif response.status_code in (401, 403):
```
lferran, line 188: We are using 402 code for billing-related limit errors. Look at https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/ingest/processing.py#L290

or 

https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/search/predict.py#L182-L184

lferran (approved):
👏🏼 indeed much easier to maintain! just one comment

vangheem:
[sc-5450]

PR #865, opened 2023-04-28
Replace chitchat monitor for an HTTP connection

lferran:
### Description
- Removed the old implementation of a custom procotol over TCP to notify about cluster members to ingest / search.
Now we are using a simple HTTP request.

- Also using `nucliadb_telemetry.metrics` for available node metrics

### How was this PR tested?
Integration, unit and local tests


lferran:
[ch-5320]

hermeGarcia (approved):
LGTM

.github/workflows/nucliadb_one.yml
```diff
@@ -130,13 +130,17 @@ jobs:
...
-          context: .
-          file: Dockerfile.cluster_monitor
-          push: false
-          cache-from: type=gha
-          cache-to: type=gha,mode=min
-          tags: eu.gcr.io/stashify-218417/cluster_manager:main
+        run: |
+          docker build -t eu.gcr.io/stashify-218417/cluster_manager:main -f Dockerfile.cluster_monitor .
+
+        # TODO: un-comment when we sort out image cache issues
```
vangheem, line 136: Just remove this now. We don't need the commented code as we have other examples on how to set it up. Also can you fix the 2 above this?

vangheem (changes requested):

vangheem (approved):

PR #870, opened 2023-04-28
Detect icon value from text fields automatically

lferran:
### Description

### How was this PR tested?
- [x] todo tests


vangheem (approved):
I think this approach is fine but let's try to start thinking about a way to build better patterns around parsing and mapping types to desired values.

I feel like these functions are getting quite large and have a lot of functionality packed into them.

PR #873, opened 2023-05-01
Ingest Updates

vangheem:
### Description

- Decouple pull workers from ingest consumers
- Move to constants for stream configuration
- Reuse nats connections between consumers instead of having a unique connection for every one
- Be able to split out processed messages writes from regular transaction writes
- Decouple ingest write seq id tracking from the db transaction handling

### How was this PR tested?
Unit, integration and pain

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/processor/__init__.py
+++ b/nucliadb/nucliadb/ingest/orm/processor/__init__.py
@@ -41,6 +41,7 @@ from nucliadb.ingest.orm.exceptions import (
     SequenceOrderViolation,
 )
 from nucliadb.ingest.orm.knowledgebox import KnowledgeBox
+from nucliadb.ingest.orm.processor import sequence_manager
 from nucliadb.ingest.orm.resource import Resource
 from nucliadb.ingest.orm.shard import Shard, ShardCounter
 from nucliadb.ingest.orm.utils import get_node_klass, set_basic
@@ -70,6 +71,18 @@ AUDIT_TYPES: Dict[TxnAction, int] = {
 
 
 class Processor:
+    """
+    This class is responsible for processing messages from the broker
+    and attempts to manage sequencing correctly with a txn id implementation.
+
+    The "txn" in this implementation is oriented around the sequence id of
+    messages coming through the message broker.
+
+    Not all writes are going to have a transaction id. For example, writes
+    coming from processor can be coming through a different channel
+    and can not use the txn id
+    """
+
     messages: Dict[str, List[BrokerMessage]]
 
     def __init__(
@@ -86,16 +99,6 @@ class Processor:
         self.audit = audit
         self.partition = partition
         self.cache = cache
-
-    async def initialize(self):
-        await self.driver.initialize()
-        if self.cache is not None:
-            await self.cache.initialize()
-
-    async def finalize(self):
-        await self.driver.finalize()
-        if self.cache is not None:
-            await self.cache.finalize()
 
     @staticmethod
     def iterate_auditable_fields(resource_keys, message):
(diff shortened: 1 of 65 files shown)


vangheem:
[sc-5275]

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -277,6 +280,11 @@ async def process(
                     else None
                 )
         elif message.type == BrokerMessage.MessageType.MULTI:
+            # XXX Not supported right now
+            # MULTI, COMMIT and ROLLBACK are all not supported in transactional mode right now
+            # This concept is probably not tenable with current architecture because
+            # of how nats works and how we would need to manage rollbacks.
+            # XXX Should this be removed?
```
lferran, line 287: for me, we can remove all of this logic if it's not used.
I am OK if we do it in a separate PR though

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -429,16 +448,18 @@ async def txn(
...
+                if transaction_check:
+                    await sequence_manager.set_last_seqid(txn, partition, seqid)
+                await txn.commit()
 
                 # Slug may have conflicts as its not partitioned properly. We make it as short as possible
-                txn = await self.driver.begin()
-                resource.txn = txn
-                await resource.set_slug()
-                await txn.commit(resource=False)
+                # XXX Does this write need to happen every time?
```
lferran, line 456: no it doesn't. I will put up a PR fixing this after this is merged

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -429,16 +448,18 @@ async def txn(
...
+                async with self.driver.transaction() as txn:
```
lferran, line 457: I think we should name this transaction differently to avoid leaving the main one open?

nucliadb_utils/nucliadb_utils/const.py
```diff
@@ -21,6 +21,37 @@ class PubSubChannels:
...
+
+    class INGEST_PROCESSED:
+        """
+        Resources that have been processed by learning need to be
+        written to the database and then Indexed.
+        """
+
+        name = "nucliadb"
+        subject = "ndb.consumer.processed"
+        group = "nucliadb-processed-messsages"
```
lferran, line 42: too many messsssssagess! 😉

nucliadb_utils/nucliadb_utils/transaction.py
```diff
@@ -46,7 +46,12 @@ def __init__(self, uuid: str, seq: Optional[int] = None):
 
 class LocalTransactionUtility:
     async def commit(
-        self, writer: BrokerMessage, partition: int, wait: bool = False
+        self,
+        writer: BrokerMessage,
+        # same interface
```
lferran, line 51: is this comment needed?

nucliadb_utils/nucliadb_utils/transaction.py
```diff
@@ -117,7 +120,11 @@ def received(waiting_for: WaitFor, event: Event, raw_data: bytes):
             pb = Notification()
             pb.ParseFromString(data)
             if pb.uuid == waiting_for.uuid and pb.action == action_type:
-                if waiting_for.seq is None or pb.seqid == waiting_for.seq:
+                if (
+                    waiting_for.seq is None
+                    or pb.seqid == waiting_for.seq
+                    or action_type != Notification.Action.INDEXED
```
lferran, line 126: I don't understand the reason for this or condition. Can you explain?

nucliadb/nucliadb/ingest/orm/processor/__init__.py
vangheem: The other txn has already been committed at this point but good practice anyways!

nucliadb_utils/nucliadb_utils/transaction.py
vangheem: If we're waiting on the resource to be indexed, the seq id will not match.

PR #874, opened 2023-05-01
use pubsub implementation instead of nats stream

vangheem:
### Description
This implementation allows us to wait for a resource to be indexed in our transaction utility.

### How was this PR tested?


vangheem:
[sc-5274]

nucliadb_node/nucliadb_node/pull.py
```diff
@@ -345,79 +337,28 @@ async def subscribe(self):
...
+        message = Notification(
+            partition=int(indexpb.partition),
+            seqid=indexpb.txid,
+            uuid=indexpb.resource,
+            kbid=indexpb.kbid,
+            action=Notification.INDEXED,
+        )
+
+        await self.pubsub.publish(
+            f"notifyindexed.{indexpb.kbid}", message.SerializeToString()
```
lferran, line 363: should this be a const too?

nucliadb_utils/nucliadb_utils/transaction.py
```diff
@@ -100,51 +97,40 @@ async def closed_cb(self):
...
-        if self.notify_subject is None:
-            logger.warning("No subject defined")
-            return
         await self.pubsub.unsubscribe(
-            key=self.notify_subject.format(kbid=kbid), subscription_id=request_id
+            key=const.PubSubChannels.RESOURCE_NOTIFY.format(kbid=kbid),
+            subscription_id=request_id,
         )
 
+    async def _get_notification_action_type(self):
```
lferran, line 105: should not be async

lferran (commented):
nicely done! really clean 👏🏼

lferran:
With current implementation we will return as soon as one of the nodes has indexed it.

There could still be the case where the other node is behind and the resource is not returned in the search results because we are querying the lagged one.

nucliadb_node/nucliadb_node/pull.py
vangheem: oops, yes

vangheem:
@lferran agreed. I think this is as good as we want to get for the current design.

If we implement the indexnode coordinator, the knowledge of which shards are ready for a search can be encapsulated there.

lferran (approved):

PR #876, opened 2023-05-02
Run cpu bound code at ingest in an executor

jotare:
### Description
- Run `apply_field_metadata` in an executor
- Refactor some code

### How was this PR tested?
Same tests as before


nucliadb/nucliadb/ingest/orm/resource.py
```diff
@@ -606,183 +610,198 @@ async def apply_extracted(self, message: BrokerMessage):
...
+            field_key,
+            metadata,
+            replace_field=replace_field,
+            replace_splits=replace_splits,
+            page_positions=page_positions,
+            extracted_text=await field_obj.get_extracted_text(),
+            basic_user_field_metadata=user_field_metadata,
+        )
+        loop = asyncio.get_running_loop()
+        await loop.run_in_executor(None, apply_field_metadata)
```
vangheem, line 747: What do you think about creating an executor we control for this?

So maybe just create an executor at the module level:

```python
_executor = ThreadPoolExecutor(10)
```

I feel like we want to use more threads and optimize for more even performance with concurrency over individual operation performance.

vangheem (changes requested):
Thank you also for adjusting the readability of some of the code!

nucliadb/nucliadb/ingest/orm/resource.py
jotare: Good idea

jotare: I've added `apply_field_vectors` to also run in an executor

vangheem (approved):

PR #879, opened 2023-05-04
Remove document from find features

lferran:
### Description
Fulltext search is not supported by find endpoint, so it's better if we remove it from the api docs for now

### How was this PR tested?
Integration test

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/find.py
+++ b/nucliadb/nucliadb/search/api/v1/find.py
@@ -17,12 +17,14 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import json
 from datetime import datetime
 from time import time
 from typing import List, Optional, Union
 
 from fastapi import Body, Header, Query, Request, Response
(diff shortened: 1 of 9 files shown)


nucliadb/nucliadb/search/api/v1/find.py
```diff
@@ -121,6 +111,9 @@ async def find_knowledgebox(
     x_nucliadb_user: str = Header(""),
     x_forwarded_for: str = Header(""),
 ) -> Union[KnowledgeboxFindResults, HTTPClientError]:
+    if SearchOptions.DOCUMENT in features:
+        return HTTPClientError(status_code=422, detail="fulltext search not supported")
```
jotare, line 115: NIT IDEA: to avoid duplication, you can try-except `FindRequest` creation and on validation error, raise a 422

jotare (approved):

nucliadb/nucliadb/search/api/v1/find.py
lferran: yes! good idea, I like it. Then we have just one validation point!

PR #881, opened 2023-05-04
Fix handling duplicates on orderer

lferran:
### Description
This made that results matching both for bm25 and vector were showing up the very end of the list.

### How was this PR tested?
Unit testing


lferran:
[ch5350]

jotare (approved):
Good catch!

PR #886, opened 2023-05-08
Remove seek to paragraph implementation

lferran:
### Description
We found out that seeking to the protobuf is not possible since there are some characters that when encoded to binary take up 2 bytes.

### How was this PR tested?
No tests needed


vangheem (approved):

lferran:
[ch5379]

PR #888, opened 2023-05-08
Counters/auditing logic refactoring and decoupling

vangheem:
### Description

- pull out field auditing from ingest processor
- pull out index auditing from ingest processor
- pull out shard allocation from ingest processor

### How was this PR tested?


nucliadb/nucliadb/ingest/orm/node.py
```diff
@@ -101,9 +101,11 @@ async def create_shard_by_kbid(
 
         try:
             # When choosing the nodes, we need to exclude nodes in which there is already a shard from the same KB
-            kb_nodes = [
-                replica.node for shard in kb_shards.shards for replica in shard.replicas
-            ]
+            # XXX removed, why do we care about this? How was this working before?
+            # kb_nodes = [
+            #     replica.node for shard in kb_shards.shards for replica in shard.replicas
+            # ]
```
lferran, line 107: The idea was to spread the replicas of logic shards of a KB as much as possible.

If we are OK of having replicas of different logic shards assigned to the same node, then we can remove it.

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -605,36 +407,50 @@ async def apply_resource(
...
+            self.notify(
+                const.PubSubChannels.RESOURCE_NOTIFY.format(kbid=message.kbid),
+                notification.SerializeToString(),
+            )
         )
 
-    async def notify_abort(
-        self, partition: str, seqid: int, multi: str, kbid: str, uuid: str
+    def notify_abort(
+        self, *, partition: str, seqid: int, multi: str, kbid: str, rid: str
```
lferran, line 439: why is this `*`, useful for?

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -605,36 +407,50 @@ async def apply_resource(
...
             multi=multi,
-            uuid=uuid,
+            uuid=rid,
             kbid=kbid,
             action=Notification.ABORT,
         )
-        await self.notify(
-            const.PubSubChannels.RESOURCE_NOTIFY.format(kbid=kbid),
-            message.SerializeToString(),
+        asyncio.create_task(
```
lferran, line 449: Since we are not awaiting for this anymore, are we OK with not catching possible errors on `self.notify`?

nucliadb/nucliadb/writer/api/v1/resource.py
```diff
@@ -338,11 +337,10 @@ async def reprocess_resource(
     set_info_on_span({"nuclia.rid": rid, "nuclia.kbid": kbid})
 
     storage = await get_storage(service_name=SERVICE_NAME)
-    cache = await get_cache()
     driver = await get_driver()
 
     txn = await driver.begin()
-    kb = KnowledgeBox(txn, storage, cache, kbid)
+    kb = KnowledgeBox(txn, storage, kbid)
```
lferran, line 343: thanks for the unused cache param cleanup!

nucliadb/nucliadb/ingest/consumer/auditing.py
```diff
@@ -0,0 +1,295 @@
...
+            key=const.PubSubChannels.RESOURCE_NOTIFY.format(kbid="*"),
+            group="audit-shard-stats",
+            subscription_id=self.subscription_id,
+        )
+
+    async def finalize(self) -> None:
+        await self.pubsub.unsubscribe(self.subscription_id)
+        await self.task_handler.finalize()
+
+    async def handle_message(self, raw_data: bytes) -> None:
```
lferran, line 92: no need to be async

nucliadb/nucliadb/ingest/consumer/auditing.py
```diff
@@ -0,0 +1,295 @@
...
+        except ShardsNotFound:
+            logger.warning(f"No shards found for kbid {kbid}, skipping")
+            return
+
+        total_fields = 0
+        total_paragraphs = 0
+
+        for shard_obj in shard_groups:
+            node, shard_id, _ = self.node_manager.choose_node(shard_obj)
+            shard_counter: Counter = await node.sidecar.GetCount(ShardId(id=shard_id))  # type: ignore
```
lferran, line 119: The `GetCount` logic at the sidecar is caching the counters. May be a good time to review it and see if it makes sense?

https://github.com/nuclia/nucliadb/blob/main/nucliadb_node/nucliadb_node/reader.py#L44

nucliadb/nucliadb/ingest/consumer/shard_creator.py
```diff
@@ -0,0 +1,108 @@
...
+        self.node_manager = NodesManager(driver)
+        self.task_handler = DelayedTaskHandler(check_delay)
+
+    async def initialize(self) -> None:
+        self.subscription_id = str(uuid.uuid4())
+        await self.task_handler.initialize()
+        await self.pubsub.subscribe(
+            handler=self.handle_message,
+            key=const.PubSubChannels.RESOURCE_NOTIFY.format(kbid="*"),
+            group="audit",
```
lferran, line 72: group="shard-creator"?

lferran (commented):
great decoupling effort 👏🏼 just come comments

nucliadb/nucliadb/ingest/orm/node.py
vangheem: Okay, I think I'll tweak this logic. I think right now we might be in a situation where we wouldn't find enough nodes when we did have them available

nucliadb/nucliadb/ingest/orm/processor/__init__.py
vangheem: it forces the calling to provide all keyword args. I can remove it if you like but I prefer to have all named arguments for readability and preventing position argument errors.

nucliadb/nucliadb/ingest/orm/processor/__init__.py
vangheem: I think I'll just put this back to synchronous. I guess this is premature optimization and having it hard fail would probably be better right now

nucliadb/nucliadb/ingest/orm/processor/__init__.py
lferran: nice! no, I like it. I was wondering for a long time... where can I read about this in the python docs?

nucliadb/nucliadb/ingest/consumer/auditing.py
vangheem: ahh nice and then I could just call GetShard on the reader directly and we can remove all that logic

lferran: I don't know the reasoning behind caching it... maybe @bloodbare or @sunbit know

vangheem: I'm going to leave this alone for now. We'll have to re-evaluate if we want caching at the side car.

nucliadb/nucliadb/ingest/tests/unit/consumer/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+    counter = 0
+
+    async def handler():
+        nonlocal counter
+        counter += 1
+
+    dth.schedule("key1", handler)
+    dth.schedule("key1", handler)
+    dth.schedule("key1", handler)
```
lferran, line 42: Can you write a test that makes it explicit which of these schedule operations will be ignored?

We want the latter to be effective, don't we?

vangheem: I think I have a better test to show how the utility is meant to work!

lferran (approved):

PR #890, opened 2023-05-09
Do not commit slug every time

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/processor/__init__.py
+++ b/nucliadb/nucliadb/ingest/orm/processor/__init__.py
@@ -382,7 +382,6 @@ class Processor:
         partition: str,
         transaction_check: bool = True,
     ) -> Optional[TxnResult]:
-        breakpoint()
         if len(messages) == 0:
             return None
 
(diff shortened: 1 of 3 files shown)


lferran:
[ch5406]

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -452,12 +462,8 @@ async def txn(
...
                 await txn.commit()
 
-                # Slug may have conflicts as its not partitioned properly. We make it as short as possible
-                # XXX Does this write need to happen every time?
-                async with self.driver.transaction() as slug_txn:
-                    resource.txn = slug_txn
-                    await resource.set_slug()
-                    await slug_txn.commit()
+                if created or resource.slug_modified:
+                    await self.commit_slug(resource)
```
jotare, line 466: Is this function useful? Why not adding a `resource.set_slug()` inside this if before `txn.commit`? Makes sense to have two transactions?

lferran: Yes, we want to commit the slug in a separate (and shorter) transaction on purpose. That was the previous logic and I prefer to keep it as is

IMO the function is useful for readability: we are doing more than one thing to commit the slug.

jotare (approved):

PR #891, opened 2023-05-09
Add search cache config

lferran:
### Description

### How was this PR tested?


vangheem (changes requested):
Can we just have the integration be done through the `env` key instead?

I'd like to stop doing this config mapping everywhere. (which reminds me that we should probably try to clean all this up as well)

jotare (approved):

vangheem (approved):
beautiful!

PR #896, opened 2023-05-10
Standalone deploy improvements

vangheem:
### Description
- add standalone chart to the repo
- default to europe1 zone ( we do not have a "dev" zone)
- back off for 429 and other errors instead of retrying endlessly

### How was this PR tested?

### DIFF
--- a/charts/nucliadb/Chart.yaml
+++ b/charts/nucliadb/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - nucliadb
   - nuclia
 home: https://nuclia.com
-icon: https://nuclia.com/logo.png
+https://nucliadb.com/wp-content/uploads/2023/02/logo-2.svg
 sources:
   - https://github.com/nuclia/nucliadb
 maintainers:
--- a/charts/nucliadb/Chart.yaml
+++ b/charts/nucliadb/Chart.yaml
@@ -7,7 +7,7 @@ keywords:
   - nucliadb
   - nuclia
 home: https://nuclia.com
-https://nucliadb.com/wp-content/uploads/2023/02/logo-2.svg
+icon: https://nucliadb.com/wp-content/uploads/2023/02/logo-2.svg
 sources:
   - https://github.com/nuclia/nucliadb
 maintainers:
--- a/charts/nucliadb/Chart.yaml
+++ b/charts/nucliadb/Chart.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+description: NucliaDB Chart for standalone deployment
+name: nucliadb
+version: 99999.99999.99999
+appVersion: 99999.99999.99999
+keywords:
+  - nucliadb
+  - nuclia
+home: https://nuclia.com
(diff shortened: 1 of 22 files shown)


charts/nucliadb/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+description: NucliaDB Chart for standalone deployment
+name: nucliadb
+version: 99999.99999.99999
+appVersion: 99999.99999.99999
+keywords:
+  - nucliadb
+  - nuclia
+home: https://nuclia.com
+icon: https://nuclia.com/logo.png
```
lferran, line 10: this page does not exist

lferran (approved):

charts/nucliadb/Chart.yaml
vangheem: it's a copy from other chats haha :)

vangheem:
[sc-5410]

charts/nucliadb/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+#  limits:
+#    memory: "2600Mi"
+#  requests:
+#    memory: "600Mi"
+#    cpu: 1
+
+storage:
+  class: ssd
+  size: 100Gi
+  data_path: /data
```
lferran, line 29: is it the same setting as `env: DATA_PATH?`

nucliadb/nucliadb/ingest/consumer/pull.py
```diff
@@ -131,64 +121,36 @@ async def loop(self):
...
-                            await asyncio.sleep(self.pull_time_empty_backoff)
-                        else:
-                            logger.info(f"Proxy pull answered with error: {data}")
-                            await asyncio.sleep(self.pull_time_error_backoff)
+                    data = await processing_http_client.pull(self.partition)
+                    if data.status == "ok":
+                        logger.info(
+                            f"Message received from proxy, partition: {self.partition}"
+                        )
+                        async with self.lock:
```
lferran, line 137: I know it was already there but: why is the lock needed here?

nucliadb/nucliadb/ingest/consumer/pull.py
```diff
@@ -202,7 +164,7 @@ async def _loop(self):
 
                 except ClientConnectorError:
                     logger.error(
-                        f"Could not connect to {url}, verify your internet connection"
+                        f"Could not connect to processing engine, verify your internet connection"
```
lferran, line 167: It would be nice to keep the url in the log, to know if it's the internal or the public service what we can't connect to

lferran (approved):
LGTM!
Just a few NIT comments

nucliadb/nucliadb/ingest/consumer/pull.py
vangheem: yeah, it's weird. I'm not sure either. Probably some legacy reason that no longer applies. I'm going to remove it.

PR #899, opened 2023-05-11
Improve ingest metrics and health checks

vangheem:
### Description
- allows us to define custom health checks
- utilize grpc health checks in components other than ingest
- better health checks for subscribers
- remove ingest grpc from other components

### How was this PR tested?


nucliadb/nucliadb/ingest/consumer/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+from nucliadb_telemetry import metrics
+
+total_messages = metrics.Counter(
+    "nuclia_subscriber_messages", labels={"type": "", "action": "scheduled"}
```
lferran, line 23: `nucliadb_` prefix instead? (and the same below)

nucliadb/nucliadb/ingest/consumer/metrics.py
```diff
@@ -0,0 +1,25 @@
...
+)
+handler_histo = metrics.Observer("nuclia_shard_creator_handler", labels={"type": ""})
```
lferran, line 25: Should the metric name be `nucliadb_handler` and then use `type` label to distinguish between handlers?

vangheem: blah, definitely. I forgot to fix these. Thanks.

PR #904, opened 2023-05-15
Extra metadata

lferran:
### Description
This PR adds a new metadata at the resource level named `extra`.
The purpose is to allow users to store arbitrary json metadata along their resources that is rendered on search results.
This metadata is not supposed to be searchable.

Implementation details:
- I'm storing this data in a new maindb key.
- The metadata will be optionally serialized when `show=extra` option is passed on resource GET or in search endpoints.

### How was this PR tested?
Integration tests


nucliadb_models/nucliadb_models/writer.py
```diff
@@ -78,6 +81,15 @@ def icon_check(cls, v):
 
         return v
 
+    @validator("extra")
+    def extra_check(cls, value):
+        limit = 400_000
```
jotare, line 86: Any reason for this number?

jotare (approved):

PR #909, opened 2023-05-16
Sharding on paragraph count

lferran:
### Description
This PR aims to change the way we create new shards: from number of fields to number of paragraphs.

### How was this PR tested?
Modified existing tests


nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -759,21 +763,15 @@ async def ReIndex(self, request: IndexResource, context=None) -> IndexStatus:  #
...
-
-                if counter is not None and counter.fields > settings.max_shard_fields:
-                    # check to see if we've exceeded the max resources per shard
-                    # and create a new shard for new resources to land on
-                    similarity = await kbobj.get_similarity()
-                    shard = await node_klass.create_shard_by_kbid(
-                        txn, request.kbid, similarity=similarity
-                    )
+                # trigger possibly creating new shard manually here
+                await self.sc.process_kb(request.kbid)
```
vangheem, line 774: I'm not sure we should import and use this here.

Indexing will already be listening to indexed events and be doing this new shard logic.

In fact, it could cause problems if we do it in 2 places right?

vangheem (changes requested):

nucliadb/nucliadb/ingest/service/writer.py
lferran: 🤦🏼 you're right! I forgot that the notifications are sent via nats pubsub. removed!

vangheem (approved):

PR #915, opened 2023-05-19
better grpc retries

vangheem:
### Description
It turns out the grpc client already has a less advertised feature of implementing retry logic for errors: https://github.com/grpc/proposal/blob/master/A6-client-retries.md

other things:
- increase dns cache for gcs connector
- more exceptions to retry on for gcs errors

[sc-5463]

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/processor/__init__.py
+++ b/nucliadb/nucliadb/ingest/orm/processor/__init__.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import asyncio
 import logging
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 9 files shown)


nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -271,6 +272,13 @@ async def txn(
                     partition=partition, seqid=seqid, multi=multi, kbid=kbid, rid=uuid
                 )
                 logger.warning(f"This message did not modify the resource")
+        except (asyncio.TimeoutError, asyncio.CancelledError):  # pragma: no cover
```
lferran, line 275: Should we catch ConnectionError too? (both server and client errors)

vangheem: Yes, agreed. Let me see if there is a better way to do this.

nucliadb_utils/nucliadb_utils/storages/gcs.py
```diff
@@ -469,7 +474,9 @@ async def initialize(self, service_name: Optional[str] = None):
         loop = asyncio.get_event_loop()
 
         await setup_telemetry(service_name or "GCS_SERVICE")
-        self.session = aiohttp.ClientSession(loop=loop)
+        self.session = aiohttp.ClientSession(
+            loop=loop, connector=aiohttp.TCPConnector(ttl_dns_cache=60 * 5)
```
lferran, line 478: great! what was the default value?

nucliadb/nucliadb/ingest/orm/processor/__init__.py
lferran: Maybe all those exceptions that we define as retriable should be catched so we hard-fail (and retry even more on the next processing attempt)

nucliadb_utils/nucliadb_utils/storages/gcs.py
vangheem: 10 seconds

PR #917, opened 2023-05-19
Implement telemetry context variables

vangheem:
### Description
This PR implements a contextvar implementation for telemetry.

This enables us to annotate "contextual" info across logs and http requests.

Scope of this PR:
- add `context` module to allow pushing new metadata
- when adding "context", also add to the current active span
- add ContextInjectorMiddleware which automatically injects path params into the context(should allow us to remove a lot of manual context marking for requests)
- automatically add context information to structured logs

### How was this PR tested?

### DIFF
--- a/nucliadb_telemetry/nucliadb_telemetry/context.py
+++ b/nucliadb_telemetry/nucliadb_telemetry/context.py
@@ -26,8 +26,9 @@
 # This allows us to leverage context data for both tracing and logs.
 #
 import contextvars
+from typing import Optional
 
-context_data = contextvars.ContextVar("data", default=None)
+context_data = contextvars.ContextVar[Optional[dict[str, str]]]("data", default=None)
 
 
 def add_context(new_data: dict[str, str]):
--- a/nucliadb_telemetry/nucliadb_telemetry/context.py
+++ b/nucliadb_telemetry/nucliadb_telemetry/context.py
@@ -1,0 +1,49 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+# ABOUT
+# The module is meant to manage telemetry oriented context data.
+# This is a little different than the span context because we do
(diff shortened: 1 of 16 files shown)


nucliadb_telemetry/nucliadb_telemetry/context.py
```diff
@@ -0,0 +1,80 @@
...
+
+def add_context(new_data: dict[str, str]):
+    """
+    This implementation always merges and sets the context, even if is was already set.
+
+    This is so data is propated forward but not backward.
+    """
+
+    # set the data on the current active span
+    set_info_on_span({f"nuclia.{key}": value for key, value in new_data.items()})
```
lferran, line 46: nitpick: `nucliadb.{key}` instead?

nucliadb_telemetry/nucliadb_telemetry/tests/integration/test_fastapi.py
```diff
@@ -210,7 +180,7 @@ class TestCasePrometheusMiddlewareFilterUnhandledPaths:
     @pytest.fixture(scope="class")
     def app(self):
         app_ = Starlette()
-        app_.add_middleware(PrometheusMiddleware, filter_unhandled_paths=True)
```
lferran, line 213: IIRC I added `filter_unhandled_paths` so that urls that do not match any of our endpoints do not get into prometheus metrics.

Without it, if someone would do a GET `api/kb/<kbid>/non-existing-endpoint`, the <kbid> was appearing in the metric labels

nucliadb_telemetry/nucliadb_telemetry/context.py
```diff
@@ -0,0 +1,80 @@
...
+# This allows us to leverage context data for both tracing and logs.
+#
+import contextvars
+from typing import Optional, Sequence, Union
+
+from opentelemetry.trace import get_current_span
+
+from nucliadb_telemetry.settings import telemetry_settings
+
+context_data = contextvars.ContextVar[Optional[dict[str, str]]]("data", default=None)
```
lferran, line 35: Shouldn't the typing be specified like this instead?
```
context_data: contextvars.ContextVar[Optional[dict[str, str]]] = ContextVar('data', default=None)
```

nucliadb_telemetry/nucliadb_telemetry/context.py
vangheem: everywhere else it's "nuclia." and since this library is potentially consumed by other teams, I thought it should stay this way.

nucliadb_telemetry/nucliadb_telemetry/tests/integration/test_fastapi.py
vangheem: Yup, totally makes sense. I just make this the default behavior now and not optional.

nucliadb_telemetry/nucliadb_telemetry/fastapi/__init__.py
```diff
@@ -69,16 +71,20 @@ def instrument_app(
 ):
     if metrics:
         # b/w compat
-        app.add_middleware(PrometheusMiddleware, filter_unhandled_paths=True)
-    if SentryAsgiMiddleware is not None:
-        app.add_middleware(SentryAsgiMiddleware)
+        app.add_middleware(PrometheusMiddleware)
```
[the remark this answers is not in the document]
vangheem: This option is default and only way that it'll run now.

lferran (approved):

PR #933, opened 2023-05-25
Improve search openapi docs

lferran:
### Description
Initial effort to have a more complete openapi specs where all api endpoint parameters are documented.
It also aims to have a unified way to document fields that appear both as query params and body params.

### How was this PR tested?
Existing tests should cover any possible error


nucliadb_models/nucliadb_models/search.py
```diff
@@ -341,16 +341,43 @@ def from_message(cls: Type[_T], message: PBShards) -> _T:
...
 
+class ParamDefaults:
+    query = dict(default="", title="Query", description="The query to search for")
+    advanced_query = dict(
+        default=None,
+        title="Advanced query",
+        description="An advanced query to search for. Follows the tantivy query parser syntax: https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html",  # noqa: E501
+    )
+    fields = dict(
+        default=[], title="Fields", description="The list of fields to search in"
```
vangheem, line 352: Can we mention how to use this field?

`f` -> only file fields
`f/file` -> only file field named `file`

but `t`, etc are supported as well right?

nucliadb_models/nucliadb_models/search.py
```diff
@@ -341,16 +341,43 @@ def from_message(cls: Type[_T], message: PBShards) -> _T:
...
+    )
+    filters = dict(
+        default=[],
+        title="Filters",
+        description="The list of filters to apply. The filters follow the syntax defined in https://github.com/nuclia/nucliadb/blob/main/docs/internal/SEARCH.md#filters-and-facets",  # noqa: E501
```
vangheem, line 357: Let's update these doc links to reference docs.nuclia.dev instead

lferran: done!

nucliadb_models/nucliadb_models/search.py
lferran: done!

vangheem (approved):

PR #938, opened 2023-05-29
Add JSON format to text field

jotare:
### Description

### How was this PR tested?


nucliadb_models/nucliadb_models/text.py
```diff
@@ -87,6 +89,7 @@ class PushTextFormat(TextFormatValue, Enum):  # type: ignore
     HTML = 1
     MARKDOWN = 2
     RST = 3
+    JSON = 4
```
jotare, line 92: @bloodbare make sure this is expected in the processor

bloodbare: +1

bloodbare (commented):
Should'nt we validate on writing that the field is real json ?

bloodbare (approved):

PR #949, opened 2023-06-02
Remove unneeded diff from tokio task metrics collect

jotare:
### Description
As we are collecting on `PrometheusMeter` the metrics from an interval, we don't need to remove the previous value

### How was this PR tested?


hermeGarcia (approved):
Good catch

PR #963, opened 2023-06-07
Add tracing middleware to core and clean gRPC instrumentation

jotare:
### Description
Instrument gRPC servers using tower middleware and remove `#[tracing::instrument]` like decorations.

**TRADE-OFF**: we lose the ability to know which exact line of code is handling the gRPC call (as it's not longer using a proc_macro, we cannot know it at compile time). However, knowing who is handling the request is trivial (as there's only a server) so I don't think it's a problem.

### How was this PR tested?
Locally using a local Jaeger and enabling telemetry on tests.


hermeGarcia (approved):
Great work!

PR #971, opened 2023-06-08
add migration strategy proposal

vangheem:
The purpose of this PR is to add a proposal for implementation a data migration strategy.

### DIFF
--- a/proposals/records/3-migration-strategy.md
+++ b/proposals/records/3-migration-strategy.md
@@ -1,0 +1,99 @@
+# Migration Strategy
+
+This is a proposal for how we manage registering and running migrations on the database.
+
+The proposal focuses on providing an automated way to migrate data and index nodes.
+
+
+## Current solution/situation
+
+We currently do not have any cohesive strategy around migrations on NucliaDB.
+It is currently managed through manually running scripts against the cluster.
+
+Additionally, the ingest grpc service has a method `CleanAndUpgradeKnowledgeBoxIndex`
+which only deletes a shard and recreates it. Reindexing data there needs to be
+done manually.
+
+
+### Types of data migrations
+
+- Blob storage
+- K/V Storage
+- IndexNodes
+
+## Proposed Solution
+
+1. Build a framework for registering and running migrations on the platform.
+2. Require all changes to to be b/w compatible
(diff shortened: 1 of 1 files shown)


proposals/records/3-migration-strategy.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. Build a framework for registering and running migrations on the platform.
+2. Require all changes to to be b/w compatible
+3. Migration are not reversible(register a new migration if you need to "undo")
+
+
+### Types of migrations supported
+
+- IndexNode shard rollovers: create new shards, index all data to new shards, swap active shards over
+- manual: manually run a snippet of code against every knowledge box
```
lferran, line 34: do you mean "custom" instead (custom snippets of code that tweak/fix data)? 
"manual" to me implies "not automated".

proposals/records/3-migration-strategy.md
```diff
@@ -0,0 +1,106 @@
...
+
+- Provide metrics for every migration
+- Alerts for failed migrations
+- Errors reported to sentry
+- Hard failure, migration should retry with backoff and show hard failures in k8s
+
+
+#### Implementation
+
+Running migrations will be done through a command line utility `nucliadb-run-migrations`.
```
lferran, line 90: Would be nice if this command allows:
- running migration in dry run mode
- running migration for a particular kb

proposals/records/3-migration-strategy.md
```diff
@@ -0,0 +1,106 @@
...
+
+This utility will be run automatically with a post release helm/argo hook.
+
```
lferran, line 93: I think we will need a way (endpoint, or in the migrations command) to introspect migration metadata like:
 - at which version a particular KB is
 - whether it is running the migration at that moment in time

lferran, line 93: One thing that comes to mind is that newly created KBs should be updated to point to the latest migration sequence so that only new migrations added from that point in time are applied to them.

vangheem: I was thinking this should be done with metrics.

We still need an ndb metrics exporter component btw.

lferran (approved):

PR #968, opened 2023-06-07
Refactor cluster manager/nodes/shards

vangheem:
### How was this PR tested?
Describe how you tested this PR.

PR #973, opened 2023-06-09
Sort generative ai input text by score

lferran:
### Description
Since the context may be chopped when it's too long, we need to make sure that the context is properly sorted and that the most relevan results are on the top.

### How was this PR tested?
Unit tests


lferran:
[sc5873]

lferran:
closing, as it will be done in https://github.com/nuclia/nucliadb/pull/968

PR #974, opened 2023-06-09
HNSW: take into account the user defined min score

lferran:
### Description
This PR passes through the min score defined by the user down to the index node so that it can efficiently stop looking for more results if it doesn't find results with higher score than the `min_score` defined by the user.

### How was this PR tested?
Tests on the node

### Shortcut story
[sc-5583]


hermeGarcia (changes requested):
Looks good to me, just one thing needs to be changed.

hermeGarcia (approved):
Good job!

lferran:
superseded by https://github.com/nuclia/nucliadb/pull/1023

PR #990, opened 2023-06-14
Return auto filters on search response

lferran:
### Description
This PR aims to return the automatic filters added to the query so that the user is aware of what is going on with the search

### How was this PR tested?
Integration test


lferran:
[sc5929]

PR #996, opened 2023-06-14
Handle find incomplete results on chat endpoint

lferran:
### Description
Reporting that we could not get an answer is better than providing a poor answer!

### How was this PR tested?
Integration test


lferran:
[sh-5521]

nucliadb/nucliadb/search/api/v1/find.py
```diff
@@ -191,6 +195,7 @@ async def find(
     x_nucliadb_user: str,
     x_forwarded_for: str,
     do_audit: bool = True,
+    raise_on_incomplete_results: bool = False,
```
vangheem, line 198: this feels awkward -- to cause an exception to be raised from a parameter.

vangheem (approved):

nucliadb/nucliadb/search/api/v1/find.py
lferran: you are right. let me try to refactor

PR #1008, opened 2023-06-16
fix TUS PG Support

vangheem:
### Description

### How was this PR tested?


nucliadb/nucliadb/writer/tus/pg.py
```diff
@@ -0,0 +1,129 @@
...
+            async with conn.transaction():
+                dl = PostgresFileDataLayer(conn)
+                await dl.delete_file(kbid, uri)
+
+
+class PGBlobStore(BlobStore):
+    def __init__(self, dsn: str):
+        self.dsn = dsn
+        self.source = CloudFile.POSTGRES
+        self._lock = asyncio.Lock()
```
lferran, line 114: why is this lock needed?

nucliadb_utils/nucliadb_utils/storages/pg.py
```diff
@@ -188,7 +204,16 @@ async def move(
                 origin_kb,
                 origin_key,
             )
-
+            # make sure to delete the destination first in
```
lferran, line 207: Just out of curiosity: I don't understand why we need a delete here. Can't we just overwrite the existing values?

lferran (approved):

nucliadb_utils/nucliadb_utils/storages/pg.py
vangheem: This is essentially doing a replace.

The new upload is in a different file id and this clears out the old

vangheem: This is essentially doing a replace.

The new upload is in a different file id and this clears out the old

nucliadb/nucliadb/writer/tus/pg.py
vangheem: It's probably not necessary and the lock should be at another level

PR #1009, opened 2023-06-16
Chat on a resource (with find)

lferran:
### Description
- Adds endpoints to chat on a resource (by slug and by rid).
- Refactors existing chat streaming logic to not duplicate the code
- Adds handlers to sdk

### How was this PR tested?
e2e tests (with sdk) + local tests


nucliadb/nucliadb/search/api/v1/resource/chat.py
```diff
@@ -0,0 +1,189 @@
...
+from nucliadb_utils.authentication import requires
+from nucliadb_utils.exceptions import LimitsExceededError
+
+
+class ResourceNotFoundError(Exception):
+    pass
+
+
+class IncompleteFindResourceResults(Exception):
+    pass
```
jotare, line 48: Should we have them in an errors/exceptions module?

nucliadb/nucliadb/search/search/chat/query.py
```diff
@@ -0,0 +1,136 @@
...
+        user_id=x_nucliadb_user,
+    )
+    return await predict.rephrase_query(kbid, req)
+
+
+async def chat(
+    kbid: str,
+    find_results: KnowledgeboxFindResults,
+    chat_request: ChatRequest,
+    x_nucliadb_user: str,
```
jotare, line 63: Same

nucliadb_protos/nodereader.proto
```diff
@@ -108,6 +108,7 @@ message ParagraphSearchRequest {
     bool with_duplicates = 14;
     bool only_faceted = 15;
     optional string advanced_query = 16;
+    repeated string keys = 17;
```
jotare, line 111: What are this `keys`? Could we have a more descriptive name?

Dockerfile.withbinding
```diff
@@ -51,7 +51,7 @@ RUN ARCH="$(uname -m)"; \
 
 RUN mkdir -p /usr/src/app
 COPY --from=python_rust /nucliadb/target/wheels/ /wheels/
-RUN pip install /wheels/*.whl
+RUN pip install /wheels/*.whl --force-reinstall
```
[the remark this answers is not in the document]
lferran: yup, we don't need it. removing

nucliadb_protos/nodereader.proto
lferran: AFAIK "keys is a core concept in the node" which is essentially a paragraph id.
I added the `_filter` suffix, so that at least it is clear that we use this attribute for filtering.

jotare: I didn't know. If it is, that's okay for me

jotare (commented):
Not sure about Rust filtering, Python side approved :)

hermeGarcia (approved):
Rust looks good!

PR #1024, opened 2023-06-20
Improve logging on node errors

lferran:
### Description
Some handled internal errors from the node were not being logged

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb/nucliadb/search/requesters/utils.py
+++ b/nucliadb/nucliadb/search/requesters/utils.py
@@ -155,7 +155,7 @@ async def node_query(
 
     if not ops:
         await abort_transaction()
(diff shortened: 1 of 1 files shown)


nucliadb/nucliadb/search/requesters/utils.py
```diff
@@ -198,6 +198,9 @@ def validate_node_query_results(results: list[Any]) -> Optional[HTTPException]:
                     if "AllButQueryForbidden" in result.details():
                         status_code = 412
                         reason = result.details().split(":")[-1].strip().strip("'")
+                    else:
```
jotare, line 201: Nice one!

PR #1026, opened 2023-06-21
Return trace ids on response headers

lferran:
### Description
This PR includes:
- Adding a new middleware `CaptureTraceIdMiddleware` in nucliadb_utils that will inject the trace id of the request on a response header.
- This middleware is added to the app only if specified.
- Configures `reader`, `train`, `writer` and `search` components to enable this middleware.

### How was this PR tested?
Integration and unit tests


nucliadb_telemetry/nucliadb_telemetry/fastapi/__init__.py
```diff
@@ -88,3 +96,6 @@ def instrument_app(
         # add last to catch all exceptions
         # `add_middleware` always adds to the beginning of the middleware list
         app.add_middleware(SentryAsgiMiddleware)
+
+    if trace_id_header is not None:
+        app.add_middleware(CaptureTraceIdMiddleware, response_header=trace_id_header)
```
jotare, line 101: We have to be careful with middleware order. `OpenTelemetryMiddleware` starts the span you are getting in `CaptureTraceIdMiddleware`. Could you add a comment to note this?

jotare (approved):

nucliadb_telemetry/nucliadb_telemetry/fastapi/__init__.py
```diff
@@ -40,6 +41,9 @@
     SentryAsgiMiddleware = None  # type: ignore
 
 
+NUCLIADB_TRACE_ID_HEADER = "X-NUCLIADB-TRACE-ID"
```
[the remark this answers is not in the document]
lferran: @sunbit that is why I left the name of the header as a parameter of the `instrument_app` function.

Would you rather simply hardcode it to `X-NUCLIA-TRACE-ID`?

nucliadb_telemetry/nucliadb_telemetry/fastapi/__init__.py
lferran: agree, done!

PR #1027, opened 2023-06-21
optimise suggest

hermeGarcia:
### Description
`split_suggest_query` was spending too much time allocating new strings due to prepending using `format!("{word} {prefix}")`. The new algorithm allocates an additional vector of size `max_group`, but makes the main loop faster.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_node/src/services/reader.rs
+++ b/nucliadb_node/src/services/reader.rs
@@ -43,6 +43,7 @@ use crate::telemetry::run_with_telemetry;
 const RELOAD_PERIOD: u128 = 5000;
 const FIXED_VECTORS_RESULTS: usize = 10;
 const MAX_SUGGEST_COMPOUND_WORDS: usize = 3;
+const MIN_VIABLE_PREFIX_SUGGEST: usize = 3;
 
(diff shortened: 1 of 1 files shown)


nucliadb_node/src/services/reader.rs
```diff
@@ -298,26 +299,20 @@ impl ShardReaderService {
...
+        // prepend to the partial strings.
+        let relevant_words: Vec<_> = query.split(' ').rev().take(max_group).collect();
+        let mut prefixes = vec![String::new(); max_group];
+        for (index, word) in relevant_words.into_iter().rev().enumerate() {
+            // The inner loop is upper-bounded by max_group
+            for prefix in prefixes.iter_mut().take(index + 1) {
+                if !prefix.is_empty() {
+                    prefix.push(' ');
+                }
+                prefix.push_str(word);
```
jotare, line 312: I don't get why this two loops are more efficient than one.

With something like:
```rust
for word in words.rev().take(max_group)
```
you can remove the need for the counter and iterate only one time per word.

jotare (approved):
Nice work!

PR #1032, opened 2023-06-22
Text index indexing optimizations

jotare:
### Description
Small optimizations to avoid unneeded `clone`s and some renaming

### How was this PR tested?

### DIFF
--- a/nucliadb_texts/src/writer.rs
+++ b/nucliadb_texts/src/writer.rs
@@ -249,7 +249,7 @@ impl TextWriterService {
                 let facet = Facet::from(label.as_str());
                 subdoc.add_facet(self.schema.facets, facet);
             }
-            self.writer.add_document(subdoc.clone()).unwrap();
+            self.writer.add_document(subdoc).unwrap();
         }
     }
 }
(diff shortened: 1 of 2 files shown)


nucliadb_texts/src/writer.rs
```diff
@@ -225,35 +231,31 @@ impl TextWriterService {
...
-        for label in resource_labels {
+        for label in resource.labels.iter() {
             let facet = Facet::from(label.as_str());
-            doc.add_facet(self.schema.facets, facet);
+            base_doc.add_facet(self.schema.facets, facet);
         }
 
         for (field, text_info) in &resource.texts {
-            let mut subdoc = doc.clone();
+            let mut doc = base_doc.clone();
```
hermeGarcia, line 247: Maybe call it `new_doc` instead of just `doc` to not create confusion with the macro.

hermeGarcia (changes requested):
LGTM, just one detail

nucliadb_texts/src/writer.rs
jotare: What about `field_doc`, as it's a doc for a field?

hermeGarcia: sounds good

hermeGarcia (approved):

PR #1041, opened 2023-06-26
fix trace id format

vangheem:
### Description
Converting to string does not actually work and integer value trace ids are converted to hex.


vangheem:
[sc-6086]

lferran (approved):

PR #1046, opened 2023-06-27
Use model min score in searches

lferran:
### Description
New KBs created in the cloud offering are created specifying the default min score for the semantic model associated to that KB.
This PR aims to utilize this min score as default for all search operations. 

### How was this PR tested?
Integration and unit tests


nucliadb/nucliadb/search/search/query.py
```diff
@@ -326,3 +334,26 @@ def pre_process_query(user_query: str) -> str:
             result.append(term)
 
     return " ".join(result)
+
+
+# TODO: should this go into a common module? or into a ingest exposed utils module?
```
vangheem, line 339: this orm code needs a lot of refactoring.

PR #1048, opened 2023-06-27
Ask a document

lferran:
### Description
This PR implements the functionality to ask a question to the whole document. That means that we will be sending all extracted text to the predict api.

### How was this PR tested?
- Integration tests
- unit tests


nucliadb/nucliadb/search/api/v1/resource/ask.py
```diff
@@ -0,0 +1,139 @@
...
+        for field_type, field_id in await orm_resource.get_fields_ids():
+            field_obj = await orm_resource.get_field(field_id, field_type, load=False)
+            etxt = await field_obj.get_extracted_text()
+            if etxt is None:
+                logger.warning(
+                    f"Skipping field {field_id}, as it does not have extracted text yet!"
+                )
+                continue
+
+            fcm = await field_obj.get_field_metadata()
```
vangheem, line 105: Do you really think we need to split it? If we're pulling everything, maybe we can just send the data as is?

nucliadb/nucliadb/search/api/v1/resource/ask.py
```diff
@@ -0,0 +1,139 @@
...
+            if fcm is None:
+                logger.warning(f"Field metadata not found for {field_id}")
+                blocks.append(split_by_newlines(etxt))
```
vangheem, line 108: At the very least, maybe we don't do this manual splitting here and just send add the block without newline split?

vangheem (approved):
okay, I left comments about splitting but I guess it's required for how we process the input on the understanding API.

Looks good!

nucliadb/nucliadb/search/api/v1/resource/ask.py
lferran: yes, learning needs this. since they are already splitting by paragraphs on processing time, it would be double work. Correct me if I am wrong @Ciniesta

lferran:
[sc-6027]

nucliadb/nucliadb/search/api/v1/resource/ask.py
Ciniesta: yes!  we need it in paragraphs if possible. 
The reasoning behind this is that if we get it as a single string and it's bigger than the LLM we will have to split it in parts to make multiple calls. The issue is how to divide it in parts that fit the context but that still make sense on their own. To mitigate this we just group paragraphs together until we fill the context size. And since the text division in paragraphs is already calculated at processing time, it does not make a lot of sense for us to do it again.

PR #1053, opened 2023-06-28
Add annotation related metadata to the API

jotare:
### Description
Related with DONUT training.
This is the first part, with API and storing, dataset/train in another PR

### How was this PR tested?


nucliadb_models/nucliadb_models/metadata.py
```diff
@@ -292,6 +307,15 @@ def from_message(cls: Type[_T], message: resources_pb2.UserFieldMetadata) -> _T:
             including_default_value_fields=True,
             use_integers_for_enums=True,
         )
+        value["selections"] = [
+            MessageToDict(
+                selections,
+                preserving_proto_field_name=True,
+                including_default_value_fields=True,
+                use_integers_for_enums=True,
```
lferran, line 315: don't we want to serialize the enums with the string instead of with the integer?

lferran (approved):
lgtm! 👍🏼 

this doesn't need any extra logic in the ingest so that the data is persisted because it's inside the basic protobuf, right?

nucliadb_models/nucliadb_models/metadata.py
jotare: Not sure about this, I'm using the same `MessageToDict` as above

jotare:
> lgtm! 👍🏼
> 
> this doesn't need any extra logic in the ingest so that the data is persisted because it's inside the basic protobuf, right?

I don't think so. As metadata is already stored as a protobuffer inside the basic

PR #1056, opened 2023-06-28
python refactoring proposal

vangheem:
We have some technical debt around many of our python modules(especially ingress) that we need to find a good path forward one.

This is a proposal to help start defining some patterns to follow when refactoring our code base

### DIFF
--- a/proposals/records/4-python-refactoring.md
+++ b/proposals/records/4-python-refactoring.md
@@ -152,12 +152,14 @@ The proposal focusses on moving toward 3 patterns:
 
 - `ingest` focuses on components for ingesting data and not providing interfaces to other components
 - shared tooling and data layers are moved to the `nucliadb.common` module
+- no main component should import from the `injest` module
+- nucliadb should stop using the `ingest` grpc service
 
 ### Isolate our data layers into separate interfaces
 
 - create data layer management components
 - limit direct exposure of k/v and blob storage drivers
-- limit exposure of protobuf types
+- limit exposure of protobuf types(this might be very difficult and cumbersome)
 - Attempt to separate reads/writes into separate methods in case they need more explicit isolation later.
   We do not need command/query type system but it is similar to that idea.
 
@@ -169,7 +171,7 @@ The proposal focusses on moving toward 3 patterns:
 - unify application state between main components
 
 
-### Example FastAPI setup
+#### Example FastAPI setup
 
 Currently, we setup our FastAPI services like this:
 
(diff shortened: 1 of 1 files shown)


proposals/records/4-python-refactoring.md
```diff
@@ -0,0 +1,306 @@
...
+1. Do not allow imports between our main components
+2. Isolate our data layers into separate interfaces
+3. Do not use global application state
+
+
+### Do not allow imports between our main components
+
+- `ingest` focuses on components for ingesting data and not providing interfaces to other components
+- shared tooling and data layers are moved to the `nucliadb.common` module
+- no main component should import from the `injest` module
```
jotare, line 155:
```suggestion
- no main component should import from the `ingest` module
```

proposals/records/4-python-refactoring.md
```diff
@@ -0,0 +1,306 @@
...
+- nucliadb should stop using the `ingest` grpc service
+
+### Isolate our data layers into separate interfaces
+
+- create data layer management components
+- limit direct exposure of k/v and blob storage drivers
+- limit exposure of protobuf types(this might be very difficult and cumbersome)
```
jotare, line 162: How are you planning to limit exposure with protobuf types?

proposals/records/4-python-refactoring.md
```diff
@@ -0,0 +1,306 @@
...
+async def get_kbs(prefix: str = "", app_context: AppContext) -> KnowledgeBoxList:
+    kb_data_manager = app_context.kb_data_manager
+    response = KnowledgeBoxList()
+    async for kbid, slug in kb_data_manager.list(prefix=prefix):
+        response.kbs.append(KnowledgeBoxObjSummary(slug=slug or None, uuid=kbid))
+    return response
+```
+
+- `app_context` is injected from the FastAPI runtime initialization but not global state
+- does not expose transation details
```
jotare, line 297:
```suggestion
- does not expose transaction details
```

jotare (commented):
I like it! A comment and a couple typos

proposals/records/4-python-refactoring.md
vangheem: This is the part of the proposal I feel least good about.

I feel like our use of pb is wrong but an alternative would be mapping into internal data types which is a lot of churn and maybe not worth promoting too much at this point. 

An example of this can be found here: https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/migrator/datamanager.py

the pb is only for storage but is not exposed

jotare: Oh, I see, that's really nice. Ideally we have it like this (and also on the Rust side, don't propagate protos down the stack of services).

In the current status of things, might be unfeasible though... As we talked some times, our public API is also bound to our internal protos, so I think it'd need a global rethink on what we want to expose, whether an API with highly nested structures is worth...

Maybe we can start the refactor "ignoring" this, and at some point work on it

lferran (approved):

PR #1060, opened 2023-06-29
Implement new training: image classifier

jotare:
### Description
Second part of PDF annotations: train and datasets

### How was this PR tested?
Integration test (wip)


nucliadb/nucliadb/train/generators/image_classifier.py
```diff
@@ -0,0 +1,186 @@
...
+    node: AbstractIndexNode,
+    shard_replica_id: str,
+) -> AsyncIterator[ImageClassificationBatch]:
+    request = StreamRequest()
+    request.shard_id.id = shard_replica_id
+    request.reload = True
+    batch = ImageClassificationBatch()
+    async for item in node.stream_get_fields(request):
+        rid = item.uuid
+        resource = await get_resource_from_cache(kbid, rid)
```
lferran, line 58: NIT: Does this mean that if the resource is not in the cache it will return None?
If not, the naming of this util could be improved to something more accurate like: `get_resource_from_cache_or_db` or `cached_get_resource`?

nucliadb/nucliadb/train/generators/image_classifier.py
```diff
@@ -0,0 +1,186 @@
...
+
+        page_selections = await get_page_selections(resource, field_key)
+        if len(page_selections) == 0:
+            # Generating a payload without annotations makes no sense
+            continue
+
+        field = await resource.get_field(field_key, field_type, load=True)
+        page_structure = await get_page_structure(field)
+
+        print("LEN DE PAGE STRCTURE:", len(page_structure))
```
lferran, line 77: print should removed

nucliadb/nucliadb/train/generators/image_classifier.py
```diff
@@ -0,0 +1,186 @@
...
+                    for selection in page_selections[page]
+                ],
+            }
+
+            ic = ImageClassification()
+            ic.uri = page_uri
+            ic.selections = json.dumps(pawls_payload)
+            batch.data.append(ic)
+
+            if len(batch.data) == trainset.batch_size:
```
lferran, line 122: What happens if batch_size is exceeded?

nucliadb/nucliadb/train/generators/image_classifier.py
```diff
@@ -0,0 +1,186 @@
...
+                yield batch
+                batch = ImageClassificationBatch()
+
+    if len(batch.data):
+        yield batch
+
+
+async def get_page_selections(
+    resource: Resource, field_key: str
```
lferran, line 131: Is field key the field id only? If so, I think it should include both field type and field id, as you can have the same field id in different field types.

nucliadb/nucliadb/train/generators/image_classifier.py
jotare: `batch.data` is incremented always by one, but I can add a `>=` condition

nucliadb/nucliadb/train/generators/image_classifier.py
jotare: Changed for `field` and properly checked field id and type

nucliadb/nucliadb/train/generators/image_classifier.py
lferran: ah, sorry I missed that! Then I guess it's totally fine. up to you!

lferran (approved):

PR #1061, opened 2023-06-29
Validate facets

lferran:
### Description
Tantivy does not support nested facets in the same query. This validation prevents users from getting a 500 error from the node and instead we validate as first step

### How was this PR tested?
Integration tests and local tests


lferran:
[sc-6160]

jotare (approved):

PR #1064, opened 2023-06-30
Use stored model min score

lferran:
### Description
New KBs created in the cloud offering are created specifying the default min score for the semantic model associated to that KB.
This PR aims to utilize this min score as default for all search operations. 

### How was this PR tested?
Integration and unit tests


nucliadb_models/nucliadb_models/search.py
```diff
@@ -45,6 +45,11 @@ class ModelParamDefaults:
         title="Autofilters",
         description="List of filters automatically applied to the search query",
     )
+    min_score = ParamDefault(
+        default=...,
+        title="Minimum score",
+        description="Minimum similarity score applied to the vector index search. Results with a lower score have been ignored.",  # noqa
```
jotare, line 51:
```suggestion
        description="Minimum similarity score used to filter vector index search. Results with a lower score will be ignored.",  # noqa
```

jotare (approved):

PR #1069, opened 2023-07-03
Fix chat context token counting

lferran:
### Description

### How was this PR tested?


bloodbare (approved):
I do not understand when this feature was added, still think that all logic to cut needs to be provided by NUA on their response.

PR #1072, opened 2023-07-04
Fix purge: actually delete shards

lferran:
### Description
Purge has not been cleaning up shards for a long time now, and it has left a big amount of orphaned replicas in the nodes.

### How was this PR tested?
Local tests (running purge command against stage cluster)


lferran:
[sc-6200]

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -379,15 +378,16 @@ async def purge(cls, driver: Driver, kbid: str):
             shards_obj = writer_pb2.Shards()
             shards_obj.ParseFromString(payload)  # type: ignore
 
-            if cluster_settings.standalone_mode:
-                await load_active_nodes()
+            await load_active_nodes()
```
lferran, line 381: This is. the actual fix: turns out that purge command does not run with the cluster monitor sidecar container, thus it will not be part of the chitchat cluster.

jotare (approved):

PR #1073, opened 2023-07-05
Removing noise from bm25 results

lferran:
### Description
Essentially discards all results from the paragraph index that are `factor` % away from the maximum score on that response.

### How was this PR tested?


lferran:
closing in favour of a better approach

PR #1090, opened 2023-07-11
Sdk feedback + fix bug in find

lferran:
### Description
We were setting always the same score for all vector results

### How was this PR tested?
Unit tests


nucliadb/nucliadb/search/search/find_merge.py
```diff
@@ -282,7 +285,7 @@ async def merge_paragraphs_vectors(
     for merged_paragraph in merged_paragrahs:
         if merged_paragraph.vector_index is not None:
             merged_paragraph.paragraph = FindParagraph(
-                score=vector.score,
+                score=merged_paragraph.vector_index.score,
```
lferran, line 288: this is the bug. We were setting the same score for all vector results

vangheem (approved):

PR #1095, opened 2023-07-12
Node reader HA

vangheem:
### Description
This PR provides a HA implementation for our node reader and provides a mechanism to decouple the reader from the writers when deployed with a managed node pool implementation.

The advantages here are that we have more freedom to scale reads separately from writes. Additionally, we can implement a mostly zero downtime node reader implementation and rollout node statefulset upgrades in a non-sequential, simultaneous way.


vangheem:
[sc-6313]

charts/nucliadb_node/templates/node.reader.deploy.yaml
```diff
@@ -0,0 +1,155 @@
...
+      securityContext: {}
+      volumes:
+        - name: nats-creds
+          secret:
+            defaultMode: 420
+            secretName: nats-nucliadb-creds
+        - name: regional-nats-creds
+          secret:
+            defaultMode: 420
+            secretName: nats-regional-creds
```
lferran, line 151: why are these needed?

vangheem: oh, it's just copy paste from existing deploy. I can remove

PR #1096, opened 2023-07-13
Do not return semantic results on exact match only queries

lferran:
### Description

### How was this PR tested?
Unit tests

### DIFF
--- a/nucliadb/nucliadb/search/search/utils.py
+++ b/nucliadb/nucliadb/search/search/utils.py
@@ -19,7 +19,13 @@
 #
 from fastapi import HTTPException
 
-from nucliadb_models.search import SearchRequest, SortField, SortOptions, SortOrder
+from nucliadb_models.search import (
+    BaseSearchRequest,
+    SearchRequest,
+    SortField,
+    SortOptions,
+    SortOrder,
+)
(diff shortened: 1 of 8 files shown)


nucliadb/nucliadb/search/search/utils.py
```diff
@@ -63,7 +69,33 @@ def parse_sort_options(item: SearchRequest) -> SortOptions:
...
+def has_user_vectors(request: BaseSearchRequest) -> bool:
+    return request.vector is not None and len(request.vector) > 0
+
+
+def is_exact_match_only_query(request: BaseSearchRequest) -> bool:
+    """
+    '"something"' -> True
+    'foo "something" else' -> False
+    """
+    query_split = request.query.strip().split('"')
```
vangheem, line 87: Would a better check just be something like:

```
query = request.query.strip()
return len(query) > 0 and query[0] == '"' and query[-1] == '"'
```

?

vangheem (approved):

nucliadb/nucliadb/search/search/utils.py
lferran: changed!

PR #1100, opened 2023-07-13
Auto load kubernetes cluster index nodes

vangheem:
### Description
The purpose of this PR is to provide a path forward to auto load cluster information from kubernetes directly instead of using chitchat.

#### Chitchat does not make a lot of sense for our system right now

Why?

The chitchat implementation assumes...
- all members of a cluster have equal value
- members should all know about each other and communication is multi directional between members
- we have a system where it makes sense for cluster members to be added/removed dynamically
- a node going down is self healing
- we aren't behind a system that already provides load balancing, namespace resolution, etc

When, in reality we have:
- A pretty static cluster
- It is known when members should be upgraded
- Kubernetes to load balance and handle pods coming in and out of service for us when we introduce multiple readers for a single node
- Chitchat implementation isn't very stable and has an awkward implementation that requires a sidecar
- Communication with index nodes is unidirectional
- Istio

### How was this PR tested?


charts/nucliadb_ingest/templates/ingest-processed-consumer.deploy.yaml
```diff
@@ -20,12 +20,13 @@ spec:
...
       name: ingest-processed-consumer
       annotations:
-        traffic.sidecar.istio.io/excludeInboundPorts: "{{.Values.serving.metricsPort }}"
+        traffic.sidecar.istio.io/excludeInboundPorts: "{{ .Values.serving.metricsPort }}"
         traffic.sidecar.istio.io/excludeOutboundPorts: "{{.Values.services.maindb }},{{.Values.services.nats }},20160,2380"
         # do not have access to dependency chart cm this component references
         checksum/cm: {{ include (print $.Template.BasePath "/ingest.cm.yaml") . | sha256sum }}
       labels:
         app: ingest-processed-consumer
+        metrics: "enabled"
```
lferran, line 29: What is the `metrics: enbled` label needed for?

charts/nucliadb_search/templates/search.sm.yaml
lferran, line 1: nice charts unification/cleanup! 👏🏼

nucliadb/nucliadb/metrics_exporter.py
```diff
@@ -0,0 +1,62 @@
...
+    all_nodes = manager.get_index_nodes()
+
+    for node in all_nodes:
+        SHARD_COUNT.set(node.shard_count, labels=dict(node=node.id))
+
+
+async def run_exporter():
+    while True:
+        await asyncio.sleep(10)
+        update_node_metrics()
```
lferran, line 44: maybe not for this PR, but the waiting time / interval for each metric in the exporter will be different right?

nucliadb/nucliadb/metrics_exporter.py
lferran, line 1: it´s nice to see how easy it is to bootstrap a new component now

lferran (approved):

nucliadb/nucliadb/common/cluster/discovery/k8s.py
```diff
@@ -0,0 +1,192 @@
...
+                )
+            else:
+                logger.debug(
+                    "Update node",
+                    extra={"sts_name": sts_name, "node_id": node_data.node_id},
+                )
+                node.address = node_data.address
+                node.shard_count = node_data.shard_count
+        else:
+            logger.debug(f"Remove node", extra={"node_id": node_data.node_id})
```
lferran, line 141: What do you think about WARNING log level here? It should not be as chatty, and we may want to be able confirm when a node is actually down

vangheem: Fine with me!

nucliadb/nucliadb/metrics_exporter.py
vangheem: Yes, possibly depending on what we're doing

PR #1101, opened 2023-07-14
Add migration to populate all fields key in maindb

lferran:
### Description
Old KBs do not have the all fields key in tikv populated, thus iterating on the resource fields means doing a tikv scan keys, which is very slow and error prone.

This PR adds a migration that will iterate over all resources and populate that key (if missing)

### How was this PR tested?
Tested manually against staging cluster

### DIFF
--- a/nucliadb/migrations/0003_allfields_key.py
+++ b/nucliadb/migrations/0003_allfields_key.py
@@ -1,0 +1,59 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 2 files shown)


nucliadb/migrations/0003_allfields_key.py
```diff
@@ -0,0 +1,60 @@
...
+    async for resource_id in rdm.iterate_resource_ids(kbid):
+        resource = await rdm.get_resource(kbid, resource_id)
+        if resource is None:
+            logger.warning(
+                f"kb={kbid} rid={resource_id}: resource not found. Skipping..."
+            )
+            continue
+
+        async with context.kv_driver.transaction() as txn:
+            resource.txn = txn
```
vangheem, line 44: This is really annoying with our API :(

You did it right, it just makes me sad

nucliadb/migrations/0003_allfields_key.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+                continue
+
+            # Migrate resource
+            logger.warning(f"kb={kbid} rid={resource_id}: migrating...")
+            all_fields = AllFieldIDs()
+            async for (field_type, field_id) in resource._scan_fields_ids():
+                fid = FieldID(field_type=field_type, field=field_id)
+                all_fields.fields.append(fid)
+            await resource.set_all_field_ids(all_fields)
```
vangheem, line 59: should we check if it's already set before committing?

vangheem (approved):
looks fine!

nucliadb/migrations/0003_allfields_key.py
lferran: we are checking it at line 46: https://github.com/nuclia/nucliadb/pull/1101/files#diff-28c87aea8a2094097113c5974966555e917197a9bc8c47c15cb36867703f4fb8R46

nucliadb/migrations/0003_allfields_key.py
lferran: yes... maybe all methods of the `orm.Resource` class be class methods and/or accept the txn as an argument.

PR #1104, opened 2023-07-17
Audit chat 

lferran:
### Description
The purpose of this PR is to send audit events for chat interactions on nucliadb so that clients can see the history of chat questions and answers

### How was this PR tested?
Unit tests


nucliadb/nucliadb/search/search/chat/query.py
```diff
@@ -79,59 +150,57 @@ async def chat(
...
+        ),
         media_type="plain/text",
         headers={
             "NUCLIA-LEARNING-ID": ident or "unknown",
             "Access-Control-Expose-Headers": "NUCLIA-LEARNING-ID",
         },
     )
+
+
+def audited_generator(
```
vangheem, line 172: hmmm, I'm not sure the wrapper approach is more readable than just doing the check and audit directly in the code like before

vangheem (approved):

nucliadb/nucliadb/search/search/chat/query.py
lferran: actually you are right, this is too complicated

PR #1105, opened 2023-07-17
Implement a better settings module index

jotare:
### Description
- Create a new module `settings` in `nucliadb_node` to improve how configuration is managed
- Replace some `env` methods for settings

### Missing
- Fully replace `env` with `settings` (some code uses `env` as global configuration)
- Propagate configuration to constructors (include bindings)

### How was this PR tested?
- Using existing tests


nucliadb_node/src/settings/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        self.sentry_url.clone()
+    }
+
+    /// Sentry environment to report errors
+    pub fn sentry_env(&self) -> &'static str {
+        self.sentry_env
+    }
+
+    /// Log levels. Every element is a crate-level pair
+    pub fn log_levels(&self) -> Vec<(String, Level)> {
```
hermeGarcia, line 135: Is a better practice to return `&[(String, Level)]` and let the caller decide how and when the allocation should be done.

nucliadb_node/src/settings/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        self.log_levels.clone()
+    }
+
+    /// Span levels. Every element is a crate-level pair
+    pub fn span_levels(&self) -> Vec<(String, Level)> {
```
hermeGarcia, line 140: Same, better to not clone in the function and return `&[...]`

nucliadb_node/src/settings/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub fn writer_listen_address(&self) -> SocketAddr {
+        self.writer_listen_address
+    }
+
+    pub fn chitchat_port(&self) -> u16 {
+        self.chitchat_port
+    }
+
+    /// List of known nodes to connect wqith in order to join the cluster
+    pub fn seed_nodes(&self) -> Vec<String> {
```
hermeGarcia, line 169: Better to return `&[String]`.

hermeGarcia (changes requested):

nucliadb_node/src/services/mod.rs
```diff
@@ -23,12 +23,42 @@ mod versions;
...
+
+    /// Path where all shards are stored
+    pub fn shards_path(data_path: &Path) -> PathBuf {
+        data_path.join(SHARDS_DIR)
+    }
+
+    /// Path where shard `id` is stored
+    pub fn shard_path_by_id(data_path: &Path, id: &str) -> PathBuf {
+        shards_path(data_path).join(id)
+    }
```
hermeGarcia, line 63: Why not add this to `SettingsProvider`? We can built those once we have the data path and we would reduce code duplication (this paths are also used in the writer) and improve in code modularity.

jotare: Even though this are derived from `data_path` and can be seen as configuration, all this values belong to index node's disk organization. Providing this as settings means users of multiple of this settings need to receive either a `Settings` instance or multiple settings.

Get this as an example: https://github.com/nuclia/nucliadb/blob/main/nucliadb_node/src/writer/mod.rs `NodeWriterService` or it's new replacement `UnboundedShardCache` would need to know about `data_path`, `shards_path` and `shard_path_by_id`. Passing all this to the `new` or passing a `Settings` instance seems too much

hermeGarcia (approved):

PR #1110, opened 2023-07-18
Delete deprecated POST /entitiesgroups/{group}

jotare:
### Description
- Delete deprecated endpoint POST `/entitiesgroups/{group}`
- Remove `set_entitygroup_entities` from SDKv2 (it shouldn't be there as the endpoint was deprecated)


vangheem (approved):

nucliadb/nucliadb/search/tests/node.py
```diff
@@ -208,6 +208,8 @@ def start(self):
         else:
             # Valid when using github actions
             docker_internal_host = "172.17.0.1"
+        # Always use IP
+        docker_internal_host = "172.17.0.1"
```
vangheem, line 212: This will not always work. I think we'll need to figure out another way.

vangheem (changes requested):

vangheem (approved):

PR #1118, opened 2023-07-22
POC standalone clustering support

vangheem:
### Description
This is a POC for supporting cluster with standalone server.

Right now, it's done in a silly way by just reusing the http service available to provide the communication.

Also, it's naively done when you want to query a shard that is already on the current node.

In any case, it's a POC and something to use for discussion.

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/common/cluster/discovery/abc.py
+++ b/nucliadb/nucliadb/common/cluster/discovery/abc.py
@@ -23,7 +23,6 @@ import logging
 
 from nucliadb.common.cluster import manager
 from nucliadb.common.cluster.discovery.types import IndexNodeMetadata
-from nucliadb.common.cluster.index_node import IndexNode
 from nucliadb.common.cluster.settings import Settings
 from nucliadb_telemetry import metrics
 
@@ -47,11 +46,9 @@ def update_members(members: list[IndexNodeMetadata]) -> None:
         if node is None:
             logger.debug(f"{member.node_id} add {member.address}")
             manager.add_index_node(
-                IndexNode(
-                    id=member.node_id,
-                    address=member.address,
-                    shard_count=shard_count,
-                )
+                id=member.node_id,
+                address=member.address,
+                shard_count=shard_count,
             )
             logger.debug("Node added")
         else:
--- a/nucliadb/nucliadb/common/cluster/settings.py
+++ b/nucliadb/nucliadb/common/cluster/settings.py
@@ -27,6 +27,7 @@ class ClusterDiscoveryMode(str, enum.Enum):
     CHITCHAT = "chitchat"
     MANUAL = "manual"
     KUBERNETES = "kubernetes"
+    STANDALONE = "standalone"
 
 
 class Settings(BaseSettings):
--- a/nucliadb/nucliadb/standalone/api_router.py
+++ b/nucliadb/nucliadb/standalone/api_router.py
@@ -17,17 +17,28 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import functools
+import inspect
 import logging
+from socket import gethostname
 
-from fastapi import Request
+import pydantic
+from fastapi import Request, Response
 from fastapi.responses import JSONResponse
 from fastapi.routing import APIRouter
 from fastapi_versioning import version
 
+from nucliadb.common.cluster import manager
+from nucliadb.common.cluster.standalone import grpc_node_binding
+from nucliadb.common.cluster.standalone.index_node import StandaloneIndexNode
+from nucliadb.common.cluster.utils import get_shard_manager as get_shard_manager
 from nucliadb.common.http_clients.processing import ProcessingHTTPClient
+from nucliadb.standalone.utils import get_standalone_node_id
 from nucliadb_models.resource import NucliaDBRoles
 from nucliadb_utils.authentication import requires
 from nucliadb_utils.settings import nuclia_settings
+
+from .settings import Settings
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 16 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Install test deps
+        run: pip install -r e2e/requirements.txt
+
+      - name: Run tests
+        run:
+          kubectl port-forward -n nucliadb service/nucliadb 8080:8080 &
+          kubectl logs -n nucliadb nucliadb-0
+          kubectl logs -n nucliadb nucliadb-1
+          pytest -s --tb=native e2e/test_e2e.py || true
+          kubectl logs -n nucliadb nucliadb-0
```
lferran, line 83: we could run this in case of failure! could be useful to debug what is the problem

vangheem: Definitely!

vangheem:
[sc-6730]

nucliadb/nucliadb/common/cluster/utils.py
```diff
@@ -27,23 +28,41 @@
...
-    return mng
+    async with _lock:
+        if get_utility(Utility.SHARD_MANAGER) is not None:
+            # already setup
+            return get_utility(Utility.SHARD_MANAGER)
+
+        await setup_cluster_discovery()
+        mng: Union[KBShardManager, StandaloneKBShardManager]
+        if settings.standalone_mode:
+            from .standalone.service import start_grpc
```
tarekziade, line 45: what is the motivation for the lazy import? That module does not seem to have any global init

charts/nucliadb/values.yaml
```diff
@@ -5,11 +5,16 @@ imagePullPolicy: IfNotPresent
 image: nuclia/nucliadb
 imageVersion: latest
 
+replicas: 1
```
lferran, line 8: `replicas: 1` means there will be two instances, right?

However we are recommending having at least two replicas in the docs:
https://github.com/nuclia/docs/pull/144/files#diff-17d23fcbec1f332640a5341afcac8f13b5f20bd6a272e34bea8ebaa6af8d8a8bR33

e2e/test_e2e.py
```diff
@@ -0,0 +1,149 @@
...
+
+def test_resource_processed(kbid: str, resource_id: str):
+    start = time.time()
+    while True:
+        resp = requests.get(
+            os.path.join(BASE_URL, f"api/v1/kb/{kbid}/resource/{resource_id}"),
+            headers={
+                "content-type": "application/json",
+                "X-NUCLIADB-ROLES": "READER",
+                "x-synchronous": "true",
```
lferran, line 86: not needed for the GET

e2e/test_e2e.py
```diff
@@ -0,0 +1,149 @@
...
+                "x-ndb-client": "web",
+            },
+            json={
+                "links": {"link": {"uri": "https://en.wikipedia.org/wiki/Cricket"}},
+                "usermetadata": {"classifications": []},
+                "title": "https://en.wikipedia.org/wiki/Cricket",
+                "icon": "application/stf-link",
+                "origin": {"url": "https://en.wikipedia.org/wiki/Cricket"},
```
lferran, line 94: not needed for the GET request

nucliadb/nucliadb/common/cluster/settings.py
```diff
@@ -18,24 +18,23 @@
...
 class ClusterDiscoveryMode(str, enum.Enum):
     MANUAL = "manual"
     KUBERNETES = "kubernetes"
+    SINGLE_NODE = "single_node"
 
 
 class Settings(BaseSettings):
     data_path: str = "./data/node"
     standalone_mode: bool = False
+    standalone_node_port: int = 10009
```
lferran, line 34: NIT: you could add pydantic `Field` annotations so that we get descriptions on settings fields

nucliadb/nucliadb/ingest/settings.py
```diff
@@ -69,7 +69,7 @@ class Settings(DriverSettings):
 
     # ingest consumer sts replica settings
     replica_number: int = -1
-    total_replicas: int = 1
+    total_replicas: int = 1  # number of ingest processor replicas in the cluster
```
lferran, line 72: nit: why not adding pydantic.Field annotations?

nucliadb/nucliadb/standalone/api_router.py
```diff
@@ -94,3 +95,29 @@ def get_temp_access_token(request: Request):
...
+    )
+
+
+@standalone_api_router.get("/health/alive")
+async def alive(request: Request) -> JSONResponse:
+    return JSONResponse({"status": "ok"})
+
+
+@standalone_api_router.get("/health/ready")
+async def ready(request: Request) -> JSONResponse:
```
lferran, line 122: WDYT about returning an error here if the nodes can't discover each other (ie: the cluster can't be formed)?

Dockerfile
```diff
@@ -45,14 +45,9 @@ RUN mkdir -p /data
 ENV NUA_ZONE=europe-1
 ENV NUA_API_KEY=
 ENV NUCLIA_PUBLIC_URL=https://{zone}.nuclia.cloud
-
+ENV PYTHONUNBUFFERED=1
```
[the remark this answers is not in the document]
vangheem: I've usually used this in the past because it's possible that you lose logs in a crash.

I think with the amount of logs we have, we should be okay!

nucliadb/nucliadb/train/utils.py
```diff
@@ -72,24 +72,26 @@ async def stop_train_grpc():
...
-    set_utility(
-        Utility.SHARD_MANAGER, TrainShardManager(driver=driver, storage=storage)
-    )
+    set_utility(_TRAIN_SM_NAME, TrainShardManager(driver=driver, storage=storage))
 
 
 async def stop_shard_manager():
-    if get_utility(Utility.SHARD_MANAGER):
-        clean_utility(Utility.SHARD_MANAGER)
+    if get_utility(_TRAIN_SM_NAME):
```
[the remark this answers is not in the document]
vangheem: I would like to do that as well! However, our lifecycle management is built all around startup/shutdown handling right now.

We could also make https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/common/context/__init__.py an async context manager and focus on refactoring everything.

nucliadb/nucliadb/common/cluster/discovery/abc.py
```diff
@@ -21,11 +21,20 @@
 import asyncio
 import logging
 
+import backoff
```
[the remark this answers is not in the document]
vangheem: Yeah, I guess this is "foolish consistency" from the parent module.

I'll fix.

nucliadb/nucliadb/common/cluster/standalone/utils.py
```diff
@@ -0,0 +1,65 @@
...
+import uuid
+from socket import gethostname
+
+from nucliadb.common.cluster.settings import settings as cluster_settings
+from nucliadb.common.cluster.standalone.index_node import StandaloneIndexNode
+
+logger = logging.getLogger(__name__)
+
+
+def get_standalone_node_id() -> str:
```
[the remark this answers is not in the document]
vangheem: Yup! `uuid.uuid4().bytes` is not the same as the string representation.

And on the rust side, we store the bytes version, not the string version.

nucliadb/nucliadb/common/cluster/standalone/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    if _SELF_INDEX_NODE is None or node_id != _SELF_INDEX_NODE.id:
+        if "NUCLIADB_SERVICE_HOST" in os.environ:
+            hn = os.environ["HOSTNAME"]
+            ns = os.environ.get("NAMESPACE", "nucliadb")
+            host = f"{hn}.{ns}"
+        else:
+            host = gethostname()
+        _SELF_INDEX_NODE = StandaloneIndexNode(id=node_id, address=host, shard_count=0)
+    try:
+        # XXX this is weird right? How silly is this?
```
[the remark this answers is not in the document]
vangheem: ah, nice. I can make it work in the same way.

charts/nucliadb/values.yaml
vangheem: Agreed. I'll make the default 2!

lferran (approved):

tarekziade (approved):

PR #1119, opened 2023-07-24
Do not audit processor messages

lferran:
### Context
The goal of the audit log is to audit user actions.

When a user modifies a resource/field, there will be two messages processed through ingest, in the following order:
- One coming from the writer component: with the metadata that the user patched.
- One coming from the processing pipeline: with the corresponding extracted/processed metadata.

Right now, both messages are audited, therefore having two logs for the same user interaction.

### Description
This PR makes processor messages not being audited

### How was this PR tested?
Unit tests

### DIFF
--- a/nucliadb/nucliadb/ingest/consumer/auditing.py
+++ b/nucliadb/nucliadb/ingest/consumer/auditing.py
@@ -237,6 +237,8 @@ class ResourceWritesAuditHandler:
         self, message: writer_pb2.BrokerMessage
     ) -> list[audit_pb2.AuditField]:
         if message.type == writer_pb2.BrokerMessage.MessageType.DELETE:
+            # If we are fully deleting a resource we won't iterate the delete_fields (if any).
+            # Make no sense as we already collected all resource fields as deleted
             return []
 
         audit_storage_fields: list[audit_pb2.AuditField] = []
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/ingest/consumer/auditing.py
```diff
@@ -291,13 +289,17 @@ async def handle_message(self, raw_data) -> None:
             metrics.total_messages.inc({"action": "ignored", "type": "audit_fields"})
             return
 
+        message = notification.message
+        if message.source == message.MessageSource.PROCESSOR:
+            metrics.total_messages.inc({"action": "ignored", "type": "audit_fields"})
+            return
+
```
lferran, line 296: This is the actual fix

nucliadb/nucliadb/ingest/consumer/auditing.py
```diff
@@ -236,49 +236,47 @@ def iterate_auditable_fields(
     async def collect_audit_fields(
         self, message: writer_pb2.BrokerMessage
     ) -> list[audit_pb2.AuditField]:
+        if message.type == writer_pb2.BrokerMessage.MessageType.DELETE:
```
lferran, line 239: the changes in the `collect_audit_fields` method are just checking for the message type at the beginning of the method to avoid opening and closing a transaction for nothing.

bloodbare (approved):

PR #1124, opened 2023-07-25
remove advance query from external API

hermeGarcia:
### Description
Remove advance query from external API

### How was this PR tested?
Local tests


jotare (changes requested):
You have to remove advanced query also from `nucliadb_models/search.py`, remove the integration `test_synonyms.py::test_search_errors_if_advanced_query`, from the endpoint parameters of search (in the same find and search files you have modified and remove the documentation from the SDKv2.

Maybe we should add a deprecation warning in the SDKv2 too

PR #1132, opened 2023-07-26
Store KB creation time

lferran:
### Description
Stores the creation date in the KB config object

### How was this PR tested?
Integration tests


jotare (approved):
What's the use case?

lferran:
> What's the use case?
There may be more use cases but the one I am thinking right now is:
It is useful when having to run data migrations. You can short-circuit the migration time if you know that the KB was created after a particular point in time because the fix had been deployed already

PR #1136, opened 2023-07-27
add simple VM install script

vangheem:
### Description
A simple VM install script that we can reference with `curl <download loc> | bash` to be able to install NucliaDB in a VM easily.

### How was this PR tested?
Manual


lferran (approved):
lgtm! 👍🏼 

NOTE: If we are going to have more on-prem users, I think it'd be safer if we start maintaining a `stable` version that all our docs / scripts point to. WDYT?

vangheem:
Maybe we could update the script to only install some "stable" tag by default?

PR #1147, opened 2023-07-28
Standalone temp token support

vangheem:
### Description

### How was this PR tested?


nucliadb/nucliadb/standalone/auth.py
```diff
@@ -153,12 +188,30 @@ async def authenticate(
...
+        NucliaCloudAuthenticationBackend.__init__(
+            self,
             roles_header=settings.auth_policy_roles_header,
             user_header=settings.auth_policy_user_header,
         )
+
+    async def authenticate(
+        self, request: HTTPConnection
+    ) -> Optional[Tuple[AuthCredentials, BaseUser]]:
+        token_resp = await AuthTokenAuthenticationBackend.authenticate(self, request)
```
lferran, line 205: NIT: I think this pattern is better than the one above (inheritance) because it's more explicit.

vangheem: I think you're right. I will tweak the PR to make it use less inheritance.

PR #1162, opened 2023-08-02
Fix parsing rephrase response so that it does not include " characters

lferran:
### Description
This was making the whole rephrase feature not work, because it was causing an exact match search on the successive find request to retrieve context for chat.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb/nucliadb/search/predict.py
+++ b/nucliadb/nucliadb/search/predict.py
@@ -262,7 +262,7 @@ class PredictEngine:
             headers=self.get_predict_headers(kbid),
         )
         await self.check_response(resp, expected=200)
-        return await resp.text()
+        return await resp.json()
 
     @predict_observer.wrap({"type": "chat"})
     async def chat_query(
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/search/predict.py
```diff
@@ -262,7 +262,7 @@ async def rephrase_query(self, kbid: str, item: RephraseModel) -> str:
             headers=self.get_predict_headers(kbid),
         )
         await self.check_response(resp, expected=200)
-        return await resp.text()
+        return await resp.json()
```
vangheem, line 265: Isn't this going to return a dict? This changes the types expected.

vangheem (changes requested):

nucliadb/nucliadb/search/predict.py
lferran: The problem is that learning is returning a json response, but the payload is simply a string (which is json compatible, so there is nothing wrong...)

But the fact that we were parsing it as text, it made the string be wrapped in quotes...

vangheem (approved):

PR #1177, opened 2023-08-04
trigger sdk test on stage after publishing a new version

ebrehault:
### Description
After publishing a new version on pypi, we will trigger the stage integration tests in the `nuclia.py` repo.

(Not sure yet how we will notify the user in case of failure)

### How was this PR tested?


tarekziade (approved):
LGTM

PR #1182, opened 2023-08-07
Simplify the shard counting 

tarekziade:
### Description

Removes the metadata static file to avoid read/write edge cases.

### How was this PR tested?

This is mostly removing code, so the current tests are good enough


hermeGarcia (approved):
LGTM!

PR #1185, opened 2023-08-08
Retry client connection errors on predict and processing api clients

lferran:
### Description
We have seen occasional client connection errors (temporary failed name resolution).
This PR adds retry logic for these errors on all interactions with predict api and processing api

### How was this PR tested?
I don't think it is worth adding a test for the backoff retry decorator, and existing tests should cover any regressions.

### DIFF
--- a/nucliadb/nucliadb/search/predict.py
+++ b/nucliadb/nucliadb/search/predict.py
@@ -84,6 +84,7 @@ predict_observer = metrics.Observer(
 
 
 RETRIABLE_EXCEPTIONS = (aiohttp.client_exceptions.ClientConnectorError,)
+MAX_TRIES = 2
 
 
 async def start_predict_engine():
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/search/predict.py
```diff
@@ -218,6 +222,11 @@ async def check_response(self, resp, expected: int = 200) -> None:
         else:
             raise SendToPredictError(f"{resp.status}: {await resp.read()}")
 
+    @backoff.on_exception(backoff.expo, RETRIABLE_EXCEPTIONS, max_tries=2)
```
tarekziade, line 225: nit: I would make `2` a `MAX_TRIES` constant at the top of the module

tarekziade (approved):
LGTM just one nit

PR #1187, opened 2023-08-08
make sure we properly finalize

tarekziade:
### Description
The test was leaking a task

### How was this PR tested?

```
cd nulciadb_node; make test
```
-> no more `Task was destroyed but it is pending!`


vangheem (approved):
yay! thank you!

PR #1189, opened 2023-08-08
Better garbage collection scheduling

vangheem:
### Description
Right now, on active nodes, garbage collection might not ever happen.

This PR forces garbage collecting when more than 10 resources have been indexed on a shard and also allows implements a lock for every shard.

### How was this PR tested?


tarekziade (approved):
neat! LGTM

PR #1200, opened 2023-08-09
Cache implementation refactor

vangheem:
### Description

- Avoid supporting multiple types of cache implementations
- Avoid supporting multiple types of pubsub implementations
- Do not use in-memory cache implementation for counters
- implement number of resources count as a materialized value in kv storage

### How was this PR tested?


nucliadb/nucliadb/ingest/consumer/materializer.py
```diff
@@ -0,0 +1,89 @@
...
+        data = self.pubsub.parse(raw_data)
+        notification = writer_pb2.Notification()
+        notification.ParseFromString(data)
+
+        if notification.action != writer_pb2.Notification.Action.COMMIT:
+            return
+
+        self.task_handler.schedule(
+            notification.kbid, partial(self.process, notification.kbid)
+        )
```
lferran, line 83: Maybe it's not the case but, could it be that we are scheduling this twice on resource creation?
Not sure if it's available but this could be scheduled only if the message comes from the writer.

Also deletions should schedule the materializer

nucliadb/nucliadb/common/datamanagers/resources.py
```diff
@@ -100,3 +102,34 @@ async def get_resource_index_message(
             if res is None:
                 return None
             return (await res.generate_index_message()).brain
+
+    async def calculate_number_of_resources(self, kbid: str) -> int:
+        """
+        Calculate and the number of resources in a knowledgebox.
```
lferran, line 108:
```suggestion
        Calculate the number of resources in a knowledgebox.
```

lferran (approved):

nucliadb/nucliadb/ingest/consumer/materializer.py
vangheem: `writer_pb2.Notification.Action.COMMIT` action happens for both create and delete.

also, if many events happen at the same time, it's okay, it is done with a delayed handler and unique against the kb id.

The point of the implementation is to consolidate all potential changes into 1 action at some interval. For example, 100 writes in 30 seconds on one kb will only cause 1 re-calc to happen.

vangheem: my bad, I see what you mean now. I've updated it to only recalc on create/delete and not modified. That should do the trick.

PR #1201, opened 2023-08-10
Prevent icon from being modified once is set

lferran:
### Description
The `basic.icon` field is mainly used by the dashboard to display what kind of document a resource is (pdf, text, link, doc, excel sheet, etc).
If the user does not set its value on the create resource request, we infer it from the mimetype of its fields.

This PR aims to prevent the icon field from being updated once it has been set.

### How was this PR tested?
Integration test

### DIFF
--- a/nucliadb/nucliadb/ingest/orm/resource.py
+++ b/nucliadb/nucliadb/ingest/orm/resource.py
@@ -129,6 +129,10 @@ PB_TEXT_FORMAT_TO_MIMETYPE = {
     FieldText.Format.JSON: "application/json",
 }
 
+BASIC_IMMUTABLE_FIELDS = [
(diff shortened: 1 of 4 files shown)


nucliadb/nucliadb/ingest/orm/resource.py
```diff
@@ -129,6 +129,10 @@
     FieldText.Format.JSON: "application/json",
 }
 
+BASIC_IMMUTABLE_FIELDS = [
```
tarekziade, line 132: nit: if this `list` is constant and not going to be changed by some code at runtime, use a `tuple` instead

tarekziade (approved):
LGTM, just one nit

PR #1205, opened 2023-08-10
dont rely on the fs for vectorset updates

hermeGarcia:
### Description
Relying on the file system for identifying now or deleted vector sets can be problematic due to data races. This PR fixes it by assuming the cost of reloading the vector set list (usually VERY small) even when it has not changed.

### How was this PR tested?
Local tests.


tarekziade (approved):
LGTM

PR #1215, opened 2023-08-14
Purge small fixes

lferran:
### Description
- Fix k8s connection leak. (We were seeing some `Unclosed client session ...` warning logs)
- Give some time to k8s cluster discovery to discover the nodes after initialization.
- Fix connection leak on writer/tus/gcs.py (`self.session` was assigned twice in the `__init__`)

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb/nucliadb/common/cluster/discovery/k8s.py
+++ b/nucliadb/nucliadb/common/cluster/discovery/k8s.py
@@ -143,38 +143,41 @@ class KubernetesDiscovery(AbstractClusterDiscovery):
             kubernetes_asyncio.config.load_incluster_config()
         else:
             await kubernetes_asyncio.config.load_kube_config()
-        v1 = kubernetes_asyncio.client.CoreV1Api()
 
-        watch = kubernetes_asyncio.watch.Watch()
-        try:
-            while True:
-                try:
-                    async for event in watch.stream(
-                        v1.list_namespaced_pod,
-                        namespace=self.settings.cluster_discovery_kubernetes_namespace,
-                        label_selector=self.settings.cluster_discovery_kubernetes_selector,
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/common/cluster/discovery/k8s.py
```diff
@@ -202,6 +205,8 @@ async def initialize(self) -> None:
         self.update_node_data_cache_task = asyncio.create_task(
             self.update_node_data_cache()
         )
+        # Give some time to initialize and discover nodes
+        await asyncio.sleep(2)
```
tarekziade, line 209: nit: adding an arbitrary time can often lead to issues when this runs in a slower environment. 
If possible, this should be replaced by an event-based waiter. e.g. is there a way for this code to check that the node discovery is over?

nucliadb/nucliadb/writer/tus/gcs.py
```diff
@@ -109,7 +109,6 @@ async def initialize(
         self.upload_url = (
             object_base_url + "/upload/storage/v1/b/{bucket}/o?uploadType=resumable"
         )  # noqa
-        self.session = aiohttp.ClientSession()
```
tarekziade, line 112: good catch!

tarekziade (approved):

nucliadb/nucliadb/common/cluster/discovery/k8s.py
lferran: @tarekziade good point indeed.

I've added a more "adaptative" way of checking it...
The thing is we don't know how many nodes are expected to be discovered.
However, IMO this is way better that not checking / waiting at all, which is what we had before.
WDYT?

PR #1217, opened 2023-08-14
Retry timeout errors for tikv

lferran:
### Description
TIKV server is slow In the staging environment, and we get occasional timeout errors on search, leading to HTTP500 errors.
This PR handles those timeouts to tikv by retrying a couple of times.

[SEARCH-E0](https://nuclia.sentry.io/issues/4387430811/?referrer=Shortcut)
[sc6785]

### How was this PR tested?
unit tests

### DIFF
--- a/nucliadb/nucliadb/common/maindb/tikv.py
+++ b/nucliadb/nucliadb/common/maindb/tikv.py
@@ -22,6 +22,8 @@ from __future__ import annotations
 import logging
 from typing import Any, List, Optional
 
+import backoff
+
 from nucliadb.common.maindb.driver import (
     DEFAULT_BATCH_SCAN_LIMIT,
     DEFAULT_SCAN_LIMIT,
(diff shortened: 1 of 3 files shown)


nucliadb/nucliadb/common/maindb/tikv.py
```diff
@@ -80,9 +84,17 @@ async def batch_get(self, keys: List[str]):
...
 
+    @backoff.on_exception(backoff.expo, (TimeoutError,), max_tries=2)
     async def get(self, key: str) -> Optional[bytes]:
         with tikv_observer({"type": "get"}):
-            return await self.txn.get(key.encode())
+            try:
+                return await self.txn.get(key.encode())
+            except Exception as exc:
+                exc_text = str(exc)
+                if "4-DEADLINE_EXCEEDED" in exc_text:
```
tarekziade, line 94: nit: if possible let's not rely on the message text and use the constant (I assume you have that somewehere in the exc object)

```
>>> import grpc
>>> grpc.StatusCode.DEADLINE_EXCEEDED
<StatusCode.DEADLINE_EXCEEDED: (4, 'deadline exceeded')>
```

vangheem: The client is funky. It's rust grpc under the hood but only throws raw base `Exception` errors with only strings stuffed inside

tarekziade (approved):

nucliadb/nucliadb/common/maindb/tikv.py
lferran: let me put up a comment! sorry, should have mentioned hehe

PR #1221, opened 2023-08-15
WIP: Unified index GRPC and replication interface

vangheem:
### Description

### How was this PR tested?


vangheem:
[sc-6787]

nucliadb_protos/unified.proto
```diff
@@ -0,0 +1,92 @@
...
+
+    // Shard management RPCs
+    rpc CreateShard(CreateShardRequest) returns (CreateShardResponse) {}
+    rpc DeleteShard(DeleteShardRequest) returns (DeleteShardResponse) {}
+
+    // Indexing RPCs
+    rpc Index(IndexRequest) returns (IndexResponse) {}
+    rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+    rpc Search(SearchRequest) returns (SearchResponse) {}
+}
```
jotare, line 92: If there are clearly defined groups of methods, does it make sense to split them in different service definitions?

vangheem: hmm, maybe. In practice, they are still a single server interface so it seems more simple to just keep them on one service definition

PR #1222, opened 2023-08-15
Standalone support for gc and shard splitting

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/common/cluster/manager.py
+++ b/nucliadb/nucliadb/common/cluster/manager.py
@@ -35,9 +35,9 @@ from nucliadb.common.cluster.exceptions import (
     ShardsNotFound,
 )
 from nucliadb.common.datamanagers.cluster import ClusterDataManager
+from nucliadb.common.datamanagers.kb import KnowledgeBoxDataManager
 from nucliadb.common.maindb.driver import Transaction
 from nucliadb.common.maindb.utils import get_driver
-from nucliadb.ingest.orm.knowledgebox import KnowledgeBox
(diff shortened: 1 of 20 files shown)


vangheem:
[sc-6765]

nucliadb/nucliadb/common/cluster/manager.py
```diff
@@ -310,8 +316,53 @@ async def add_resource(
...
+    ) -> None:
+        if (node_id, shard_id) not in self._change_count:
+            self._change_count[(node_id, shard_id)] = 0
+        self._change_count[(node_id, shard_id)] += 1
+        if self._change_count[(node_id, shard_id)] < self.max_ops_before_checks:
+            return
+
+        self._change_count[(node_id, shard_id)] = 0
+        async with self._lock:
+            index_node = get_index_node(node_id)
```
lferran, line 357: IIRC, for standalone this will return a proxy index node and all the following method calls will be proxied to the `StandaloneClusterServiceStub` of each "standalone" node right?

Do you think it would be better to be more explicit with this? Either with the variable name `index_node` -> `proxy_index_node`, or with type hints `index_node: ProxyIndexNode`, or with a comment?

lferran (approved):
LGTM 👍🏼 Thanks for the refactoring

nucliadb/nucliadb/common/cluster/manager.py
vangheem: I like it. added.

PR #1226, opened 2023-08-16
Fix lock gc

hermeGarcia:
### Description

### How was this PR tested?


tarekziade (approved):
LGTM - from our discussions, it would be nice if we could have a benchmark or a test that would be able to simulate garbage collections being done during high loads

PR #1227, opened 2023-08-16
have ListShards list all the shards in a node

vangheem:
### Description
Current ListShards grpc service only lists shards in active cache, not actually all the shards on the node.

### How was this PR tested?
unit

### DIFF
--- a/nucliadb_node/src/grpc/grpc_writer.rs
+++ b/nucliadb_node/src/grpc/grpc_writer.rs
@@ -39,6 +39,7 @@ use crate::shards::metadata::ShardMetadata;
 use crate::shards::providers::unbounded_cache::AsyncUnboundedShardWriterCache;
 use crate::shards::providers::AsyncShardWriterProvider;
 use crate::shards::writer::ShardWriter;
+use tokio::fs;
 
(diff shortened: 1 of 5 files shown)


vangheem:
[sc-6793]

nucliadb_node/src/grpc/grpc_writer.rs
```diff
@@ -169,13 +170,22 @@ impl NodeWriter for NodeWriterGRPCDriver {
...
+            let entry_path = entry.path();
+            if entry_path.is_dir() {
+                shard_ids.push(ShardId {
+                    id: entry_path
+                        .file_name()
+                        .unwrap()
+                        .to_str()
+                        .unwrap()
+                        .to_string(),
+                })
```
hermeGarcia, line 185: To avoid unwrapping on strings taken from disk I would do the following:
```rust
if entry_path.is_dir() {
         let Some(id) = entry_path.file_name().map(|s| s.to_str().map(String::from)) else {
                continue;
          };
          shard_ids.push(ShardId { id });
}
```

hermeGarcia (approved):
just a comment

PR #1247, opened 2023-08-23
FST integration into DataPoint

tarekziade:
### Description

Implements a brute-force filtering search using FSTs.

### How was this PR tested?

Unit tests, functional tests, a new perf benchmark


nucliadb_vectors/src/data_point/mod.rs
```diff
@@ -484,6 +574,19 @@ impl DataPoint {
...
+        let label_index: Option<LabelIndex>;
+        let key_index: Option<KeyIndex>;
+
+        if LabelIndex::exists(&fst_dir) && KeyIndex::exists(&fst_dir) {
+            label_index = Some(LabelIndex::open(&fst_dir)?);
+            key_index = Some(KeyIndex::open(&fst_dir)?);
+        } else {
+            label_index = None;
+            key_index = None;
+        }
```
hermeGarcia, line 588: Usually prefer: 
```rust
        let (label_index, key_index) = if LabelIndex::exists(&fst_dir) && KeyIndex::exists(&fst_dir)
        {
            (
                Some(LabelIndex::open(&fst_dir)?),
                Some(KeyIndex::open(&fst_dir)?),
            )
        } else {
            (None, None)
        };
```

hermeGarcia (changes requested):

nucliadb_node/src/shards/metadata.rs
```diff
@@ -67,8 +68,12 @@ impl From<ShardMetadata> for GrpcMetadata {
 }
 impl From<NewShardRequest> for ShardMetadata {
     fn from(value: NewShardRequest) -> Self {
-        // TODO: set channel in protos
-        let channel = Channel::EXPERIMENTAL;
+        let channel = if env::get("NUCLIADB_CHANNEL").unwrap_or("STABLE").into() == "EXPERIMENTAL" {
+            Channel::EXPERIMENTAL
+        else {
+            Channel::STABLE
+        }
```
hermeGarcia, line 75: IDK why this compiles on the GA, its missing a `;` at the end

hermeGarcia (changes requested):

nucliadb_node/src/shards/metadata.rs
```diff
@@ -55,6 +56,7 @@ impl From<Similarity> for VectorSimilarity {
 pub struct ShardMetadata {
     kbid: Option<String>,
     similarity: Option<Similarity>,
+    pub channel: Channel,
 }
```
hermeGarcia, line 60: there are already created shards  that do not have `channel` on their metadata so there are two choices here:
```rust
pub struct ShardMetadata {
    kbid: Option<String>,
    similarity: Option<Similarity>,
     #[serde(default)] 
    pub channel: Channel, // now if the value is not present it will use the default
}
```
or
```rust
pub struct ShardMetadata {
    kbid: Option<String>,
    similarity: Option<Similarity>,
    channel: Option<Channel>, // now if the value is not present it will use None
}
// In this approach a getter should be define 
```
The approach I like most is the first one, in fact maybe `similarity` should do the same

hermeGarcia (changes requested):

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -203,10 +215,16 @@ impl ShardWriter {
             num_threads: num_cpus::get(),
         };
 
+        let channel = match metadata.channel {
+            Some(channel) => channel,
+            _ => Channel::default(),
+        };
+
```
hermeGarcia, line 222:
```rust
let channel = metadata.channel.unwrap_or_default();
```

nucliadb_vectors/src/fst_index/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            if let Some(offset) = self.fst_map.get(label) {
+                let _ = records.seek(io::SeekFrom::Start(offset))?;
+                let label_struct: Label = bincode::deserialize_from(&mut *records)?;
+                let set: HashSet<u64> = label_struct.node_addresses.iter().cloned().collect();
+                if index == 0 {
+                    results.extend(&set);
+                } else {
+                    results = results.intersection(&set).cloned().collect();
+                }
+            }
```
hermeGarcia, line 119: If a label is not in the map shouldn't the intersection be empty?

nucliadb_vectors/src/data_point/mod.rs
```diff
@@ -103,6 +107,96 @@ impl Address {
...
+    ) -> FormulaFilter<'a> {
+        let matching_nodes: HashSet<u64>;
+        let empty_formula: bool;
+        let use_fst: bool;
+
+        if key_index.is_none() && label_index.is_none() {
+            empty_formula = true;
+            matching_nodes = HashSet::new();
+            use_fst = false;
+        } else {
```
hermeGarcia, line 133: Instead of checking with an if, and unwrapping in the `else` branch I think it would be cleaner to use:
```rust
let (Some(key_index), Some(label_index)) = (key_index, label_index) else {
         return FormulaFilter {
                filter,
                total_nodes,
                matching_nodes: HashSet::new(),
                empty_formula: true,
                use_fst: false,
         };
};
```
Less indentation on the rest of the code and no unwraps

hermeGarcia (changes requested):
Good work! Just a couple of things

nucliadb_node/src/shards/shard_reader.rs
```diff
@@ -176,10 +177,16 @@ impl ShardReader {
             num_threads: env::num_paragraph_search_threads(),
         };
 
+        let channel = match metadata.channel {
+            Some(channel) => channel,
+            _ => Channel::default(),
+        };
+
```
hermeGarcia, line 184:
```rust
let channel = metadata.channel.unwrap_or_default();
```

hermeGarcia (commented):
missed one

nucliadb_vectors/src/fst_index/mod.rs
tarekziade: good catch! fixed (+one test)

hermeGarcia (approved):
LGTM!

vectors_benchmark/src/binaries/1m_stats.rs
```diff
@@ -51,19 +54,16 @@ macro_rules! measure_time {
 pub struct Args {
     /// Size of index
     #[clap(short, long, default_value_t = INDEX_SIZE)]
-    index_size: usize,
-    /// Size of batch
+    index: usize,
+    /// Size of index
     #[clap(short, long, default_value_t = BATCH_SIZE)]
-    batch_size: usize,
```
tarekziade, line 57: XXX fix merge

PR #1250, opened 2023-08-23
Proc macros crate and metrics measure macro

jotare:
### Description
- Add a new crate: `nucliadb_procs` for procedural macros
- Add metrics measure proc macro 

### How was this PR tested?

[sc-5460]

### DIFF
--- a/nucliadb_procs/src/lib.rs
+++ b/nucliadb_procs/src/lib.rs
@@ -1,0 +1,141 @@
+// Copyright (C) 2021 Bosutech XXI S.L.
+//
+// nucliadb is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at info@nuclia.com.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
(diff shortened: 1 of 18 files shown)


nucliadb_procs/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+            use std::time::SystemTime;
+            use nucliadb_core::metrics;
+            use nucliadb_core::metrics::request_time::RequestTimeKey;
+
+            let time = SystemTime::now();
+
+            // execute function body
+            let return_value = #block;
+
+            let took = time.elapsed().map(|elapsed| elapsed.as_secs_f64()).unwrap_or(f64::NAN);
```
tarekziade, line 65: it would be nice to be able to pass to the macro the desired unit. micro second, milliseconds or seconds

tarekziade: oh mm maybe prometheus takes only seconds in the current code, so nevermind. that said, I would defintiely use that macro in our benchmarks without sending the metrics just to get the value.

jotare: We have different implementation of meters (trait to collect metrics). As the macro is using `metrics::get_metrics` we only need to set the desired meter and it'll printed instead of sent to Prometheus, for example (`ConsoleMeter` instead of `PrometheusMeter`)

tarekziade (approved):
LGTM

PR #1253, opened 2023-08-25
implement node reader health check

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/grpc/grpc_reader.rs
+++ b/nucliadb_node/src/grpc/grpc_reader.rs
@@ -32,14 +32,14 @@ use crate::shards::providers::AsyncShardReaderProvider;
 use crate::shards::reader::ShardReader;
 
 pub struct NodeReaderGRPCDriver {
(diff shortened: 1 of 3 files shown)


nucliadb_node/src/grpc/grpc_reader.rs
```diff
@@ -32,14 +32,14 @@ use crate::shards::providers::AsyncShardReaderProvider;
 use crate::shards::reader::ShardReader;
 
 pub struct NodeReaderGRPCDriver {
-    shards: AsyncUnboundedShardReaderCache,
+    pub shards: Arc<AsyncUnboundedShardReaderCache>,
```
jotare, line 35: It's a bad practice to publish fields like this, as you are creating dependencies between layers of abstraction knowing too many implementation details.

The cache provider can be passed to `NodeReaderGRPCDriver`'s constructor, so you don't need to expose internal fields.

PR #1268, opened 2023-08-29
Add tokio runtime metrics for prometheus meter

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/metrics/meters/prometheus.rs
+++ b/nucliadb_core/src/metrics/meters/prometheus.rs
@@ -17,20 +17,33 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+use std::sync::Mutex;
+
 use prometheus_client::encoding;
 use prometheus_client::registry::Registry;
+use tokio_metrics::{RuntimeIntervals, RuntimeMonitor};
 
 use crate::metrics::meters::Meter;
-use crate::metrics::metric::tokio_tasks::TaskLabels;
-use crate::metrics::metric::{request_time, tokio_tasks};
+use crate::metrics::metric::request_time::{RequestTimeKey, RequestTimeMetric, RequestTimeValue};
+use crate::metrics::metric::tokio_runtime::TokioRuntimeMetrics;
+use crate::metrics::metric::tokio_tasks::{TaskLabels, TokioTaskMetrics};
+use crate::metrics::metric::{request_time, tokio_runtime, tokio_tasks};
 use crate::metrics::task_monitor::{Monitor, MultiTaskMonitor, TaskId};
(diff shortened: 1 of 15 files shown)


nucliadb_core/src/metrics/meters/prometheus.rs
```diff
@@ -69,20 +82,46 @@ impl Meter for PrometheusMeter {
...
+                        "Cannot export tokio runtime metrics, iterator didn't return values"
+                    ));
+                }
+            }
+            Err(error) => {
+                return Err(node_error!(
+                    "Cannot acquire runtime metrics lock. There's a concurrent export going on? \
+                     {error}"
+                ));
+            }
```
hermeGarcia, line 123: A `try_lock` can also return a `PoisonError`, which I think should be treated differently, in this case I would just get the inner lock using `into_inner` and collect either way.
[More info](https://doc.rust-lang.org/std/sync/enum.TryLockError.html) 
```rust
match self.runtime_intervals.try_lock() {
       Ok(..) => {..}
       Err(TryLockError::Poisoned(inner)) => { I would get the inner lock and collect }
       Err(TryLockError::WouldBlock) => { return the error if you want} 
}

File: nucliadb_core/src/metrics/metric/tokio_runtime.rs [hunk 2]
```diff
@@ -0,0 +1,444 @@
...
+        //     poll_count_histogram.clone(),
+        // );
+
+        let total_noop_count = Gauge::default();
+        registry.register(
+            "nucliadb_node_total_noop_count",
+            "The number of times worker threads unparked but performed no work before parking \
+             again. [...]",
+            total_noop_count.clone(),
+        );
```
hermeGarcia, line 139: Shouldn't this be a  `Counter`? Based on the description it seems like is something that can only increase right?

nucliadb_core/src/metrics/metric/tokio_runtime.rs
```diff
@@ -0,0 +1,444 @@
...
+            "The minimum number of times any on worker saturated its local queue",
+            min_overflow_count.clone(),
+        );
+
+        let total_polls_count = Gauge::default();
+        registry.register(
+            "nucliadb_node_total_polls_count",
+            "The number of tasks that have been polled across all worker threads. [...]",
+            total_polls_count.clone(),
+        );
```
hermeGarcia, line 253: Based on the description, this seems like only increases so maybe should be a `Counter`

hermeGarcia (approved):

nucliadb_core/src/metrics/meters/prometheus.rs
jotare: Solved

tarekziade (approved):

PR #1270, opened 2023-08-29
thread dump service rust

hermeGarcia:
### Description

Adds the `/__dump` endpoint to the rust HTTP service to help debugging.

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/http_server/mod.rs
+++ b/nucliadb_node/src/http_server/mod.rs
@@ -30,11 +30,11 @@ use axum::Router;
 
 use crate::env::metrics_http_port;
 
-pub struct MetricsServerOptions {
+pub struct ServerOptions {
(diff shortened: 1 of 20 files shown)


nucliadb_node/src/http_server/mod.rs
```diff
@@ -33,13 +34,13 @@ pub struct MetricsServerOptions {
     pub default_http_port: u16,
 }
 
-pub async fn run_http_metrics_server(options: MetricsServerOptions) {
+pub async fn run_http_server(options: MetricsServerOptions) {
```
jotare, line 37: NIT: you could also rename `MetricsServerOptions` to `ServerOptions` for example, as the function is now more generic

jotare (approved):

PR #1280, opened 2023-08-30
/chat ignores empty chunks at the end

jotare:
### Description
`/chat` was returning the status code because last chunk was empty. So by the time we parsed last empty chunk, the status code had been yielded

### How was this PR tested?
New tests

### DIFF
--- a/nucliadb/nucliadb/search/search/chat/query.py
+++ b/nucliadb/nucliadb/search/search/chat/query.py
@@ -236,15 +236,19 @@ async def chat(
     )
 
 
-async def async_gen_lookahead(gen):
-    """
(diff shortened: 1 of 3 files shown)


vangheem (approved):

nucliadb/nucliadb/search/search/chat/query.py
```diff
@@ -247,6 +248,9 @@ async def async_gen_lookahead(gen):
             buffered_chunk = chunk
             continue
 
+        if chunk == b"":
```
vangheem, line 251:
```suggestion
        if len(chunk) == 0:
```

might be better?

PR #1281, opened 2023-08-30
NUCLIADB_ENV unnecessary since env always loaded anyways

vangheem:
### Description

### How was this PR tested?


hermeGarcia (approved):
Would it be a good idea to put something like:
```Python
assert os.environ.get("NUCLIADB_ENV")
```
in purge and setup. to express through the code that we assume that env is always loaded?

jotare (approved):

PR #1294, opened 2023-09-04
Enable processing in on prem

bloodbare:
### Description

### How was this PR tested?


nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -364,6 +365,32 @@ def _check_response(self, response: httpx.Response):
         response_type=ResourceFieldAdded,
     )
 
+    # Configuration
+    set_configuration = _request_builder(
+        name="set_configuration",
```
tarekziade, line 370: nit `create_configuration` (like other `POST` calls)

nucliadb/nucliadb/ingest/processing.py
```diff
@@ -227,6 +196,20 @@ async def initialize(self):
...
+            return None
+
+        ingest = get_ingest()
+
+        kb_obj = KnowledgeBoxID()
+        kb_obj.uuid = kbid
+        pb_response: GetConfigurationResponse = await ingest.GetConfiguration(kb_obj)
+        if pb_response.status.status == OpStatusWriter.Status.OK:
+            return KBConfiguration.from_message(pb_response.config)
+        return None
```
tarekziade, line 211: I would do it the other way, to avoid having a `return None` as the last line since that's the default:

```
if pb_response.status.status!= OpStatusWriter.Status.OK:
    return None
return KBConfiguration.from_message(pb_response.config)
```

nucliadb/nucliadb/ingest/processing.py
```diff
@@ -440,3 +424,58 @@ async def send_to_process(
         return ProcessingInfo(
             seqid=seqid, account_seq=account_seq, queue=QueueType(queue_type)
         )
+
+
+class DummyProcessingEngine(ProcessingEngine):
+    def __init__(self):
+        self.calls: List[List[Any]] = []  # type: ignore
+        self.values: Dict[str, List[Any]] = {}
```
tarekziade, line 432: use a defaultdict(list) here so you don't have to worry about initializing all of them (unless you require a keyerror on unkown keys but it does not look like it)

nucliadb/nucliadb/search/predict.py
```diff
@@ -229,9 +258,24 @@ def get_predict_url(self, endpoint: str) -> str:
...
 
-    def get_predict_headers(self, kbid: str) -> dict[str, str]:
+    async def get_predict_headers(self, kbid: str) -> dict[str, str]:
         if self.onprem:
-            return {"X-STF-NUAKEY": f"Bearer {self.nuclia_service_account}"}
+            headers = {"X-STF-NUAKEY": f"Bearer {self.nuclia_service_account}"}
+            config = await self.get_configuration(kbid)
+            if self.local_predict_headers is not None:
+                headers.update(self.local_predict_headers)
+            if config is not None:
```
tarekziade, line 267: they all follow the same naming, you can do a loop with something like: (untested)

```
SFT_HEADERS = ("ANONYMIZATION-MODEL", "SEMANTIC-MODEL", ...)

for header in SFT_HEADERS:
    value = getattr(config, header.lower().replace('-', '_'), None)
    if value:
        headers[f"X-SFT-{header}"]  = value
```

tarekziade (approved):
LGTM, just nits

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
bloodbare: This function does not create , just sets the config so the name is better like labels.

nucliadb/nucliadb/search/predict.py
bloodbare: We will do in another iteration

nucliadb_models/nucliadb_models/configuration.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class KBConfiguration(BaseModel):
+    # Do not touch this model synced on Processing side
+    semantic_model: Optional[str] = None
+    generative_model: Optional[str] = None
+    ner_model: Optional[str] = None
+    anonymization_model: Optional[str] = None
+    visual_labeling: Optional[str] = None
+
```
lferran, line 38: Maybe not as part of this PR, but @bloodbare do you think it makes sense to add the following:
- Similarity function
- Vector dimension
- Default min score
?
And unify the `KBConfiguration` with https://github.com/nuclia/nucliadb/blob/de4ca9f09942f96aad03e29463aab36f049492ee/nucliadb_models/nucliadb_models/vectors.py#L55
?

PR #1303, opened 2023-09-06
Parse status code from rephrase predict endpoint

lferran:
### Description
Rephrase endpoint is also sending the final LLM status code on the content.

### How was this PR tested?
Unit tests


nucliadb/nucliadb/search/predict.py
```diff
@@ -392,3 +400,24 @@ async def _iter_answer_chunks(gen):
...
+async def _parse_rephrase_response(
+    resp: aiohttp.ClientResponse,
+) -> str:
+    """
+    Predict api is returning a json payload that is a string with the following format:
+    <rephrased_query><status_code>
+    where status_code is "0" for success, "-1" for error and "-2" for no context
+    it will raise an exception if the status code is not 0
+    """
+    content = await resp.json()
```
tarekziade, line 414: nit if it's a json payload containing just a string with no mapping or list structure, you can probably use the body content directly

tarekziade (approved):

PR #1312, opened 2023-09-08
do not use rayon thread for search requests, fix deadlock

vangheem:
### Description

### How was this PR tested?


lferran (commented):
should we fix suggest operations aswell?

lferran (approved):

PR #1329, opened 2023-09-12
Export / Import logic

lferran:
### Description
This pr adds a new nucliadb package with all export and import logic encapsulated.
A subsequent PR will integrate it in the reader and writer components to have the feature available via HTTP.

### Implementation notes
I copied over most of the existing import/export logic from the nucliadb_client and ingest grpc service.
I tried to follow the proposal about not using global context utils

### How was this PR tested?
Integration test


nucliadb/nucliadb/export_import/models.py
```diff
@@ -0,0 +1,27 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+from enum import Enum
+
+
+class CODEX(str, Enum):
```
jotare, line 23: What's the meaning of `codex` in this context? Isn't a codex an ancestor of a book?

nucliadb/nucliadb/export_import/datamanager.py
```diff
@@ -0,0 +1,229 @@
...
+BinaryStream = AsyncGenerator[bytes, None]
+BinaryStreamGenerator = Callable[[int], BinaryStream]
+
+
+class KBExporterDataManager:
+    def __init__(self, driver: Driver, storage: Storage):
+        self.driver = driver
+        self.storage = storage
+
+    async def iter_broker_messages(self, kbid: str) -> AsyncIterator[BrokerMessage]:
```
vangheem, line 69: Do we need a generic broker message datamanager?

nucliadb/nucliadb/export_import/datamanager.py
```diff
@@ -0,0 +1,229 @@
...
+        async with self.driver.transaction() as txn:
+            kb = KnowledgeBoxORM(txn, self.storage, kbid)
+            async for resource in kb.iterate_resources():
+                resource.disable_vectors = False
+                bm = await resource.generate_broker_message()
+                yield bm
+
+    def get_binaries(self, bm: BrokerMessage) -> list[CloudFile]:
```
vangheem, line 77: I was thinking data managers should focus on isolating the interaction with the database. Maybe this function doesn't fit?

nucliadb/nucliadb/export_import/exporter.py
```diff
@@ -0,0 +1,77 @@
...
+async def iter_file_chunks(filename: str) -> AsyncIterator[bytes]:
+    async with aiofiles.open(filename, mode="rb") as f:
+        while True:
+            chunk = await f.read(BINARY_CHUNK_SIZE)
+            if not chunk:
+                break
+            yield chunk
+
+
+async def export_kb(context: KBExporterContext, kbid: str) -> AsyncIterator[bytes]:
```
vangheem, line 42: Should we have some documentation on the export format?

nucliadb/nucliadb/export_import/exporter.py
```diff
@@ -0,0 +1,77 @@
...
+    with tempfile.TemporaryDirectory() as dest_dir:
+        async for bm in context.data_manager.iter_broker_messages(kbid):
+            # Stream the binary files of the broker message
+            binary_aux_file = f"{dest_dir}/{uuid4().hex}"
+            for cloud_file in context.data_manager.get_binaries(bm):
+                async with aiofiles.open(binary_aux_file, "wb") as binary_file:
+                    file_size = await context.data_manager.download_binary(
```
vangheem, line 49: why not iterate on the results of the file download and immediately stream out?

Is it because we can't get the filesize ahead of time?

nucliadb/nucliadb/export_import/exporter.py
```diff
@@ -0,0 +1,77 @@
...
+                        cloud_file, binary_file
+                    )
+                serialized_cf = cloud_file.SerializeToString()
+                yield ExportedItemType.BINARY.encode("utf-8")
+                yield len(serialized_cf).to_bytes(4, byteorder="big")
+                yield serialized_cf
+                yield file_size.to_bytes(4, byteorder="big")
+                async for file_chunk in iter_file_chunks(binary_aux_file):
+                    yield file_chunk
```
vangheem, line 58: After we read, should we delete the file just in case it's a large export and this folder fills up?

vangheem (commented):
I wonder if the boundaries in the data manager need to be understood a bit differently?

At least in my mind, there shouldn't be any datamanager in the export_import folder since we do not store any data related to export/import; however, we might need more generic datamanagers in common/datamanagers to facilitate the needs of export/import

wdyt?

looks sweet!

nucliadb/nucliadb/export_import/exporter.py
lferran: Yes, the main reason was to know the filesize ahead of time.

However, I just realised that the metadata in the cloud file includes the binary size, so there is no need to do it.

vangheem (approved):

PR #1330, opened 2023-09-12
Migrate node metrics to use new proc macro

jotare:
### Description
- Use new `nucliadb_procs::measure` macro for the current metrics

### How was this PR tested?

### DIFF
--- a/nucliadb_vectors/src/service/reader.rs
+++ b/nucliadb_vectors/src/service/reader.rs
@@ -20,8 +20,6 @@
 use std::fmt::Debug;
 use std::time::SystemTime;
 
-use nucliadb_core::metrics;
-use nucliadb_core::metrics::request_time;
 use nucliadb_core::prelude::*;
 use nucliadb_core::protos::prost::Message;
 use nucliadb_core::protos::{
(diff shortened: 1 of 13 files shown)


tarekziade (approved):
very nice improvement!

nucliadb_vectors/src/service/reader.rs
```diff
@@ -65,44 +64,25 @@ impl Debug for VectorReaderService {
...
-            std::mem::drop(lock);
-
-            let metrics = metrics::get_metrics();
-            let took = time.elapsed().map(|i| i.as_secs_f64()).unwrap_or(f64::NAN);
-            let metric = request_time::RequestTimeKey::vectors("count".to_string());
-            metrics.record_request_time(metric, took);
-            debug!("Ending at {took} ms");
-
-            Ok(no_nodes)
+            self.index_count(&index)
```
lferran, line 73: Shouldn't this be a `Ok(self.index_count(&index)` ?

lferran (approved):
nice cleanup indeed! 👏🏼

nucliadb_vectors/src/service/reader.rs
jotare: Index count already returns a Result, so no problem. Anyway, Rust will complain if it wasn't right

PR #1331, opened 2023-09-12
Support synchronous chat response

vangheem:
### Description
Allow synchronous chat responses.

This includes a refactor to make supporting this more tenable

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/chat.py
+++ b/nucliadb/nucliadb/search/api/v1/chat.py
@@ -17,28 +17,43 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
-from typing import Union
+import base64
+from typing import Optional, Union
 
-from fastapi import Body, Header, Request, Response
+import pydantic
+from fastapi import Body, Header, Response
 from fastapi_versioning import version
 from starlette.responses import StreamingResponse
 
 from nucliadb.models.responses import HTTPClientError
 from nucliadb.search import predict
-from nucliadb.search.api.v1.find import find
 from nucliadb.search.api.v1.router import KB_PREFIX, api
-from nucliadb.search.search.chat.query import chat, rephrase_query_from_chat_history
+from nucliadb.search.search.chat.query import (
+    FoundStatusCode,
+    chat,
+    get_relations_results,
+)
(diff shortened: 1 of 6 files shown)


jotare (approved):

nucliadb/nucliadb/search/tests/unit/search/search/chat/test_query.py
```diff
@@ -42,28 +36,6 @@ def predict():
         yield predict
 
 
-async def test_chat_does_not_call_predict_if_no_find_results(
```
lferran, line 45: This feature is no longer wanted?
AFAIR if we send a question to the LLM without context it is more likely to do hallucinations.

nucliadb/nucliadb/search/api/v1/chat.py
```diff
@@ -95,60 +107,68 @@ async def chat_knowledgebox_endpoint(
...
+            yield END_OF_STREAM.encode()
+            if ChatOptions.RELATIONS in chat_request.features:
+                relations_results = await get_relations_results(
+                    kbid=kbid, chat_request=chat_request, text_answer=text_answer
+                )
+                yield base64.b64encode(relations_results.json().encode())
+
+        return StreamingResponse(
+            _streaming_response(),
+            media_type="plain/text",
```
lferran, line 169: the media type is actually `application/octet-stream`, isn't it?

nucliadb/nucliadb/search/api/v1/chat.py
```diff
@@ -63,16 +75,16 @@
...
 async def chat_knowledgebox_endpoint(
     request: Request,
-    response: Response,
     kbid: str,
     item: ChatRequest = Body(examples=CHAT_EXAMPLES),
     x_ndb_client: NucliaDBClientType = Header(NucliaDBClientType.API),
     x_nucliadb_user: str = Header(""),
     x_forwarded_for: str = Header(""),
-) -> Union[StreamingResponse, HTTPClientError]:
+    x_synchronous: bool = Header(False),
```
lferran, line 83: Can you add a description to the header? then we have it in the docs automatically (at least in the openapi spec)

nucliadb/nucliadb/search/api/v1/chat.py
vangheem: I think I just copy/pasted but I'll update nontheless

PR #1332, opened 2023-09-12
fixing the error message for utils

danilopsss:
### Description
When trying to delete a kb that does not exists was generating an error informing that the URL was missing

### How was this PR tested?
Through make test

NOTE: There's some inconsistencies while running the command `make lint`. The linter complain about other folder/files, mainly about the import order.  The errors for utils has been fixed.


nucliadb_sdk/nucliadb_sdk/utils.py
```diff
@@ -123,8 +124,10 @@ def get_or_create(
 
 def delete_kb(slug: str, nucliadb_base_url: Optional[str] = "http://localhost:8080"):
     kb = get_kb(slug, nucliadb_base_url)
-    if kb is None or kb.client.url is None:
+    if kb.client.url is None:
         raise AttributeError("URL should not be none")
+    if not kb:
+        raise NotFoundError("KB does not exists.")
```
lferran, line 130: I think the order of the checks should be the opposite. Otherwise if kb is None, we would crash.

nucliadb_sdk/nucliadb_sdk/tests/test_crud.py
```diff
@@ -98,5 +98,8 @@ def test_crud_resource(kb: KnowledgeBoxObj, sdk: nucliadb_sdk.NucliaDB):
 
     sdk.delete_resource(kbid=kb.uuid, rid=resource.id)
 
+    with pytest.raises(nucliadb_sdk.exceptions.NotFoundError):
+        sdk.delete_resource(kbid=kb.uuid, rid=resource.id)
```
lferran, line 102: This is not testing your changes. You will need to create a test in `test_kb.py` for the `get_kb` function

lferran (commented):
Thanks for your contribution @danilopsss!
Just a few comments

PR #1335, opened 2023-09-13
Integrate Export / Import into HTTP api

lferran:
### Description
Integrates export and import logic into our reader and writer components so that the feature is available via HTTP api

### How was this PR tested?
Integration and unit tests


nucliadb/nucliadb/writer/run.py
```diff
@@ -18,8 +18,8 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 
+from nucliadb.reader.app import get_application
```
vangheem, line 21: you're calling get_application from reader in the writer component!

nucliadb/nucliadb/reader/app.py
```diff
@@ -81,25 +83,29 @@ async def client_disconnect_handler(request: Request, exc: ClientDisconnect):
 )
 
 
-base_app = FastAPI(title="NucliaDB Reader API", **fastapi_settings)  # type: ignore
+def get_application() -> FastAPI:
```
vangheem, line 86: I wonder is "create_application" would be a better word actually?

lferran: yup. or application factory?

nucliadb/nucliadb/writer/run.py
lferran: haha no wonder why tests were not working

jotare (approved):

PR #1336, opened 2023-09-13
nats msg progress updater

vangheem:
### Description
Adds context manager for nats messages to let nats know we're still working on it.

This should allow us to set much lower ack_wait times in our messages since we'll be updating them with progress.

### How was this PR tested?

### DIFF
--- a/nucliadb_utils/nucliadb_utils/nats.py
+++ b/nucliadb_utils/nucliadb_utils/nats.py
@@ -47,6 +47,37 @@ def get_traced_jetstream(nc: NATSClient, service_name: str) -> JetStreamContext:
     return jetstream
 
 
+class message_progress_updater:
+    _task: asyncio.Task
+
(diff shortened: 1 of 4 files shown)


vangheem:
[sc-7166]

lferran (approved):

nucliadb_utils/nucliadb_utils/nats.py
```diff
@@ -47,6 +47,44 @@ def get_traced_jetstream(nc: NATSClient, service_name: str) -> JetStreamContext:
     return jetstream
 
 
+class message_progress_updater:
```
tarekziade, line 50: PEP8 (no reason to hide the fact that this is a class even if it's a context manager)

tarekziade (commented):
nit

jotare (approved):

PR #1345, opened 2023-09-18
NucliaDB Query Grammar proposal

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/proposals/records/006-nucliadb-query-grammar.md
+++ b/proposals/records/006-nucliadb-query-grammar.md
@@ -1,0 +1,81 @@
+# NucliaDB Query Grammar
+
+Based on a recent story, we need to change the query parsing behavior for
+`/catalog` endpoint to allow prefix matching in the last word. `/catalog`
+internally calls `Search` gRPC call to the node (same as `/search` endpoint).
+
+As `/catalog` and `/search` use the same underlaying gRPC call, both need to be
+able to choose a query parsing strategy.
+
+## Current solution/situation
+
+[Detailed description of problem to be solved]
+
+Currently, we are not able to choose query parsing strategies with the current
+`SearchRequest` protobuffer.
+
+In addition, while working on this story, I've found we are parsing
+`SearchRequest` `query` field directly with tantivy's `QueryParser` (in both
+`nucliadb_paragraphs` and `nucliadb_texts`). That means our API users can make
+use of it (although not intended). We use to have an `advanced_query` field to
+expose tantivy's query grammar. However, as we are building the unified index,
+we decided to remove this field to avoid users depending on it and forcing us to
+support tantivy's query grammar.
+
+As the unified index will need to support `/catalog` and `/search` semantics,
+we'll need some way to use different query parsing strategies. And it'll be
+interesting to not expose tantivy's query grammar anymore.
+
+## Proposed Solution
+
+I have 3 proposed solutions for this problem:
+
+1. Modify `SearchRequest` protobuffer so we can know which query parsing we want
+   and ignore tantivy query grammar exposition.
+2. Implement a custom query parser on the node and add some field in
+   `SearchRequest` protobuffer to choose it.
+3. Implement a Query AST using protobuffers, moving the query parsing
+   responsability to the Python side
+
+### Solution 1: protobuf modification
+
+Create a new field and/or deprecate the `document` field, so we can choose if we
+want to search using `nucliadb_texts` or not and with which query parser.
+Instead of a bool, we could create an enum with multiple options like:
+`NO_DOC_SEARCH`, `TERMS_QUERY`, `LAST_WORD_PREFIX`...
+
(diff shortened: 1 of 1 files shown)


proposals/records/006-nucliadb-query-grammar.md
```diff
@@ -0,0 +1,81 @@
...
+`NO_DOC_SEARCH`, `TERMS_QUERY`, `LAST_WORD_PREFIX`...
+
+The advantage is simplicity, we only need to add a new field and implement a
+simple conditional logic.
+
+However, we'll still be parsing tantivy query grammar in the `query` field.
+
+### Solution 2: Rust query parser
+
+Implementing a query parser in Rust replacing tantivy's one will solve the
```
tarekziade, line 54: would that be implementing the same DSL or something new?

jotare: We could implement something smaller. Exact terms, phrase ("exact text quoted") and prefix

proposals/records/006-nucliadb-query-grammar.md
```diff
@@ -0,0 +1,81 @@
...
+To explain this, a simple example: the node receives this query "Ramon progra*".
+It's the user using our query grammar or it's search component asking for a
+prefix search? We'll need an extra field (as Solution 1) to know it.
+
+### Solution 3: Implement query AST using protobuffers
+
+Implementing the query AST in protobuffers would have a simple mapping in the
+Rust side from the protobuf grammar to tantivy or unified index one. However,
+we'll need to move query parsing logic to Python side.
+
```
hermeGarcia, line 68: I think we should implement a query language.
Using protobuffers for defining the AST may be a problem, either the protobuffer types will be spread all over our codebase, making language modifications a nightmare. Or sooner than later we will end up implementing a lowering pass and a new IR just to get rid of the protobuf (which is basically implementing a new parser). We also need and ergonomic way of defining the AST if we dont want to end up suffering every time we define a pass. I dont think Protobuf can give us that.
In option 2 what I dont like is that it misses the point of defining a language. If we are defining our own query language having different parsers based on some client logic seems like is doomed for code repetition and, as always, time consuming and error prone language modifications. I think it would be better to implement this feature inside the language, maybe through something like tags (but we need to give it a though). I would avoid as much as possible implementing different parsers for the same language inside the project, project for which we have defined the language.

jotare: I don't think protobuf is that bad defining the AST. We could implement kind of the same tantivy does with types (BooleanQuery, TermQuery, PraseQuery...). Totally agree if we are not careful we'll end up with even more protobuffer infection though.

Could you explain more this tags idea?

The problem of defining the language and only pass this to the node is the inability to know where it comes from (users or python). If we want to do something special in the Python side (as with /catalog  and /seach), we'll need to parse the user query to build the Python one. So Pyhon will need to know about the query language (either directly or with a binding)

hermeGarcia: Protobuffer infection is a problem, but there are other reasons why that is a bad road to take. Even though Protobuf allows defining recursive types, we can not control how they are represented which is a crucial part when dealing with AST's. We will be forced to use however the library decides to implement this (Most likely a nested `Box`), which may be far from the best fit at some point in time.  Another problem with using protobuf is how difficult defining utils like the visitor trait is going to be, the actual rust code of the AST will only be available after compilation.
If want we want is to make really easy to modify the query based on some business logic there are better approaches, in my opinion our life would be much easier if:
1. We implement a parser from the textual representation of the query language to json and viceversa, we can do this in Rust and provide bindings. This way we have an representation of the language that is easy to modify and in a format that most languages understand.
1. The unified index has its own internal representation of the query language (probably optimised for the kind of passes it needs to do) and knows how to translate to it.
The Python side then just needs to translate the query to its json representation, do some changes if it needs to do so and send it to the unified index.

jotare: Wow, sounds really cool and powerful. But maybe is an overkill for the requested story (that was basically provide a prefix match on certain `nucliadb_texts` searches).

Do you think we should go directly for your implementation idea or maybe do a simpler approach. If we are going to develop this, we'll need to prioritize first

hermeGarcia: Sorry I thought we already had a solution for the feature that is needed right now  and this was just for discussing a query language implementation for the near future.
For prefix match in `nucliadb_texts` lets go for the quick and fast solution, the query language is more of a unified index topic IMO. Right now why dont we implement the following:
As I can see on the [Tantivy documentation](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html) we can add prefix match to frase queries. If you need an endpoint that needs to run a prefix query for a given word _A_,
just send the query _"A"*_  to the node. That will do right?

PR #1346, opened 2023-09-18
Run nucliadb tests on all maindb drivers

lferran:
### Description

The purpose of this PR is to be able to run nucliadb tests by using any supported maindb drivers.
Right now, most of our tests use the redis driver while we use tikv in production.

The idea is that when running tests, one can specify which maindb drivers to test with by using the following env var: 

`TESTING_MAINDB_DRIVERS="tikv,local" pytest nucliadb/nucliadb/...`

Available maindb drivers are: `local`, `redis`, `tikv` and `pg`, and by default it is set to `redis`.

This PR also changes nucliadb CI and nucliadb_ingest CI so that tests use tikv and pg drivers (in parallel jobs)

### DIFF
--- a/nucliadb/nucliadb/common/maindb/pg.py
+++ b/nucliadb/nucliadb/common/maindb/pg.py
@@ -37,6 +37,7 @@ CREATE TABLE IF NOT EXISTS resources (
 class DataLayer:
     def __init__(self, connection: asyncpg.Connection):
         self.connection = connection
+        self.lock = asyncio.Lock()
 
     async def get(self, key: str) -> Optional[bytes]:
         async with self.lock:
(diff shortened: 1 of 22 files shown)


nucliadb/nucliadb/common/maindb/pg.py
```diff
@@ -37,34 +37,39 @@
 class DataLayer:
     def __init__(self, connection: asyncpg.Connection):
         self.connection = connection
+        self.lock = asyncio.Lock()
```
jotare, line 40: Why do we need a lock?

vangheem: Because a single connection can not have concurrent queries run against it

vangheem (approved):

PR #1349, opened 2023-09-18
Improve list of stop words

tarekziade:
### Description

The stop words implementation combines all languages and run against every query, so stop words from one language can impact queries from another.

An example is the word "stop" that is listed in our French stop words, and this English query:

```
what stop this is?
```

ends up being:

```
is?
```

and returns bad results.

This PR does two changes:

- uses NTLK's stop words, which are less aggressive
- uses lingua's language detection so it picks the right language for the stop words list

The language detection is in the critical path and the test ensures that it always takes less than a few ms to run.

### How was this PR tested?

Everything is unit tested.


nucliadb_paragraphs/src/search_query.rs
```diff
@@ -245,7 +250,9 @@ fn preprocess_raw_query(query: &str, tc: &mut TermCollector) -> ProcessedQuery {
     for (qstart, qend) in quote_starts.into_iter().zip(quote_ends.into_iter()) {
         let quote = query[(qstart + 1)..qend].trim();
         let unquote = query[start..qstart].trim();
+        println!("Before stop words: {:?}", unquote);
```
vangheem, line 253: guessing we won't want these here?

nucliadb_paragraphs/src/stop_words.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        stop_words
+    };
+}
+
+pub fn detect_language(query: &str) -> Language {
+    let detected_language: Option<Language> = LANGUAGE_DETECTOR.detect_language_of(query);
+
+    let res = detected_language.unwrap_or(English);
+    println!("Detected language for {:?} is {:?}", query, res);
```
vangheem, line 105: here too

vangheem (commented):
Looks nice. I guess my only concern is adding this language detection in this spot of the pipeline.

nucliadb_paragraphs/src/stop_words.rs
tarekziade: yeah that was for debugging, the final patch won't have them ;)

tarekziade:
> Looks nice. I guess my only concern is adding this language detection in this spot of the pipeline.

yes I had the same concern, and told Carmen we would first bench. and the overhead is <1ms so it's negligible

hermeGarcia (approved):
LGTM

PR #1352, opened 2023-09-19
Build test node Docker images in debug mode

jotare:
### Description
- The idea here is that building a release docker image takes more than actually running the tests in debug (non optimized).

### How was this PR tested?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,16 +98,19 @@ base-node-image:
 	docker buildx build --platform=linux/amd64 -t eu.gcr.io/stashify-218417/basenode:latest . -f Dockerfile.basenode
 	docker push eu.gcr.io/stashify-218417/basenode:latest
 
-build-search-images: build-local-node build-local-sidecar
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -98,16 +98,19 @@ base-node-image:
 	docker buildx build --platform=linux/amd64 -t eu.gcr.io/stashify-218417/basenode:latest . -f Dockerfile.basenode
 	docker push eu.gcr.io/stashify-218417/basenode:latest
 
-build-search-images: build-local-node build-local-sidecar
+build-search-images: build-local-node build-sidecar
```
lferran, line 101: where is the target `build-local-node`?

jotare: Oh, it's not there! Good catch! Do you think it was the current `build-node`?

jotare: I just removed the `build-search-images` target

vangheem (approved):

PR #1362, opened 2023-09-22
[draft] Export import v2

lferran:
### Description
This is a draft of the async implementation of export and import logic

### How was this PR tested?


nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+            subscription_lost_cb=self.setup_nats_subscription,
+            config=nats.js.api.ConsumerConfig(
+                deliver_policy=nats.js.api.DeliverPolicy.BY_START_SEQUENCE,
+                opt_start_seq=1,
+                ack_policy=nats.js.api.AckPolicy.EXPLICIT,
+                max_ack_pending=nats_consumer_settings.nats_max_ack_pending,
+                max_deliver=nats_consumer_settings.nats_max_deliver,
+                ack_wait=nats_consumer_settings.nats_ack_wait,
+                idle_heartbeat=nats_consumer_settings.nats_idle_heartbeat,
+            ),
```
lferran, line 74: I am not sure yet what configuration for the export/import consumers to use.
Any ideas?

nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+KB_IMPORTS = "imports/{import_id}"
+
+
+class BaseConsumer:
+    stream: const.Streams
+
+    def __init__(
+        self,
+        context: ApplicationContext,
+        partition: int,
```
lferran, line 45: Unsure yet what to do about partitioning on export/import streams.

nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+            metadata.status = Status.ERRORED
+            await self.set_metadata(metadata)
+            return
+
+        if metadata.status == Status.FAILED:
+            logger.info(f"Export {export_id} for kbid {kbid} failed, retrying")
+            metadata.status = Status.RUNNING
+        else:  # metadata.status == Status.SCHEDULED:
+            logger.info(f"Starting export {export_id} for kbid {kbid}")
+            metadata.status = Status.RUNNING
```
lferran, line 134: about all this task-related logic, I thought about creating a `async_tasks` top level component in nucliadb that would offer an api to register a function to a particular stream. thoughts?

nucliadb_utils/nucliadb_utils/const.py
```diff
@@ -51,6 +51,24 @@ class INDEX:
...
+        group = "nucliadb-exports-{partition}"
+
+    class IMPORTS:
+        """
+        Importing kbs
+        """
+
+        name = "nucliadb"
+        subject = "ndb.consumer.imports.{partition}"
+        group = "nucliadb-imports-{partition}"
```
lferran, line 70: is it OK to reuse the same "nucliadb" stream, but with different subjects?

nucliadb/nucliadb/export_import/datamanager.py
```diff
@@ -0,0 +1,75 @@
...
+        await self._set(key, data)
+
+    async def _get(self, key: str) -> Optional[bytes]:
+        async with self.driver.transaction() as txn:
+            return await txn.get(key)
+
+    async def _set(self, key: str, data: bytes):
+        async with self.driver.transaction() as txn:
+            await txn.set(key, data)
+            await txn.commit()
```
lferran, line 75: I think this could be refactored as part of a `async_tasks` nucliadb module. Kind of a framework for running async jobs. There will be an api that can be used to check the state of the task or to cancel it.

nucliadb/nucliadb/export_import/consumers.py
lferran: something like:

```python
export_consumer = TaskConsumer.consume_stream(
    stream=const.Stream.EXPORT,
    func=export_kb,
    message_type=ExportMessage,
    partitions=[1],
) 

```

nucliadb/nucliadb/export_import/consumers.py
vangheem: Do not do any partitioning. We don't need to worry about ordering so no partitioning necessary

nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+            queue=group,
+            stream=stream,
+            flow_control=True,
+            cb=self.subscription_worker,
+            subscription_lost_cb=self.setup_nats_subscription,
+            config=nats.js.api.ConsumerConfig(
+                deliver_policy=nats.js.api.DeliverPolicy.BY_START_SEQUENCE,
+                opt_start_seq=1,
+                ack_policy=nats.js.api.AckPolicy.EXPLICIT,
+                max_ack_pending=nats_consumer_settings.nats_max_ack_pending,
```
vangheem, line 70: Just remove this setting wiring. Use the default for a consumer.

nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+                max_deliver=nats_consumer_settings.nats_max_deliver,
```
vangheem, line 71: this too

nucliadb/nucliadb/export_import/consumers.py
```diff
@@ -0,0 +1,270 @@
...
+        await self.context.nats_manager.subscribe(
+            subject=subject,
+            queue=group,
+            stream=stream,
+            flow_control=True,
+            cb=self.subscription_worker,
+            subscription_lost_cb=self.setup_nats_subscription,
+            config=nats.js.api.ConsumerConfig(
+                deliver_policy=nats.js.api.DeliverPolicy.BY_START_SEQUENCE,
+                opt_start_seq=1,
```
vangheem, line 68: I think you need to remove this and allow nats to control the position you are at

nucliadb/nucliadb/export_import/consumers.py
vangheem: I like it. But remove the partitioning as that is something unusual and not a normal pattern.

I would then like to move some other things to it that I did with pub/sub that I screwed up and don't think should actually be pubsub anymore(because of how concurrency works)

nucliadb/nucliadb/export_import/datamanager.py
vangheem: I agree. I think it makes sense.

nucliadb_utils/nucliadb_utils/const.py
vangheem: No, let's create a new stream please!

I think we are already abusing streams a bit.

PR #1366, opened 2023-09-22
Semver is back

vangheem:
### Description

### How was this PR tested?


.github/workflows/_component_prechecks.yml
```diff
@@ -0,0 +1,46 @@
+name: nucliadb Client (py)
```
lferran, line 1: change to `${{ inputs.component }} pre-checks`

.github/workflows/nucliadb_protos.yml
```diff
@@ -0,0 +1,28 @@
+name: NucliaDB Client
```
lferran, line 1: NucliaDB Protos

.github/workflows/pypi_release.yml
```diff
@@ -81,11 +121,11 @@ jobs:
...
-      - name: Upload to pypi
-        run: twine upload dist/*
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+      # - name: Upload to pypi
+      #   run: twine upload dist/*
+      #   env:
+      #     TWINE_USERNAME: __token__
+      #     TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
```
lferran, line 128: does this need to be uncommented? or is t he idea to push to pypi manually ?

nucliadb/VERSION
```diff
@@ -0,0 +1 @@
+../VERSION
```
jotare, line 1: Why do we need a top level VERSION file? If we had Rust crates more organized, we'd have a folder per package with it's own version file inside

.github/workflows/_component_prechecks.yml
vangheem: Oops!

.github/workflows/pypi_release.yml
vangheem: Thanks for noticing! This was from me testing out the entire build process

nucliadb/VERSION
vangheem: We use this VERION as the main version for all of nucliadb.

Everything else is a dependency

It seemed to me that it should stay at the top level.

lferran (approved):

PR #1367, opened 2023-09-25
Tasks api

lferran:
### Description
This PR adds a new top level module named `nucliadb.tasks` which exposes an api to create background tasks via a nats stream.

See [these integration tests](https://github.com/nuclia/nucliadb/pull/1367/files#diff-db1774f0f34c15fec9779af6374e26ac4b9fd6448a6ef1f1a8f2b0c8b011d7faR50-R102) for a simple example on how it can be used.

The first use-case will be exports and imports (upcoming PR soon), for instance:

Creating a nats consumer export kbs can be done as:
```python
from nucliadb import tasks
from pydantic import BaseModel

class ExportMessage(BaseModel):
    kbid: str

class ExportStream(Stream):
    name = "exports"
    subject = "exports"

@tasks.register(name="exports", stream=ExportStream, msg_type=ExportMessage)
async def export_kb(context: ApplicationContext, msg: ExportMessage):
    # export logic here
    ...
```
Then, to start a consumer:
```python
from nucliadb import tasks

context = ApplicationContext()
await context.initialize()
await tasks.start_consumer(name="exports", context)
```
And producing some messages:
```python
from nucliadb import tasks

context = ApplicationContext()
await context.initialize()
producer = await tasks.get_producer(name="exports", context)

task_id1 = await producer(ExportMessage(kbid="kbid1"))
task_id2 = await producer(ExportMessage(kbid="kbid2"))
```
That just pushed two messages into the exports nats stream to be consumed.

### How was this PR tested?
Unit and integration tests


vangheem:
Just some tweaks to the API.

1. Can we just call it "tasks" instead of "async_tasks" -- I'm worried about the term "async" when we're also using "asyncio" and don't want people to be confused that it is some sort of asyncio task library
2. Should the producer API take a Stream parameter?

lferran:
> 2. Should the producer API take a Stream parameter?

It does take a stream parameter on the producer constructor.
Also, the way I think of it is that producer can only produce to a consumer callback with the same parameters.

That's why I did this:
https://github.com/nuclia/nucliadb/pull/1367/files#diff-f35d6a658a45765496c37aaffe17bc05ab684c73806616ae33570628418dce02R65

WDYT?

vangheem (approved):

PR #1369, opened 2023-09-25
Shard-centric merge strategy

hermeGarcia:
### Description
This PR moves the node to our modified version of Tantivy and to a unified merging process.
Now `nucliadb_node` controls when a merge happens and to which segments in `nucliadb_vectors`, `nucliadb_paragraphs` and `nucliadb_texts`. Indexes that in this PR delegate concurrency and parallelism control, so the node.
This is the first step into giving `nucliadb_node` full control  of the resources being used.

### How was this PR tested?
Local tests


nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -447,8 +476,31 @@ impl ShardWriter {
...
     }
 
+    #[tracing::instrument(skip_all)]
+    pub fn merge(&self) -> NodeResult<()> {
+        let current = self.writes_since_merge.swap(0, Ordering::SeqCst);
+
+        if current < MERGE_THRESHOLD {
+            self.writes_since_merge.fetch_add(current, Ordering::SeqCst);
+            return Ok(());
+        }
```
jotare, line 486: If swapping with a 0 and then adding current, don't you get the same value?

nucliadb_paragraphs/src/writer.rs
```diff
@@ -41,6 +41,8 @@ lazy_static::lazy_static! {
     static ref REGEX: Regex = Regex::new(r"\\[a-zA-Z1-9]").unwrap();
 }
 
+const MERGE_THRESHOLD: usize = 5;
```
jotare, line 44: Should be this configurable by settings?

jotare (commented):
Nice work!

nucliadb_node/src/shards/shard_writer.rs
hermeGarcia: Only if there were not enough writes to be merge worthy, I add instead of store because I dont want to assume concurrency mechanisms in this code. Therefore `writes_since_merge` may have change

jotare: Yes, but what I mean is the following. Imagine no concurrent merges:
- You enter and `writes_since_merge` is 3
- After the swap is 0 and `current` is 3
- If `current < MERGE_THRESHOLD` you add `current` to 0, so you have 3 again and return

You have not incremented the counter right?

hermeGarcia: > If current < MERGE_THRESHOLD you add current to 0, so you have 3 again and return

Why would I increment the counter? I did not perform a write on the shard, therefore I leave the state of `writes_since_merge` as I found it (since I did not perform a merge).
Also remember that I dont assume concurrency on top of `ShardWriter` so when I add the value of `writes_since_merge` may not be zero.

nucliadb_core/src/paragraphs.rs
```diff
@@ -33,9 +33,7 @@ pub struct ParagraphConfig {
 pub struct ParagraphIterator(Box<dyn Iterator<Item = ParagraphItem> + Send>);
 impl ParagraphIterator {
     pub fn new<I>(inner: I) -> ParagraphIterator
-    where
-        I: Iterator<Item = ParagraphItem> + Send + 'static,
-    {
+    where I: Iterator<Item = ParagraphItem> + Send + 'static {
```
tarekziade, line 36: is this the old formatting rule?

nucliadb_node/src/grpc/grpc_writer.rs
```diff
@@ -224,20 +240,24 @@ impl NodeWriter for NodeWriterGRPCDriver {
...
+        let status = tokio::task::spawn_blocking(write_task)
+            .await
+            .map_err(|error| {
+                tonic::Status::internal(format!("Blocking task panicked: {error:?}"))
+            })?;
         match status {
             Ok(mut status) => {
                 status.status = 0;
                 status.detail = "Success!".to_string();
+                tokio::task::spawn_blocking(merge_task);
```
tarekziade, line 260: I am confused, are you returning two statuses here?

nucliadb_node/src/shards/providers/unbounded_cache/async_unbounded_writer.rs
```diff
@@ -67,6 +66,7 @@ impl AsyncShardWriterProvider for AsyncUnboundedShardWriterCache {
     }
 
     async fn load(&self, id: ShardId) -> NodeResult<()> {
+        let shard_key = id.clone();
```
tarekziade, line 69: yay, much clearer

nucliadb_protos/rust/src/nodereader.rs
```diff
@@ -212,7 +212,7 @@ pub struct ParagraphSearchResponse {
     pub fuzzy_distance: i32,
     #[prost(int32, tag="1")]
     pub total: i32,
-    /// 
+    ///
```
tarekziade, line 215: empty doc string?

tarekziade (approved):
a few questions, o/wise it looks good to me

nucliadb_protos/rust/src/nodereader.rs
hermeGarcia: this is produced by protos, I dont think I can remove it

nucliadb_node/src/grpc/grpc_writer.rs
hermeGarcia: no! one status and I spawn a task that checks if I can merge

nucliadb_core/src/paragraphs.rs
hermeGarcia: I dont know why this changed happened now and not before

nucliadb_node/src/shards/shard_writer.rs
jotare: Oh sorry, I though your intention was to increment there

nucliadb_vectors/src/data_point_provider/state.rs
```diff
@@ -88,13 +90,22 @@ impl Fssc {
...
+    fn new(size: usize, with_duplicates: bool) -> Fssc {
         Fssc {
             size,
+            with_duplicates,
+            seen: HashSet::new(),
             buff: HashMap::with_capacity(size),
         }
     }
     fn add(&mut self, candidate: Neighbour) {
+        if !self.with_duplicates && self.seen.contains(candidate.vector()) {
```
tarekziade, line 102: nice!

tarekziade (approved):

nucliadb_sdk/nucliadb_sdk/tests/test_search.py
```diff
@@ -292,6 +292,7 @@ def test_search_resource(kb: KnowledgeBoxObj, sdk: nucliadb_sdk.NucliaDB):
         min_score=0.70,
         page_number=0,
         page_size=20,
+        with_duplicates=True,
```
lferran, line 295: I don't get why this was passing before, but now we need this set to `True`. Have you fixed a bug on the node?

hermeGarcia: yes, there was a bug that was returning duplicates in some cases.

PR #1378, opened 2023-09-26
initial rust rust build server integration

vangheem:
### Description

This PR integrates rust build server into our CI.

Right now, the only integration points are places we build the node to be used in our tests.

Once this is stable, we can also build our deploy images with this.

### How was this PR tested?
manual and all in CI


jotare (approved):
Looks good to me! It'll be nice to have a speedup in our buids!

PR #1381, opened 2023-09-27
Extend the release channel to KBs

tarekziade:
### Description

- Add release channel support to KBs
- Make sure all tests using KBs run both channels (STABLE/EXPERIMENTAL)
- Improved the `maybe_create_new_shard` API

### How was this PR tested?


.github/workflows/nucliadb.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       matrix:
         python-version: ["3.11"]
         maindb_driver: ["tikv", "pg"]
+        release_channel: ["EXPERIMENTAL", "STABLE"]
```
lferran, line 36: To not have an explosion of jobs in CI, maybe we can limit the release channel tests to run always with maindb_driver=tikv?

.github/workflows/nucliadb.yml
```diff
@@ -78,6 +79,7 @@ jobs:
         uses: nick-fields/retry@v2
         env:
           TESTING_MAINDB_DRIVERS: ${{ matrix.maindb_driver }}
+          RELEASE_CHANNEL: ${{ matrix.release_channel }}
```
lferran, line 82: Can you make it explicit that is for testing? 
like: `TESTING_NUCLIADB_CHANNEL` or something alike

nucliadb_protos/utils.proto
```diff
@@ -109,4 +109,9 @@ message UserVectorsList {
 enum VectorSimilarity {
     COSINE = 0;
     DOT = 1;
-}
\ No newline at end of file
+}
+
+enum ReleaseChannel {
```
lferran, line 114: Wdyt about making it explicit that this is for the shard/node?
`NodeReleaseChannel`? or `ShardReleaseChannel`?

.github/workflows/nucliadb.yml
tarekziade: do you want to have these tested?

- tikv+EXPERIMENTAL
- tikv+STABLE
- pg+STABLE

or just:

- tikv+EXPERIMENTAL
- pg+STABLE

vangheem: I guess 

- tikv+EXPERIMENTAL
- tikv+STABLE
- pg+STABLE

is better so we don't have to wonder which is the culprit for failures

but it also shouldn't be difficult to separate tikv from index node issues.

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -213,11 +215,17 @@ async def create(
             logger.error(f"{uuid} KB could not be created")
             failed = True
 
+        if release_channel is None:
+            release_channel = ReleaseChannel.STABLE
```
lferran, line 219: nit: is there any reason why not setting the default value in the argument of the function definition?

nucliadb/nucliadb/tests/benchmarks/test_search.py
```diff
@@ -69,6 +70,7 @@ async def test_search_returns_labels(
     warmup=False,
 )
 @pytest.mark.asyncio
+@pytest.mark.parametrize("knowledgebox", ["STABLE", "EXPERIMENTAL"], indirect=True)
```
lferran, line 73: oh, nice way to parametrize a fixture! I had not been able to understand this from reading the pytest docs

lferran (approved):

PR #1385, opened 2023-09-28
Export/Import v2 api

lferran:
### Description
This PR only adds the new HTTP api for exports/imports to unblock integrating with nuclia.py.
Eventually, exports and imports will happen in background tasks, therefore the need for the status endpoints.

### How was this PR tested?
Unit and integration tests.


nucliadb/nucliadb/export_import/exporter.py
```diff
@@ -39,35 +42,72 @@ async def export_kb(
...
-
-    # Stream the entities and labels of the knowledgebox
+    async for chunk in export_resources(context, kbid):
+        yield chunk
+
+    async for chunk in export_entities(context, kbid):
+        yield chunk
+
+    async for chunk in export_labels(context, kbid):
+        yield chunk
```
jotare, line 52: I like this refactor, it looks much cleaner now!

nucliadb/nucliadb/writer/api/v1/export_import.py
```diff
@@ -17,29 +17,55 @@
...
+    status_code=200,
+    name="Start an export of a Knowledge Box",
+    tags=["Knowledge Boxes"],
+)
+@requires_one([NucliaDBRoles.MANAGER, NucliaDBRoles.WRITER])
+@version(1)
+async def start_kb_export_endpoint(request: Request, kbid: str) -> CreateExportResponse:
+    export_id = uuid4().hex
+    response = CreateExportResponse(export_id=export_id)
+    return response
```
jotare, line 45: Is kind of confusing for me having an `/export` enpoint in both reader and writer to start and download an export. The purpose of the writer's `export` is to start an asynchronous export?

lferran: Yes! I know... we could just have an export/import HTTP service right? but for now we decided to include in the current reader and writer.

And yes, the purpose of the writer's export is to start the job (PR coming up..)

jotare (approved):

PR #1388, opened 2023-09-28
[sdk] Be able to properly stream downloads to a file

lferran:
### Description
Download export requests return an (async) iterator of bytes from the streamed response that can be used to store the bytes into a file without having to accumulate all the response content in memory

### How was this PR tested?
unit and local tests


vangheem (approved):
nice, this is fun

lferran:
7174

PR #1392, opened 2023-09-29
Stdout logs improvements

tarekziade:
### Description

Introduces :

- JSON output by default (use `DEBUG=true` or the new `plain_logs` settings
- glob-style patterns for our log targets. e.g. you can now use `nucliadb_*`

### How was this PR tested?

Unit tests

### DIFF
--- a/nucliadb_node/src/telemetry.rs
+++ b/nucliadb_node/src/telemetry.rs
@@ -19,6 +19,7 @@
 
 use std::sync::Arc;
 
+use nucliadb_core::tracing::*;
 use nucliadb_core::tracing::{Level, Span};
 use nucliadb_core::{Context, NodeResult};
 use opentelemetry::global;
@@ -26,6 +27,7 @@ use opentelemetry::trace::TraceContextExt;
 use sentry::ClientInitGuard;
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use tracing_subscriber::filter::{FilterFn, LevelFilter, Targets};
+use tracing_subscriber::fmt::format::Format;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{Layer, Registry};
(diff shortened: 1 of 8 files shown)


nucliadb_node/src/telemetry.rs
```diff
@@ -65,13 +66,46 @@ pub fn init_telemetry(settings: &Arc<Settings>) -> NodeResult<Option<ClientInitG
...
+                }
+                // starts with
+                if target.len() > 1 && target.ends_with('*') {
+                    target.truncate(target.len() - 1);
+                    if metadata_target.starts_with(&target) {
+                        println!("true");
+                        return true;
+                    }
+                }
+            }
```
hermeGarcia, line 93: What about moving some of this logic to `Settings`, that way we can do some previous computations that will render this filter lighter

tarekziade:
@hermeGarcia one thing that frustrates me before we land the patch is that I was not able to extract the `filter` function builder out of the `stdout_layer` function so I can add some unit tests to verify the filtering code -- if you know how. Returning a `FilterFn` instance from an isolated function seems to be problematic in Rust because of how closures work.

nucliadb_node/src/telemetry.rs
```diff
@@ -64,14 +68,100 @@ pub fn init_telemetry(settings: &Arc<Settings>) -> NodeResult<Option<ClientInitG
...
+
+        // match all
+        if self.all_target_level != LevelFilter::OFF && self.all_target_level >= *level {
+            return true;
+        }
 
+        let metadata_target = metadata.target();
+        // exact match
+        if let Some(log_level) = self.logs_map.get(metadata_target) {
+            if log_level >= level {
```
hermeGarcia, line 111:
```rust
if self.logs_map.get(metadata_target) >= Some(level) {
    return true;
}
```

nucliadb_node/src/telemetry.rs
```diff
@@ -64,14 +68,100 @@ pub fn init_telemetry(settings: &Arc<Settings>) -> NodeResult<Option<ClientInitG
...
+                return true;
+            }
+        }
+        // prefixes match
+        for (prefix, log_level) in self.prefixes.iter() {
+            if log_level >= level && metadata_target.starts_with(prefix) {
+                return true;
+            }
+        }
+        false
```
hermeGarcia, line 121:
```rust 
        let possible_match = self
            .prefixes
            .iter()
            .find(|(prefix, log_level)| log_level >= level && metadata_target.starts_with(prefix));
        possible_match.is_some()
```

hermeGarcia (approved):
LGTM just a couple of comments

tarekziade:
Example of output:

```
✗ RUST_LOG="nuclia*=DEBUG" ../target/release/node_reader 2>/dev/null

{"timestamp":"2023-10-02T10:03:38.472169Z","level":"INFO","fields":{"message":"telemetry to https://telemetry.nuclia.cloud/ is enabled."},"target":"nucliadb_telemetry::sender"}
```

PR #1395, opened 2023-09-29
Add deprecation warnings to nucliadb_client

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_client/nucliadb_client/client.py
+++ b/nucliadb_client/nucliadb_client/client.py
@@ -34,6 +34,7 @@ from opentelemetry.propagators.b3 import B3MultiFormat
 from nucliadb_client import logger
 from nucliadb_client.exceptions import ConflictError
 from nucliadb_client.knowledgebox import KnowledgeBox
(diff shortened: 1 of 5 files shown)


vangheem (approved):

nucliadb_client/setup.py
```diff
@@ -42,11 +42,5 @@ def load_reqs(filename):
...
     include_package_data=True,
     packages=find_packages(),
-    install_requires=requirements,
-    entry_points={
-        "console_scripts": [
-            "nucliadb_export = nucliadb_client.export:run",
-            "nucliadb_import = nucliadb_client.importing:run",
-        ]
-    },
+    install_requires=["nucliadb_sdk"],
```
lferran, line 45: @vangheem what do you think about this?
installing from pip will effectively install nucliadb_sdk.
Maybe it's too much?

vangheem: Yes, too much. I don't see the point here either. No one is using this except us internally.

PR #1402, opened 2023-10-02
hnsw ops doc

hermeGarcia:
### Description
First PR of the work that will be done over time in order to bring better documentation to our hnsw implementation. 

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_vectors/src/data_point/ops_hnsw.rs
+++ b/nucliadb_vectors/src/data_point/ops_hnsw.rs
@@ -27,41 +27,38 @@ use rand::prelude::*;
 use rand::rngs::SmallRng;
 
 use super::*;
+use crate::data_point::params;
 
-pub mod params {
-    pub fn level_factor() -> f64 {
-        1.0 / (m() as f64).ln()
-    }
-    pub const fn m_max() -> usize {
-        30
-    }
-    pub const fn m() -> usize {
-        30
-    }
-    pub const fn ef_construction() -> usize {
-        100
-    }
-}
+/// Implementors of this trait can guide the hnsw search
 pub trait DataRetriever: std::marker::Sync {
-    fn get_key(&self, _: Address) -> &[u8];
-    fn is_deleted(&self, _: Address) -> bool;
-    fn has_label(&self, _: Address, _: &[u8]) -> bool;
-    fn similarity(&self, _: Address, _: Address) -> f32;
-    fn get_vector(&self, _: Address) -> &[u8];
+    /// Returns the key of the embedding pointed by x.
+    fn get_key(&self, x: Address) -> &[u8];
+    /// Checks if x is a deleted embedding.
+    fn is_deleted(&self, x: Address) -> bool;
+    /// During filtering this function will be called to know if x has the necessary labels.
+    fn has_label(&self, x: Address, label: &[u8]) -> bool;
+    /// Computes the similarity between the embeddings pointed by x and y.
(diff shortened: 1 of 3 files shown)


nucliadb_vectors/src/data_point/params.rs
```diff
@@ -0,0 +1,35 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+//
+
+//! Set of parameters needed by the HNSW algorithm
+//! as named and used in the paper.
+
+pub fn level_factor() -> f64 {
```
tarekziade, line 24: maybe we can explain each param in one sentence here?

tarekziade (approved):

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -27,41 +27,38 @@ use rand::prelude::*;
...
+    /// Returns the key of the embedding pointed by x.
+    fn get_key(&self, x: Address) -> &[u8];
+    /// Checks if x is a deleted embedding.
+    fn is_deleted(&self, x: Address) -> bool;
+    /// During filtering this function will be called to know if x has the necessary labels.
+    fn has_label(&self, x: Address, label: &[u8]) -> bool;
+    /// Computes the similarity between the embeddings pointed by x and y.
+    fn similarity(&self, x: Address, y: Address) -> f32;
+    /// Returns the serialized embedding of x, is meant to be used by [`crate::data_types::vector`].
+    fn get_vector(&self, x: Address) -> &[u8];
```
jotare, line 43: I don't think functions with good names as, for example, `is_deleted(...) -> bool` need more explanation. The only difference between reading the function signature and the documentation is that you're calling an `Address` and embedding. If that's a crucial part, consider renaming the type.

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -27,41 +27,38 @@ use rand::prelude::*;
...
 }
 
+/// Implementors of this trait are an HNSW where search can be ran.
 pub trait Hnsw {
     type L: Layer;
     fn get_entry_point(&self) -> Option<EntryPoint>;
     fn get_layer(&self, i: usize) -> Self::L;
 }
 
+/// Util for storing embedding pointers and their similarity in a [`BinaryHeap`].
```
jotare, line 61: If `Cnx` is a data structure to store ordered embedding pointers by similarity, it's name doesn't give enough information

jotare (commented):
Nice! I love your docs effort!

Just a couple of comments

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: An `Address` is a pointer to an embedding, will document that part of the code in the next round.

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: `Cnx` does not store ordered embeddings, its basically a new-type for the tuple `(Address, f32)`. I had to do this because `f32` does not implement `Ord` and therefore storing the tuple directly in a `BinaryHeap` is not possible.
I fixed the documentation to express this better

PR #1412, opened 2023-10-03
Nucliadb metrics refinement

jotare:
### Description
- Remove unneeded shard metrics prefix
- Add `grpc_server_op_duration_seconds` metric in the node (same metric used in other gRPC servers in NucliaDB)
- Measure new metric in middleware

### Comments
- Does it make sense to keep metrics on shards? We can check how much different they are and decide

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/metrics/meters/mod.rs
+++ b/nucliadb_core/src/metrics/meters/mod.rs
@@ -23,16 +23,14 @@ mod prometheus;
 pub use noop::NoOpMeter;
 pub use prometheus::PrometheusMeter;
 
-use crate::metrics::metric::request_time;
+use crate::metrics::metric::grpc_ops::{GrpcOpKey, GrpcOpValue};
+use crate::metrics::metric::request_time::{RequestTimeKey, RequestTimeValue};
 use crate::metrics::task_monitor::{Monitor, TaskId};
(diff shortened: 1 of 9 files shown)


nucliadb_core/src/metrics/meters/mod.rs
```diff
@@ -23,16 +23,14 @@ mod prometheus;
...
 use crate::NodeResult;
 
 pub trait Meter: Send + Sync {
-    fn record_request_time(
-        &self,
-        metric: request_time::RequestTimeKey,
-        value: request_time::RequestTimeValue,
-    );
+    fn record_request_time(&self, metric: RequestTimeKey, value: RequestTimeValue);
+    fn record_grpc_op(&self, method: GrpcOpKey, value: GrpcOpValue);
```
hermeGarcia, line 33: Could we document a bit this trait methods to explain their difference better?

hermeGarcia (approved):
LGTM! Just one comment

nucliadb_core/src/metrics/meters/mod.rs
jotare: I'll do it in the next PR (I'll tweak a bit our buckets)

PR #1416, opened 2023-10-03
Fix bindings build

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,12 +125,12 @@ debug-run-nucliadb-redis:
 
 build-node-binding:
 	rm -rf target/wheels/*
-	RUSTFLAGS="--cfg prometheus_metrics --cfg tokio_unstable" maturin build -m nucliadb_node_binding/Cargo.toml --profile release-wheel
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -125,12 +125,12 @@ debug-run-nucliadb-redis:
 
 build-node-binding:
 	rm -rf target/wheels/*
-	maturin build -m nucliadb_node_binding/Cargo.toml --profile release-wheel
+	RUSTFLAGS="--cfg tokio_unstable" maturin build -m nucliadb_node_binding/Cargo.toml --profile release-wheel
```
jotare, line 128: I think you should add `--cfg prometheus_metrics` too. @hermeGarcia?

tarekziade: I am also curious why we can't have all the options in https://github.com/nuclia/nucliadb/blob/main/.cargo/config.toml

vangheem: `--cfg prometheus_metrics` gives us errors with the bindings

Now, I also understand why the build server was failing. This is annoying.

jotare (approved):

PR #1417, opened 2023-10-04
Fix import resumable logic

lferran:
### How was this PR tested?
Integration and unit tests


nucliadb/nucliadb/export_import/utils.py
```diff
@@ -237,14 +237,14 @@ class ExportStream:
...
-        self.length = len(export.getvalue())
+        self._length = len(export.getvalue())
 
     async def read(self, n_bytes):
         """
         Reads n_bytes from the export stream.
         Raises ExportStreamExhausted if there are no more bytes to read.
         """
-        if self.read_bytes >= self.length:
+        if self.read_bytes >= self._length:
```
tarekziade, line 247: can you read *over* ? shouldn't this be `==` ?

tarekziade (approved):

nucliadb/nucliadb/export_import/utils.py
lferran: yup! you're right. changing

PR #1419, opened 2023-10-04
Add support for json str objects on search results

bloodbare:
### Description

### How was this PR tested?


tarekziade (commented):
improvements : If this is something we want for all objects, I would create a base class on the top of pydantic `BaseModel` to add it (and maybe tweak `__repr__`), and maybe catch any error on the `json()` call to fallback to a error representation (or you might end up with some weird behaviors)

tarekziade (approved):

PR #1421, opened 2023-10-04
Add PrometheusMetricObserver abstraction

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/metrics/metric/request_time.rs
+++ b/nucliadb_core/src/metrics/metric/request_time.rs
@@ -18,9 +18,18 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue};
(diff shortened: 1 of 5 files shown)


nucliadb_core/src/metrics/metric/request_time.rs
```diff
@@ -31,14 +40,15 @@ enum RequestActor {
...
-    actor: RequestActor,
-    request: String,
-}
-
 pub type RequestTimeValue = f64;
 
+type RequestCount = Family<RequestTimeKey, Counter>;
+type RequestDuration = Family<RequestTimeKey, Histogram>;
+
+const BUCKETS: [f64; 12] = [
```
tarekziade, line 48: fly-by - maybe we can add a comment to explain why we have those values for the histogram

jotare: I'm not sure about this buckets. I suppose they're a bit generic in the range of values expected. @hermeGarcia?

PR #1423, opened 2023-10-04
A few more export/import polishments

lferran:
### Description
- cleaning up unused status
- tearing down nats first (as subscribers may depend on storage, index and maindb driver utils)
- put tasks under feature flag

### How was this PR tested?
existing tests

### DIFF
--- a/nucliadb/nucliadb/common/context/__init__.py
+++ b/nucliadb/nucliadb/common/context/__init__.py
@@ -84,11 +84,11 @@ class ApplicationContext:
         if not self._initialized:
             return
 
+        await stop_nats_manager()
         await teardown_driver()
         await teardown_cluster()
         await self.blob_storage.finalize()
         await stop_indexing_utility()
-        await stop_nats_manager()
         await stop_transaction_utility()
         stop_partitioning_utility()
         clean_utility(Utility.STORAGE)
--- a/nucliadb/nucliadb/common/context/__init__.py
+++ b/nucliadb/nucliadb/common/context/__init__.py
@@ -72,24 +72,25 @@ class ApplicationContext:
         self.shard_manager = await setup_cluster()
         self.partitioning = start_partitioning_utility()
         if not in_standalone_mode():
-            self.indexing = await start_indexing_utility()
             self.nats_manager = await start_nats_manager(
                 self.service_name,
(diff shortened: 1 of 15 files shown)


nucliadb/nucliadb/export_import/importer.py
```diff
@@ -66,6 +66,8 @@ async def import_kb(
         if item_type == ExportedItemType.RESOURCE:
             bm = cast(writer_pb2.BrokerMessage, data)
             await import_broker_message(context, kbid, bm)
+            if metadata:
```
tarekziade, line 69: `if metadata is not None:`

nucliadb/nucliadb/export_import/importer.py
```diff
@@ -85,10 +87,14 @@ async def import_kb(
             continue
 
         if metadata and count % 10 == 0:
-            # Update metadata every 10 items
+            # Save checkpoint in metadata every 10 items
             metadata.read_bytes = stream_reader.read_bytes
             await dm.set_metadata("import", metadata)
 
+    if metadata:
```
tarekziade, line 94: if metadata is not None

nucliadb/nucliadb/tasks/consumer.py
```diff
@@ -120,7 +120,7 @@ async def subscription_worker(self, msg: Msg):
                     },
                 )
                 # Nak the message to retry
-                await asyncio.sleep(1)
+                await asyncio.sleep(2)
```
tarekziade, line 123: what's the logic behind this change? in any case, could be a CONST

tarekziade (approved):
looks good just a few nits

nucliadb/nucliadb/common/context/__init__.py
```diff
@@ -84,11 +84,11 @@ async def finalize(self) -> None:
         if not self._initialized:
             return
 
+        await stop_nats_manager()
```
tarekziade, line 87: nit: in theory, the tear down sequence should be in the reverse order of the creation

lferran: totally agree! otherwise we can have unexpected exceptions on service teardowns!
Changing

PR #1429, opened 2023-10-06
Fix filtering by entities

lferran:
### Description
Currently, entities are detected at the field level.
These entities are converted into "node labels" or "tags" on that are indexed for later filtering.
However, these are not being propagated down to all entries in the HNSW index.
Therefore, in the semantic search, if you filter by a detected entity, you get no results back.

### How was this PR tested?
Integration/Functional test


lferran:
[sc-7479]

lferran:
closing for now. will take on the efforts after date range prefiltering is implemented

PR #1431, opened 2023-10-06
automatic python version and release handling

vangheem:
### Description

### How was this PR tested?


bump.py
```diff
@@ -33,22 +30,41 @@ def run(args):
...
-        with open("nucliadb_node_binding/Cargo.toml", "w") as f:
-            f.write("\n".join(new_cargo))
+    with open("nucliadb_node_binding/Cargo.toml", "w") as f:
+        f.write("\n".join(new_cargo))
+
+    # go through each requirements.txt and update the version to the new bump
+    for req_filepath in (
+        "nucliadb/requirements.txt",
+        "nucliadb_utils/requirements.txt",
+        "nucliadb_sdk/requirements.txt",
```
lferran, line 53: is nucliadb_telemetry missing?

lferran (approved):

bump.py
vangheem: It doesn't depend on anything else so we don't need to update any of the dependencies!

PR #1438, opened 2023-10-11
Add metrics on search / find features

lferran:
### Description
This can be helpful to know how much each search feature is used.
I just added a few labels to see how filters are used.
We can add more if we need.

### How was this PR tested?
new unit tests and existing tests


jotare (approved):

nucliadb/nucliadb/search/search/query.py
```diff
@@ -320,3 +321,24 @@ async def get_default_min_score(kbid: str) -> float:
...
+
+
+def record_features_counter_metric(pb_query: SearchRequest, counter: Counter):
+    if len(pb_query.filter.tags):
+        counter.inc({"type": "filters"})
+        if any(tag.startswith("/l/") for tag in pb_query.filter.tags):
+            counter.inc({"type": "labels_filters"})
+        if any(tag.startswith("/e/") for tag in pb_query.filter.tags):
+            counter.inc({"type": "entities_filters"})
+    if len(pb_query.key_filters):
```
tarekziade, line 333: small suggestion, maybe the block below can be simplified for readability

```
for field in ("key_filters", "vectorset", ..):
    if getattr(pb_query, field, False):
        counter.inc({"type": field})
```

lferran: mh, actually, if you don't have any big objection with it, with metrics recording code I prefer being more verbose and explicit

lferran: nevermind, I changed the implementation to do the increments in the pb_query function

nucliadb/nucliadb/search/search/query.py
```diff
@@ -89,10 +90,20 @@ async def global_query_to_pb(
     request.min_score = min_score
     request.body = query
     request.with_duplicates = with_duplicates
-    request.filter.tags.extend(filters)
+    if len(filters) > 0:
+        request.filter.tags.extend(filters)
+        node_features.inc({"type": "filters"})
+        if any(tag.startswith("/l/") for tag in filters):
```
tarekziade, line 96: nit (micro improvement -- maybe its overkill if the request has very few tags ) : you are iterating twice on the filter list to detect the two prefixes, and on the second pass you are trying to detect the `/e/` prefix on tags that were potentially detected as being `/l/`.

A single *sorted* loop will be slightly more efficient because the prefixed tag will be at the beginning of the list, so if you iterate and find both prefixes, you can break. And you can also break once the first char is not `/` anymore.

untested:

```
>>> def detect_prefixes(tags):
...     tags.sort()
...     e_prefix = False
...     t_prefix = False
...     for tag in tags:
...         if tag[0] != '/' or (e_prefix and t_prefix):
...             break
...         if tag.startswith('/e/'):
...             e_prefix = True
...             # set the inc
...         elif tag.startwith('/t'):
...             t_prefix = True
...             # set the inc
```

Also: maybe `/l/` and `/e/` can be constants

lferran: yup, agree! It's not that hard to make this metrics code as lightweight as possible, so let's do it!

PR #1446, opened 2023-10-12
delete existing pg tus upload in same transaction

vangheem:
### Description

### How was this PR tested?


tarekziade (approved):
The change is ok - did you get an error in prod with that sub transaction?  if it's the case, bonus point if we can add a test.

I am also thinking that this class has a small design issue -- if those operations can happen in the same transaction, we should separate concerns and have an explicit context manager to deal with it:

```
with storage.transaction() as operations_on_dl:
    operations_on_dl.op1()
    operations_on_dl.op2()
```

PR #1447, opened 2023-10-13
[Import] Publish broker message ref to nats when it's too big

lferran:
### Description
Some broker messages exceed the max payload size on nats streams. When this happens, we will upload the broker message to google cloud storage and just send a reference to it over nats.

### How was this PR tested?
Unit tests

### DIFF
--- a/nucliadb/nucliadb/export_import/utils.py
+++ b/nucliadb/nucliadb/export_import/utils.py
@@ -21,6 +21,7 @@ import functools
 from io import BytesIO
 from typing import AsyncGenerator, AsyncIterator, Callable, Optional
 
+import nats.errors
 from google.protobuf.message import DecodeError as ProtobufDecodeError
 
 from nucliadb.common.context import ApplicationContext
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/export_import/utils.py
```diff
@@ -71,13 +72,39 @@ async def import_broker_message(
...
+    """
+    try:
         await context.transaction.commit(
-            import_bm,
+            bm,
             partition,
             wait=False,
             target_subject=Streams.INGEST_PROCESSED.subject,
         )
+    except nats.errors.MaxPayloadError:
```
tarekziade, line 93: i sthere a way for us to know that the payload is too big *before* we send it? so we would pick the right strategy without having to fail first

lferran: however, one caveat of this is that we'll need to serialize the broker message every time to compute its lenght

lferran: I think it's OK to leave it as it was... It also follows more the “Easier to Ask Forgiveness Than Permission” style

tarekziade (approved):

PR #1456, opened 2023-10-16
refactored the q/a

tarekziade:
### Description

### How was this PR tested?


nucliadb_protos/resources.proto
```diff
@@ -270,6 +262,22 @@ message FieldMetadata {
     repeated Relations relations = 13;
 }
 
+message QuestionAnswer {
+  string question = 1;
+  string question_language = 2;
+  string answer = 3;
+  string answer_language = 4;
+  repeated int64 paragraph_ids = 5;
```
bloodbare, line 270: This should be stirng isn't it ? the paragraphs ids are `$RID/$FIELD/($SPLIT)/$START-$END`

nucliadb_protos/resources.proto
```diff
@@ -270,6 +262,22 @@ message FieldMetadata {
...
+}
+
+message FieldQuestionAnswerWrapper {
+    oneof file_or_data {
+        QuestionAnswer question_answer = 1;
```
lferran, line 275: I think this field should point to another message where multiple questions and answers for the same field, different paragraphs

lferran (approved):

PR #1461, opened 2023-10-17
Make sure to cleanup blob storage after exports/imports

lferran:
### Description
Otherwise we will be piling up stuff in GCS for no reason.

### How was this PR tested?
Unit and integration tests


lferran:
[sc-7563]

nucliadb/nucliadb/export_import/datamanager.py
```diff
@@ -130,6 +134,27 @@ def _get_storage_field(
...
+        bucket = self.storage.get_bucket_name(kbid)
+        await self.storage.delete_upload(key, bucket_name=bucket)
+
+    async def delete_export(self, kbid: str, export_id: str):
+        key = STORAGE_EXPORT_KEY.format(export_id=export_id)
+        bucket = self.storage.get_bucket_name(kbid)
+        await self.storage.delete_upload(key, bucket_name=bucket)
+
+    async def try_delete_from_storage(self, type: str, kbid: str, id: str):
+        assert type in ("export", "import")
```
tarekziade, line 148: nit: it's better to have a real exception in code and keep asserts in tests. They will even be ignored when running Python in certain O modes.

PR #1466, opened 2023-10-18
Tweaks in benchmark script

lferran:
### Description
- Each worker gets assigned the same kbid to perform requests to
- Handle errors to inspect after test run
- Include nucliadb_performance in docker image so we can easily run scripts from within a cluster pod

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_performance/nucliadb_performance/test_search_cluster.py
+++ b/nucliadb_performance/nucliadb_performance/test_search_cluster.py
@@ -1,17 +1,30 @@
 import asyncio
 import random
+from dataclasses import dataclass
 
 from faker import Faker
-from molotov import global_setup, scenario
+from molotov import get_context, global_setup, global_teardown, scenario
 
 from nucliadb_performance.utils import (
+    RequestError,
     get_fake_word,
-    get_random_kb,
     get_search_client,
     load_kbs,
+    pick_kb,
 )
 
 fake = Faker()
+
+
+@dataclass
+class Error:
+    kbid: str
+    endpoint: str
+    status_code: int
+    error: str
+
+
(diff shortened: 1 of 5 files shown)


nucliadb_protos/rust/src/resources.rs
```diff
@@ -414,8 +414,13 @@ pub struct QuestionAnswer {
     pub answer: ::prost::alloc::string::String,
     #[prost(string, tag="4")]
     pub answer_language: ::prost::alloc::string::String,
-    #[prost(int64, repeated, tag="5")]
```
tarekziade, line 417: unrelated change?

lferran: yeah, weird! I didn't think I touched this file. will remove the changes from the branch

tarekziade: if you rebase from `main` and run `make protos` hopefully it should be on par

nucliadb_performance/nucliadb_performance/test_search_cluster.py
```diff
@@ -80,9 +109,37 @@ async def test_find(session):
...
+        detail = (
+            err.content and err.content.get("detail", None) if err.content else err.text
+        )
+        error = Error(
+            kbid=kbid,
+            endpoint=path.split("/")[-1],
+            status_code=err.status,
+            error=detail,
+        )
+        ERRORS.append(error)
```
tarekziade, line 144: That makes me think I could provide functions to print in the console (the right side is the errors) so you can also have them while running

tarekziade (approved):

PR #1469, opened 2023-10-18
Allow field and paragraph classifications without labels

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/train/generators/field_classifier.py
+++ b/nucliadb/nucliadb/train/generators/field_classifier.py
@@ -65,19 +65,21 @@ async def generate_field_classification_payloads(
     node: AbstractIndexNode,
     shard_replica_id: str,
 ) -> AsyncIterator[FieldClassificationBatch]:
-    labelset = f"/l/{trainset.filter.labels[0]}"
-
     # Query how many resources has each label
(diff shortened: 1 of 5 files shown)


nucliadb/nucliadb/train/generators/field_classifier.py
```diff
@@ -65,19 +65,21 @@ async def generate_field_classification_payloads(
...
 ) -> AsyncIterator[FieldClassificationBatch]:
-    labelset = f"/l/{trainset.filter.labels[0]}"
-
     # Query how many resources has each label
     request = StreamRequest()
     request.shard_id.id = shard_replica_id
-    request.filter.tags.append(labelset)
-    total = 0
 
+    labelset = None
```
tarekziade, line 72: At this point should we only have 0 or 1 item in `trainset.filter.labels` ?

jotare: Yes

tarekziade (commented):
looks like we're changing `generate_field_classification_payloads` behavior but I don't see any tests, would it be possible to exercise it?

tarekziade (approved):
Thanks for the test, you rock

PR #1472, opened 2023-10-18
handle span dropped logs better

vangheem:
So we are not spammed

### DIFF
--- a/nucliadb_telemetry/nucliadb_telemetry/batch_span.py
+++ b/nucliadb_telemetry/nucliadb_telemetry/batch_span.py
@@ -110,7 +110,7 @@ class BatchSpanProcessor(SpanProcessor):
             if not self._spans_dropped:
                 logger.warning("Queue is full, likely spans will be dropped.")
                 self._spans_dropped = True
-                return
+            return
 
         try:
             self.queue.put_nowait(span)  # type: ignore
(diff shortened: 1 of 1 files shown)


nucliadb_telemetry/nucliadb_telemetry/batch_span.py
```diff
@@ -110,11 +110,15 @@ def on_end(self, span: ReadableSpan) -> None:
...
+            return
 
         try:
             self.queue.put_nowait(span)  # type: ignore
         except asyncio.QueueFull:
-            logger.warning(f"Queue is full. Queue size : {self.queue.qsize()}")
+            if not self._spans_dropped:
+                logger.warning(f"Queue is full. Queue size : {self.queue.qsize()}")
+                self._spans_dropped = True
+            return
```
lferran, line 121: should there be a `else` branch here where we set `self._spans_dropped = False`?

lferran (approved):

nucliadb_telemetry/nucliadb_telemetry/batch_span.py
vangheem: no no, we want that to be cleared later on when the batch is done

this is to prevent too many logging calls

PR #1473, opened 2023-10-18
read replica k8s config

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/charts/nucliadb_node/templates/node.replicas.deploy.yaml
+++ b/charts/nucliadb_node/templates/node.replicas.deploy.yaml
@@ -77,6 +77,10 @@ spec:
         - configMapRef:
             name: nucliadb-config
         env:
+          - name: NODE_ROLE
+            value: secondary
+          - name: PRIMARY_ADDRESS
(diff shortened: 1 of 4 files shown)


charts/nucliadb_node/templates/node.replicas.deploy.yaml
```diff
@@ -0,0 +1,170 @@
...
+      - name: data-dir
+        emptyDir:
+          sizeLimit: {{ mul $values.readReplicas.disk.numLocalDisks $values.readReplicas.disk.localDiskSize $values.readReplicas.disk.reservationPercentage | int }}Gi
+{{- if $values.nats.secretName }}
+      - name: nats-creds
+        secret:
+          secretName: {{ $values.nats.secretName }}
+{{- end }}
+      containers:
+      - name: writer
```
lferran, line 63: I guess writer is still needed because we have all the health checking in the grpc server?

vangheem: Yes and the writer is what is connecting to the primary to sync the data as well

PR #1477, opened 2023-10-19
Be more explicit on vectors chat feature

lferran:
### Description

### How was this PR tested?
existing tests should cover it


jotare (approved):
Why do we offer VECTORS feature even though we ignore it? Does it make sense to don't do vector search if omitted?

PR #1478, opened 2023-10-19
Store and retrieve Q/A data

tarekziade:
### Description

- Stores Q/A data from `BrokerMessage` into the resource's storage
- Extended the resource `read` API to allow optional retrieval

### How was this PR tested?

unit + functional test


nucliadb/nucliadb/ingest/fields/base.py
lferran, line 1: The deletion operation is missing. Can you implement it too? It must be integrated into the `delete` method.

PR #1482, opened 2023-10-20
better log handling support for standalone

vangheem:
### Description

- add support for file log handling
- default standalone to file logs
- output server config at startup
- default to plain text logs for standalone

Screenshot 2023-10-19 at 22 05 56

### DIFF
--- a/charts/nucliadb/values.yaml
+++ b/charts/nucliadb/values.yaml
@@ -15,7 +15,8 @@ env:
   cluster_discovery_mode: kubernetes
   cluster_discovery_kubernetes_namespace: nucliadb
   cluster_discovery_kubernetes_selector: "app.kubernetes.io/name=node"
-  # when we're k8s, we want structured logs to stdout
+  # when we're k8s, we want structured logs to stdout because we don't have
+  # so they can be scraped by log exporters
   log_output_type: stdout
   log_format_type: structured
 
--- a/charts/nucliadb/values.yaml
+++ b/charts/nucliadb/values.yaml
@@ -15,6 +15,9 @@ env:
   cluster_discovery_mode: kubernetes
   cluster_discovery_kubernetes_namespace: nucliadb
   cluster_discovery_kubernetes_selector: "app.kubernetes.io/name=node"
+  # when we're k8s, we want structured logs to stdout
+  log_output_type: stdout
+  log_format_type: structured
 
 envSecrets:
   # - name: NUA_API_KEY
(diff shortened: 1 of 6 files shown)


charts/nucliadb/values.yaml
```diff
@@ -15,6 +15,9 @@ env:
   cluster_discovery_mode: kubernetes
   cluster_discovery_kubernetes_namespace: nucliadb
   cluster_discovery_kubernetes_selector: "app.kubernetes.io/name=node"
+  # when we're k8s, we want structured logs to stdout
```
lferran, line 18: can you add the reason in the comment? "so that they are directly scraped to gcloud logging system" or something alike

nucliadb/nucliadb/standalone/run.py
```diff
@@ -72,11 +84,40 @@ def run():
...
+        "API": f"http://{settings.http_host}:{settings.http_port}/api",
+        "Contributor/management UI": f"http://{settings.http_host}:{settings.http_port}/contributor",
+        "Key-value backend": ingest_settings.driver,
+        "Blog storage backend": storage_settings.file_backend,
+        "Cluster discovery mode": cluster_settings.cluster_discovery_mode,
+        "Node replicas": cluster_settings.node_replicas,
+        "Index data path": cluster_settings.data_path,
+        "Node port": cluster_settings.standalone_node_port,
+        "Log output type": settings.log_output_type,
+    }
```
lferran, line 100: WDYT about adding some hints on auth type configured too?

nucliadb_telemetry/nucliadb_telemetry/logs.py
```diff
@@ -176,40 +224,47 @@ def setup_logging() -> None:
...
+        _maybe_create_file_directory(settings.error_log)
+        log_handler = RotatingFileHandler(
+            settings.info_log, maxBytes=settings.max_log_file_size, backupCount=5
         )
-    )
+        error_log_handler = RotatingFileHandler(
+            settings.error_log, maxBytes=settings.max_log_file_size, backupCount=5
+        )
+    else:
+        log_handler = error_log_handler = logging.StreamHandler()
```
lferran, line 239: careful! I think `log_handler` and `error_log_handler` will be the same instance here.
```python
>>> class A:
...     ...
... 
>>> a = b = A()
>>> a is b
True
```

vangheem: yup, that's on purpose.

We only end up using the error log handler when we're outputting to an error log file

PR #1485, opened 2023-10-20
Add id to field and paragraph classifications

jotare:
### Description
- Add field and paragraph ids to field and paragraphs classifications respectively

### How was this PR tested?
- Tests on train and datasets components

### DIFF
--- a/nucliadb_protos/dataset.proto
+++ b/nucliadb_protos/dataset.proto
@@ -38,6 +38,7 @@ message Label {
 message TextLabel {
     string text = 1;
     repeated Label labels = 2;
+    string id = 3;
 }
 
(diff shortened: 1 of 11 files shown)


nucliadb_protos/dataset.proto
```diff
@@ -38,6 +38,7 @@ message Label {
 message TextLabel {
     string text = 1;
     repeated Label labels = 2;
+    string id = 3;
```
lferran, line 41: should this be named `field_id` ?

jotare: No, it serves as field_id and paragraph_id. It's used in both classification protobuffers

lferran (approved):

PR #1499, opened 2023-10-26
sdk update_by_slug method

clayrisse:
### Description
Add "update by slug" method to SDK 
Include models  and docstring accordingly 

### How was this PR tested?
It wasn't!

### DIFF
--- a/nucliadb_sdk/nucliadb_sdk/v2/sdk.py
+++ b/nucliadb_sdk/nucliadb_sdk/v2/sdk.py
@@ -73,11 +73,9 @@ from nucliadb_models.vectors import VectorSet, VectorSets
 from nucliadb_models.writer import (
     CreateResourcePayload,
     UpdateResourcePayload,
-    UpdateResourceBySlugPayload,
     ResourceCreated,
     ResourceUpdated,
     ResourceFieldAdded,
-    ResourceBySlugUpdated, 
 
 )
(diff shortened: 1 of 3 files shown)


vangheem (changes requested):
You don't need to define a new payload. The update that is by a slug just uses the same `UpdateResourcePayload` payload.

You can see here that the endpoint is the same in code and what payload definition is used: https://github.com/nuclia/nucliadb/blob/main/nucliadb/nucliadb/writer/api/v1/resource.py#L218

vangheem (approved):

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -371,6 +371,15 @@ def _check_response(self, response: httpx.Response):
         response_type=ResourceUpdated,
         docstring=docstrings.UPDATE_RESOURCE,
     )
+    update_resource_by_slug = _request_builder(
+        name="update_resource_by_slug",
+        path_template="/v1/kb/{kbid}/slug/{slug}",
+        method="PATCH",
+        path_params=("kbid", "slug"),
```
vangheem, line 378: I think you'll need to change it to this unfortunately since we have some duplicate params

```suggestion
        path_template="/v1/kb/{kbid}/slug/{rslug}",
        method="PATCH",
        path_params=("kbid", "rslug"),
```

PR #1527, opened 2023-11-01
Set zone on sentry as tag

sunbit:


charts/nucliadb/values.yaml
```diff
@@ -9,7 +9,7 @@ replicas: 2
 
 # app settings
 env:
-  NUCLIA_ZONE: "europe-1"
+  NUCLIA_ZONE: "provider-region-ordinal"
```
vangheem, line 12: Can we keep the default here?

vangheem (changes requested):

charts/nucliadb/values.yaml
sunbit: done

vangheem (approved):

PR #1528, opened 2023-11-01
add primary_node_id to node metadata and add get_metadata to replication service

vangheem:
### Description

### How was this PR tested?


nucliadb_node/src/replication/replicator.rs
```diff
@@ -261,6 +262,11 @@ pub async fn connect_to_primary_and_replicate(
         let no_shards_to_sync = replication_state.shard_states.is_empty();
         let no_shards_to_remove = replication_state.shards_to_remove.is_empty();
 
+        if !primary_id_set {
+            set_primary_node_id(replication_state.primary_id.clone())?;
+            primary_id_set = true;
+        }
```
hermeGarcia, line 268: Im a bit concern about writing to an env var in the body of a very big hot loop based on some logic (now just in the first iteration). Maybe we could find a way of structuring better this logic, like for example have secondaries know their primary id on creation. I dont know, what do you think?

hermeGarcia (approved):
Looks good! Just some comment

nucliadb_node/src/replication/replicator.rs
vangheem: You are completely right. I was getting impatient and didn't see the better solution.

I've updated the PR.

PR #1530, opened 2023-11-01
better date range tests

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/metrics/metric/tokio_runtime.rs
+++ b/nucliadb_core/src/metrics/metric/tokio_runtime.rs
@@ -38,16 +38,7 @@ pub struct TokioRuntimeObserver {
 
 impl TokioRuntimeObserver {
     pub fn new(registry: &mut Registry) -> Self {
-        let runtime = Handle::try_current();
-        if runtime.is_err() {
-            info!("Cannot export tokio runtime metrics, no runtime available");
-            Self {
-                runtime: None,
-                intervals: None,
-                metrics: TokioRuntimeMetrics::new(registry),
-            }
-        } else {
-            let runtime = runtime.unwrap();
+        if let Ok(runtime) = Handle::try_current() {
             let monitor = RuntimeMonitor::new(&runtime);
             // We need to store RuntimeIntervals iterator instead of RuntimeMonitor to
             // get incremental values. We need a Mutex to be Send and Sync (for our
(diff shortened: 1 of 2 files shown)


nucliadb_core/src/metrics/metric/tokio_runtime.rs
```diff
@@ -19,39 +19,55 @@
...
+            // We need to store RuntimeIntervals iterator instead of RuntimeMonitor to
+            // get incremental values. We need a Mutex to be Send and Sync (for our
+            // Meter users)
+            let intervals = Mutex::new(monitor.intervals());
+            Self {
+                runtime: Some(runtime),
+                intervals: Some(intervals),
+                metrics: TokioRuntimeMetrics::new(registry),
+            }
         }
```
hermeGarcia, line 61:
```rust
let Ok(runtime) =  Handle::try_current() else {
          info!("Cannot export tokio runtime metrics, no runtime available");
           return Self {
                runtime: None,
                intervals: None,
                metrics: TokioRuntimeMetrics::new(registry),
            };
};
let monitor = RuntimeMonitor::new(&runtime);
// We need to store RuntimeIntervals iterator instead of RuntimeMonitor to
// get incremental values. We need a Mutex to be Send and Sync (for our
// Meter users)
let intervals = Mutex::new(monitor.intervals());
Self {
       runtime: Some(runtime),
       intervals: Some(intervals),
       metrics: TokioRuntimeMetrics::new(registry),
}
```

nucliadb_core/src/metrics/metric/tokio_runtime.rs
```diff
@@ -19,39 +19,55 @@
...
     pub fn observe(&self) -> NodeResult<()> {
+        if self.runtime.is_none() {
+            return Ok(());
+        }
+
         let interval = self.next_interval()?;
-        let raw_metrics = self.runtime.metrics();
+        let raw_metrics = self.runtime.as_ref().unwrap().metrics();
 
         self.metrics.update(raw_metrics, interval);
```
hermeGarcia, line 72: I would try to avoid this pattern of checking if something is the error variant or none and then unwrapping  for pattern maching checks, which are less error prone. For example here you could:
```rust
        let Some(runtime) = self.runtime.as_ref() else {
            return Ok(());
        };
        let interval = self.next_interval()?;
        let raw_metrics = runtime.metrics();
        self.metrics.update(raw_metrics, interval);

        Ok(())
```

hermeGarcia (changes requested):

nucliadb_core/src/metrics/metric/tokio_runtime.rs
vangheem: Ah dang, thanks. I actually forgot about these changes. It was quite annoying that I couldn't run these tests locally without the changes.

PR #1531, opened 2023-11-02
Create new training to stream ordered paragraphs

jotare:
### Description
- Add new training to stream ordered paragraphs as if you were reading fields sequentially. This can be used to train Q&A models for example.

### How was this PR tested?
- Integration tests in both train and datasets packages


nucliadb/nucliadb/train/generator.py
```diff
@@ -56,54 +72,50 @@ async def generate_train_data(kbid: str, shard: str, trainset: TrainSet):
...
             kbid, trainset, node, shard_replica_id
-        ):
-            payload = image_data.SerializeToString()
+        )
+
+    elif trainset.type == TaskType.PARAGRAPH_STREAMING:
+        payloads_generator = generate_paragraph_streaming_payloads(
+            kbid, trainset, node, shard_replica_id
+        )
+
```
lferran, line 115: nit: should we handle the `else` branch?

nucliadb/nucliadb/train/tests/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    response: aiohttp.ClientResponse,
+    pb_klass: type[TokenClassificationBatch],
+) -> AsyncIterator[TokenClassificationBatch]:
+    ...
+
+
+async def get_batches_from_train_response_stream(
+    response: aiohttp.ClientResponse,
+    pb_klass: TrainBatchType,
+) -> AsyncIterator[TrainBatch]:
```
lferran, line 106: This in fact should be a `AsyncGenerator[TrainBatch, None]`, and I think the overloads are not needed.

nucliadb/nucliadb/train/tests/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    while True:
+        header = await response.content.read(4)
+        if header == b"":
+            break
+        payload_size = int.from_bytes(header, byteorder="big", signed=False)
+        payload = await response.content.read(payload_size)
+        batch = pb_klass()
+        batch.ParseFromString(payload)
+        # all our batches have a data field
+        assert batch.data
```
lferran, line 116: why is this assert needed?

nucliadb/nucliadb/train/generator.py
jotare: Sure, I can return a 400 saying it's an invalid training

nucliadb/nucliadb/train/generators/paragraph_streaming.py
```diff
@@ -0,0 +1,93 @@
...
+                item.id = f"{rid}/{field_type}/{field}/{split}/{paragraph.start}-{paragraph.end}"
+                item.text = split_text[paragraph.start : paragraph.end]
+
+                batch.data.append(item)
+                if len(batch.data) == trainset.batch_size:
+                    yield batch
+                    batch = ParagraphStreamingBatch()
+
+    if len(batch.data):
+        yield batch
```
lferran, line 93: nit: you could add the batching logic into a separate function. This would ease testing too.

nucliadb/nucliadb/train/tests/utils.py
jotare: I can change it to `AsyncGenerator` but we still need the overloads to avoid mypy complaining about field access that are not in all protobufs. For example, the `ImageClassificationBatch` has a field `selections` that any other have.

nucliadb/nucliadb/train/tests/utils.py
jotare: It was there. I can remove it

lferran (approved):

nucliadb/nucliadb/train/generators/paragraph_streaming.py
jotare: Do you mean a generic "batchify" function that iterates an iterator and return batches? I'll do it for all generators then

lferran: yes, exactly. Something that is able to batch in a generic way

lferran (approved):

PR #1533, opened 2023-11-02
Be able to create a KB in experimental channel in stage

lferran:
### Description
This is to be able to create KBs in experimental mode without the need for internal nucliadb logic wiring into backend / dashboard.

### How was this PR tested?
Unit tests


nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -909,3 +912,19 @@ def parse_model_metadata(request: KnowledgeBoxNew) -> SemanticModelMetadata:
...
+    Set channel to Experimental if specified in the grpc request or if the requested
+    slug has the experimental_kb feature enabled in stage environment.
+    """
+    release_channel = request.release_channel
+
+    stage_environment = running_settings.running_environment == "stage"
+    enabled_for_slug = has_feature(
+        const.Features.EXPERIMENTAL_KB, context={"slug": request.slug}
+    )
+    if stage_environment and enabled_for_slug:
```
vangheem, line 928: Move the has_feature check to the and so it is only conditionally called if we're in stage env
```suggestion
    if stage_environment and enabled_for_slug and has_feature(
        const.Features.EXPERIMENTAL_KB, context={"slug": request.slug}
    ):
```

vangheem (changes requested):
I think you should move the check to be in the conditional so it is only called if we're in stage env

vangheem (approved):

PR #1535, opened 2023-11-02
Integrate cluster read replicas

vangheem:
### Description

### How was this PR tested?


nucliadb_utils/nucliadb_utils/const.py
```diff
@@ -76,3 +76,4 @@ class Features:
     ROLLOVER_SHARDS = "nuclaidb_rollover_shards"
```
lferran, line 76: oops! I just noticed this typo... we should be careful with this flag!

charts/nucliadb_node/templates/node.svc.yaml
lferran, line 1: how come we can delete the sidecar container? How/where is the indexing going to happen? Is this PR depending on other previous changes?

nucliadb/nucliadb/common/cluster/manager.py
```diff
@@ -437,31 +461,32 @@ async def add_resource(
...
+                if read_replica_node is None:
+                    continue
+                potential_nodes.append((replica_id, read_replica_node))
+
+    if len(potential_nodes) == 0:
+        raise NoHealthyNodeAvailable("Could not find a node to query")
 
-    raise KeyError("Could not find a node to query")
+    selected_node: AbstractIndexNode
+    replica_id, selected_node = random.choice(potential_nodes)
```
lferran, line 488: If we go for a random choice approach here, we could be giving inconsistent results during  the time that the changes after a write are being replicated to the read replicas.

By inconsistent I mean that some requests would hit the primary shard and return the "up to date" results and others would hit the replicas and return the unsynced results.

Maybe we could cache the `choose_node` decision for some time?
Just throwing a thought here, I am not sure if this is a big deal...

charts/nucliadb_node/templates/node.svc.yaml
lferran: oh nevermind! I see it now: we're removing the GRPC interface to the sidecar, as we don't need to communicate with it anymore 👍🏼

nucliadb_node/nucliadb_node/settings.py
```diff
@@ -30,7 +30,7 @@ class Settings(BaseSettings):
 
     writer_listen_address: str = "0.0.0.0:10000"
     reader_listen_address: str = "0.0.0.0:10001"
-    sidecar_listen_address: str = "0.0.0.0:10002"
+    sidecar_listen_address: str = "0.0.0.0:10002"  # this component's listen address
```
lferran, line 33: can be removed to?

nucliadb/nucliadb/common/cluster/manager.py
lferran: Another thought:
Since we know that search performance can be impacted by merges / gc, would it make sense to give priority to read only replicas here?

This way we make sure we query on the right segments.

nucliadb_utils/nucliadb_utils/const.py
vangheem: actually, I can clean this up now. It was only temporary to get out to prod

nucliadb/nucliadb/common/cluster/manager.py
vangheem: IMO, for the system we have an how writes are coming in, I think this is a good initial approach.

I agree with every one of your points; however, I think there are other things we can work toward before we add more up front decisions in this component.

For example:
1. we should reduce the time of when a change to an index happens to when it is replicated. Right now, we are really naive in our first approach. But let's measure and focus on bringing this time down long term instead of caching
2. weighting the node selection -- I really agree! But IMO, let's do something very simple to start with
3. finally, remember, we'd like to build an index coordinator. This component implement is somewhat temporary hopefully but maybe I'm overly hopeful

nucliadb_node/nucliadb_node/settings.py
vangheem: We still use this configuration to tell us which host/port to bind to for the sidecar health check

lferran (approved):

PR #1537, opened 2023-11-03
Unify release channel in config and shards obj

lferran:
### Description
Otherwise the values can end up in an inconsistent state

### How was this PR tested?
Integration tests


lferran:
[sc-7724]

jotare (approved):

PR #1538, opened 2023-11-03
Query planner and date ranges

hermeGarcia:
### Description

### How was this PR tested?


nucliadb_node/src/shards/shard_reader.rs
```diff
@@ -397,108 +398,62 @@ impl ShardReader {
...
 
         let mut rtext = None;
         let mut rparagraph = None;
         let mut rvector = None;
         let mut rrelation = None;
 
         crossbeam_thread::scope(|s| {
-            if !skip_fields {
-                s.spawn(|_| rtext = text_task());
+            if let Some(task) = text_task {
```
lferran, line 446: we should keep the skip conditions?

jotare (commented):
I'm not sure you're using the best approach for the query planner. As it is now, it's more a mutable struct of search request protobuffers with some filtering structs that will be modified by the indexes themselves. In addition, the shard reader is still responsible to execute the query plan implicitly.

As I see it:
-The indexes shouldn't be aware of a query planning, they are a lower level of abstraction. The query planner should be the responsible to create the correct query per index.
- `ShardReader` shouldn't be responsible of using the search requests protobuffers from the query planner, because that'd split the query planning semantics in two parts. Either the query planner should return a query plan (something like the Command pattern) or actually do the search.
- It's not useful to create all search request protobuffers when creating the query planner. If the first step doesn't return results, why do we have to compute the others?

nucliadb_node/src/shards/shard_reader.rs
hermeGarcia: That logic is part of the query planner now

lferran (approved):

nucliadb_node/src/shards/shard_reader.rs
lferran: ouch, right

PR #1539, opened 2023-11-03
fix test

hermeGarcia:
### Description
The following test: 
```rust
 fn test_default_env_settings() {
     let settings = EnvSettingsProvider::generate_settings().unwrap();
     assert_eq!(settings.shards_path().to_str().unwrap(), "data/shards")
}
```
Did not take into account that the system running it may already have `DATA_PATH` set, failing in such case.
More problematic, is that this tests change the state of the system running it after being ran, due to changes in env vars that are not reverted. For this reason, a system that already has `DATA_PATH` set will only have a successful run of `test_default_env_settings` if `test_env_settings_data_path` is ran before.
This PR ensures that the env vars are in the expected state before running the test and that after running such tests the state is restored.

### How was this PR tested?
Test past even with the env vars already set.

### DIFF
--- a/nucliadb_node/src/settings/providers.rs
+++ b/nucliadb_node/src/settings/providers.rs
@@ -150,6 +150,8 @@ pub mod env {
         #[test]
         #[serial]
         fn test_default_env_settings() {
+            // The system may have DATA_PATH set
+            std::env::remove_var("DATA_PATH");
             let settings = EnvSettingsProvider::generate_settings().unwrap();
             assert_eq!(settings.shards_path().to_str().unwrap(), "data/shards")
         }
--- a/nucliadb_node/src/settings/providers.rs
+++ b/nucliadb_node/src/settings/providers.rs
@@ -150,28 +150,54 @@ pub mod env {
         #[test]
         #[serial]
         fn test_default_env_settings() {
-            // The system may have DATA_PATH set
+            // Safe current state of DATA_PATH
+            let data_path_copy = std::env::var("DATA_PATH");
+            // Remove DATA_PATH for the test
             std::env::remove_var("DATA_PATH");
+
             let settings = EnvSettingsProvider::generate_settings().unwrap();
+
+            if let Ok(value) = data_path_copy {
+                // The state needs to be restored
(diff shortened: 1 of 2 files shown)


nucliadb_node/src/utils.rs
```diff
@@ -153,11 +153,13 @@ pub async fn list_shards(shards_path: PathBuf) -> Vec<String> {
 
 #[cfg(test)]
 mod tests {
+    use serial_test::serial;
     use tempfile::TempDir;
 
     use super::*;
 
     #[test]
+    #[serial]
```
jotare, line 162: Why does it need to run serially?

nucliadb_node/src/utils.rs
```diff
@@ -167,6 +169,7 @@ mod tests {
     }
 
     #[test]
+    #[serial]
```
jotare, line 172: Same

nucliadb_node/src/utils.rs
```diff
@@ -181,12 +184,14 @@ mod tests {
...
```
jotare, line 187: Same

nucliadb_node/src/settings/providers.rs
```diff
@@ -150,26 +150,54 @@ pub mod env {
         #[test]
         #[serial]
         fn test_default_env_settings() {
+            // Safe current state of DATA_PATH
```
jotare, line 153: nit: you could add a function that does this copy-and-restore work to avoid repetition

nucliadb_node/src/utils.rs
hermeGarcia: Forgot to remove it.

nucliadb_node/src/settings/providers.rs
hermeGarcia: I just did this to have something that would solve the env var mess, I would not want to loose much time in this approach because I dont think is the way to go.
In a separate PR I will try to centralise the way tests get env vars so doing this set-and-restore is not needed.

jotare (approved):

PR #1544, opened 2023-11-03
Cleanup rust settings management

vangheem:
### Description

### How was this PR tested?


nucliadb_node/src/shards/providers/unbounded_cache/async_unbounded_reader.rs
```diff
@@ -38,16 +39,16 @@ pub struct AsyncUnboundedShardReaderCache {
 }
 
 impl AsyncUnboundedShardReaderCache {
-    pub fn new(shards_path: PathBuf) -> Self {
+    pub fn new(settings: Arc<Settings>) -> Self {
```
jotare, line 42: Doubt: why is it better to propagate all `Settings` object instead of passing the required parameters?

jotare (approved):

nucliadb_node/src/shards/providers/unbounded_cache/async_unbounded_reader.rs
vangheem: it's probably not better, I was just going for some consistency in usage

I can change it to passing all the settings if you feel strongly

I think I was thinking that passing settings would prevent needing to continually churn around wiring different params up

but I really don't care either way

jotare: No, it's fine, I was just wondering. I agree it makes the code more consistent although we are passing unneeded information

nucliadb_node/src/http_server/mod.rs
```diff
@@ -24,19 +24,16 @@ mod metrics_service;
...
 
-pub struct ServerOptions {
-    pub default_http_port: u16,
-}
-
-pub async fn run_http_server(options: ServerOptions) {
+pub async fn run_http_server(settings: Arc<Settings>) {
     // Add routes to services
-    let addr = SocketAddr::from(([0, 0, 0, 0], metrics_http_port(options.default_http_port)));
+    let addr = SocketAddr::from(([0, 0, 0, 0], settings.metrics_port()));
```
hermeGarcia, line 36: Much nicer!

nucliadb_node/src/grpc/grpc_writer.rs
```diff
@@ -414,7 +414,7 @@ impl NodeWriter for NodeWriterGRPCDriver {
         &self,
         _request: Request<EmptyQuery>,
     ) -> Result<Response<NodeMetadata>, Status> {
-        let metadata = crate::node_metadata::NodeMetadata::new().await;
+        let metadata = crate::node_metadata::NodeMetadata::new(Arc::clone(&self.settings)).await;
```
hermeGarcia, line 417: I wonder if we should use the inner pattern here to avoid all of the `Arc::clone` through the codebase. What I mean by inner pattern:
Instead of having `Arc<Settings>` in the codebase we define a private type in the `settings` module called  `InnerSettings`. This is basically just the current `Settings`, but without the methods. Then we define the type `Settings` in as: 
```rust
pub struct Settings {
      inner: Arc<Settings>,
}
```
Then we can implement cheap clonning for `Settings` and all the getters we want.

hermeGarcia (changes requested):
Not really a change because , because I understand it may be to big for this PR.

hermeGarcia (approved):

nucliadb_node/src/settings/mod.rs
```diff
@@ -30,291 +30,191 @@
...
+        Settings {
+            inner: Arc::clone(&self.inner),
+        }
+    }
 }
 
 impl Settings {
-    pub fn builder() -> SettingsBuilder {
-        SettingsBuilder::default()
+    pub fn builder() -> InnerSettingsBuilder {
```
jotare, line 68: `InnerSettingsBuilder` is exposed and are public. If you're hiding the implementation details from `Settings`, why expose `InnerSettings`?

PR #1548, opened 2023-11-06
support better filtering naming and origin metadata support

vangheem:
### Description
- support filtering by origin.metadata
- support aliasing our short filter names to provide a better API experience
- some renaming, minor refactoring to improve code readability

### How was this PR tested?


nucliadb/nucliadb/search/search/query.py
```diff
@@ -51,9 +51,47 @@
...
+    Translate friendly filter names to the shortened filter names.
+    """
+    output = []
+    for fltr in filters:
+        if fltr[0] != "/":
+            raise InvalidQueryError(
+                "filters", f"Invalid label filter. It must start with a `/`: {fltr}"
+            )
+        if len(fltr) == 0:
+            raise InvalidQueryError("filters", f"Invalid label filter: {fltr}")
```
lferran, line 85: this should be moved as the first condition to check, otherwise an empty filter `""` will cause an error on line 80

nucliadb/nucliadb/ingest/orm/labels.py
```diff
@@ -28,18 +28,13 @@
     "u": [],  # contributors s (Source) o (Origin)
     "f": [],  # field keyword field (field/keyword)
     "fg": [],  # field keyword (keywords) flat
+    "m": [],  # origin metadata in the form of (key/value). Max key/value size is 255
 }
 
 
-def flat_resource_tags(tags_dict):
+def flatten_resource_labels(tags_dict: dict[str, list[str]]) -> List[str]:
```
lferran, line 35:
```suggestion
def flatten_resource_labels(tags_dict: dict[str, list[str]]) -> list[str]:
```

nucliadb/nucliadb/search/search/query.py
```diff
@@ -51,9 +51,47 @@
...
 
+from .exceptions import InvalidQueryError
+
 ENTITY_FILTER_PREFIX = "/e/"
 LABEL_FILTER_PREFIX = "/l/"
 
+LABEL_QUERY_ALIASES = {
+    # aliases to make querying labels easier
+    "tags": ["t"],
+    "labels": ["l"],
```
lferran, line 62: Thought: should we use `classification-labels` instead? To be more specific.

nucliadb/nucliadb/search/search/query.py
```diff
@@ -51,9 +51,47 @@
...
+    "entities": ["e"],
+    "icon": ["n", "i"],
+    "status": ["n", "s"],
+    "primary-language": ["s", "p"],
+    "secondary-languages": ["s", "s"],
+    "field": ["f"],
+    "field-values": ["fg"],
+    "metadata": ["m"],
```
lferran, line 70: thought: should we use `origin-metadata` instead? to be more specific.

nucliadb/nucliadb/search/search/query.py
```diff
@@ -51,9 +51,47 @@
...
+
+def translate_label_filters(filters: List[str]) -> List[str]:
+    """
+    Translate friendly filter names to the shortened filter names.
+    """
+    output = []
+    for fltr in filters:
+        if fltr[0] != "/":
+            raise InvalidQueryError(
+                "filters", f"Invalid label filter. It must start with a `/`: {fltr}"
```
lferran, line 82: Since you are reusing the `translate_label_filters` for different params (`faceted` and `filters`), you can't hardcode the param value for the exception here.

lferran (commented):
Great API improvement 👏🏼 

Do you think we should translate the values in the responses too?
This way that the filters/faceted API is consistent.

However, this would be a breaking change...so I am fine if we decide not to do the output part now too.

jotare (approved):
I love it! Nice work

nucliadb/nucliadb/search/search/query.py
vangheem: or even just "classification" I guess right?

nucliadb/nucliadb/search/search/query.py
vangheem: of course! my bad

nucliadb/nucliadb/search/search/query.py
vangheem: I was also thinking of something that mapped directly to the API like:

- "origin/tags"
- "classification/labels"
- "metadata/status"
- "metadata/language"
- "metadata/languages"
- "origin/metadata"

PR #1550, opened 2023-11-06
fix s3 delete policy

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_utils/nucliadb_utils/storages/s3.py
+++ b/nucliadb_utils/nucliadb_utils/storages/s3.py
@@ -46,11 +46,19 @@ RETRIABLE_EXCEPTIONS = (
 POLICY_DELETE = {
     "Rules": [
         {
-            "Expiration": {
-                "Days": 0,
-            },
+            "Expiration": {"Days": 1},
+            "ID": "FullDelete",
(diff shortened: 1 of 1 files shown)


nucliadb_utils/nucliadb_utils/storages/s3.py
lferran, line 1: Can you explain what the problem was and in which way have you fixed it?

sunbit: I can explain @lferran , as i asked Nathan these changes yesterday night: https://github.com/nuclia/backend/pull/1145/files

I fact i see he forgot to add the create bucket changes on this pr i reference.

Also, aside of cypress e2e tests that rely on being able to create a kb and expect that to be deleted "quickly", i think deleting a kb to reuse afterwards shouldn't be a usual problem, but you may need to think about it. There's always the option to iterate the bucket objects and delete them (one by one or in batches of 100 IIRC), but this is time and $ consuming, that's why we do the schedule delete +  async delete (on purge) right now.

lferran (approved):

sunbit (approved):

PR #1552, opened 2023-11-07
date range in vectors

hermeGarcia:
### Description
In this PR date range queries have an effect in vectors  thanks to fields pre-filtering.

### How was this PR tested?
Local tests

### DIFF
--- a/nucliadb_texts/src/reader.rs
+++ b/nucliadb_texts/src/reader.rs
@@ -115,15 +115,15 @@ impl FieldReader for TextReaderService {
             };
             let resource_id = fulfilled_doc
                 .get_first(self.schema.uuid)
-                .expect("document doesn't appear to have uuid.")
+                .expect("documents must have a uuid.")
                 .as_text()
                 .expect("uuid field must be a text")
                 .to_string();
             let field_id = fulfilled_doc
                 .get_first(self.schema.field)
-                .expect("document doesn't appear to have field.")
+                .expect("documents must have a field.")
                 .as_facet()
-                .expect("field id must be a text")
+                .expect("field id must be a facet")
                 .to_path_string();
             let fulfilled_field = ValidField {
                 resource_id,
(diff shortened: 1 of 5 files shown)


nucliadb_texts/src/reader.rs
```diff
@@ -88,9 +90,48 @@ impl Debug for TextReaderService {
...
+        let docs_fulfilled = searcher.search(&pre_filter_query, &DocSetCollector)?;
+        let mut valid_fields = Vec::new();
+        for fulfilled_doc in docs_fulfilled {
+            let Ok(fulfilled_doc) = searcher.doc(fulfilled_doc) else {
+                error!("Could not retrieve {fulfilled_doc:?}");
+                continue;
+            };
+            let resource_id = fulfilled_doc
+                .get_first(self.schema.uuid)
+                .expect("documents must have a uuid.")
```
jotare, line 118: Same here, isn't expect too hard?

jotare (commented):
Could you add some test?

nucliadb_texts/src/reader.rs
hermeGarcia: If you look at search, we are unwrapping in the same situation. Either way I do think it should be an panic, if we are not indexing that data, which is part of our schema, something is very wrong.

nucliadb_texts/src/reader.rs
```diff
@@ -136,13 +177,13 @@ impl ReaderChild for TextReaderService {
...
         for addr in searcher.search(&AllQuery, &DocSetCollector)? {
-            let Some(key) = searcher
+            let key = searcher
                 .doc(addr)?
                 .get_first(self.schema.uuid)
-                .and_then(|i| i.as_text().map(String::from))
-            else {
-                continue;
-            };
+                .expect("documents must have a uuid.")
```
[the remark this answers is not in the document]
hermeGarcia: If we are indexing data without an uuid I think is appropriate to fail as spectacularly as possible, it would need to be fixed ASAP because data is being corrupted.

nucliadb_texts/src/reader.rs
```diff
@@ -88,9 +90,48 @@ impl Debug for TextReaderService {
...
+            };
+            if let Some(query) = search_query::produce_date_range_query(field, from, to) {
+                let query: Box<dyn Query> = Box::new(query);
+                timestamp_queries.push((Occur::Should, query));
+            }
+        }
+        let pre_filter_query = BooleanQuery::new(timestamp_queries);
+        let searcher = self.reader.searcher();
+        let docs_fulfilled = searcher.search(&pre_filter_query, &DocSetCollector)?;
+        let mut valid_fields = Vec::new();
```
lferran, line 110: Nice!
Is this collector iterating over all matching fields already? it handles pagination for you?

hermeGarcia: This will take all matching fields yes, without pagination since this is prefiltering. I need to get all the matching fields

lferran (approved):

PR #1553, opened 2023-11-07
Add user prompt to chat endpoint

lferran:
### Description

### How was this PR tested?


nucliadb/nucliadb/search/api/v1/chat.py
```diff
@@ -61,6 +61,14 @@ class SyncChatResponse(pydantic.BaseModel):
             "query": "Who won the league final?",
         },
     ),
+    "search_and_chat_with_custom_prompt": Example(
```
Ciniesta, line 64: Maybe I'd explain a bit more about the fromat of the prompt

For the settings in the dashboard we use this explanation:
"Input here your prompt with the words  {context} and {question} in brackets where you want those fields to be placed, in case you want them in your prompt.\n"
Maybe it'd be worth adding that context will be the data returned by the retrieval?

vangheem (approved):

PR #1570, opened 2023-11-09
Sidecar concurrent indexing

jotare:
### Description

### How was this PR tested?


nucliadb_node/nucliadb_node/dispatcher.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        listener_id: str,
+        cb: Callable,
+        priority: ListenerPriority = ListenerPriority.DONT_CARE,
+    ):
+        if listener_id in self.callbacks:
+            raise ValueError(f"Already registered a listener with id: {listener_id}")
+        self.callbacks[listener_id] = (cb, priority.value)
+
+    def add_async_listener(
```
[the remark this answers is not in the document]
vangheem: I would just remove `add_async_listener` and `remove_async_listener`, make the regular `add_listener ` API take async and non-async callables

dispatch is already async

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,353 @@
...
+class IndexerWorker:
+    def __init__(self, writer: Writer, storage: Storage):
+        self.work_queue = asyncio.PriorityQueue()
+        self.writer = writer
+        self.storage = storage
+        self.working = False
+
+        # store the brain we are working with while we index the message to have
+        # it in case of error
+        self.brain = None
```
vangheem, line 247: Can we remove this property and change where/how we handle errors so this isn't necessary?

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,353 @@
...
+            if grpc_error.code() == StatusCode.NOT_FOUND:
+                logger.error(f"Shard does not exist {pb.shard}")
+                return None
+            else:
+                event_id = errors.capture_exception(grpc_error)
+                logger.error(
+                    f"An error on subscription_worker. Check sentry for more details. Event id: {event_id}"
+                )
+                if (
+                    pb.typemessage == TypeMessage.CREATION
```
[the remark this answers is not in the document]
jotare: I moved a bit the exception handling so I can handle it in `set_resource` and removed the need to store the brain to do wonky things

nucliadb_utils/nucliadb_utils/nats/connection_manager.py
```diff
@@ -207,6 +172,7 @@ async def subscribe(
         cb: Callable[[Msg], Awaitable[None]],
         subscription_lost_cb: Callable[[], Awaitable[None]],
         flow_control: bool = False,
+        manual_ack: bool = False,
```
vangheem, line 175: hmm, we're manually acking everywhere right now I think. I didn't know about this control

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+            if created or (not created and not worker.working):
+                # The second condition should never happen, the worker task finished
+                # between us getting the worker and adding work. We'll create a new
+                # task for it
+                self.start_worker(split)
+
+            loop = asyncio.get_event_loop()
+            wait_condition = loop.create_future()
+
+            work_unit_done.add_listener(
```
vangheem, line 222: isn't this a memory leak? We never remove the listener.

Could this also be moved to a global class without a partial and then pass in the NatsDemultiplexer instance instead of passing in partials and having dynamic listeners?

vangheem (changes requested):

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,353 @@
...
+                # between us getting the worker and adding work. We'll create a new
+                # task for it
+                _ = self.start_worker(worker)
+
+            wait_condition = asyncio.Event()
+            success: bool
+            error = None
+
+            @only_for_seqid(seqid=seqid)
+            async def on_successful_indexing(payload: SuccessfulIndexingPayload):
```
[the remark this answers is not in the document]
vangheem: I think we can still move out of the dynamic listener registration approach and just pass in the data structures we need to operate against right?

I feel like code oriented event/subscriber pattern should be statically defined and not add the runtime complexity of dynamic listener management.

let me know if I'm not making sense or am missing necessary context that makes this not possible.

nucliadb_utils/nucliadb_utils/signals.py
```diff
@@ -0,0 +1,82 @@
...
+from inspect import iscoroutinefunction
+from typing import Any, Callable, Type, Union
+
+
+class ListenerPriority(Enum):
+    DONT_CARE = 0
+    CRITICAL = 1
+
+
+class Signal:
```
lferran, line 33: Can you add some docstring here to explain what this util class can be used for?

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+    NatsDemultiplexer relies on a DemuxProcessor. It's `splitter` function will
+    be used to decide the split and the resulting work unit will be added into a
+    queue.
+
+    Concurrent workers will get work from their respective queues and use
+    `DemuxProcessor.process` to process it. Depending on the result, the message
+    will be acked or nacked to NATS.
+
+    If an exception is raised during any of this process, the message will be
+    automatically nacked.
```
lferran, line 114: So it is responsibility of the specific worker to handle any unrecoverable exceptions and then ack, right?

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+
+            work_outcome = await wait_condition
+            self.working_on.discard(work_id)
+            if work_outcome.error:
+                await msg.nak()
+            else:
+                if work_outcome.success:
+                    await msg.ack()
+                else:
+                    await msg.nak()
```
lferran, line 239: though: do we need more logging in this method overall? To have evidence that things have been acked or nacked and why.

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+            )
+
+            work_outcome = await wait_condition
+            self.working_on.discard(work_id)
+            if work_outcome.error:
+                await msg.nak()
+            else:
+                if work_outcome.success:
+                    await msg.ack()
+                else:
```
lferran, line 238: how can this condition happen? a work either errors or succeeds no?

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+            raise task.exception()  # type: ignore
+
+    @staticmethod
+    async def on_work_unit_done(
+        work_id: str,
+        notifier: asyncio.Future,
+        payload: WorkOutcome,
+    ):
+        if payload.work_id == work_id:
+            notifier.set_result(payload)
```
lferran, line 271: can the `else` condition happen here? is it ok to ignore?

nucliadb_utils/nucliadb_utils/nats/demux.py
```diff
@@ -0,0 +1,310 @@
...
+    async def work_until_finish(self):
+        self.working = True
+        while not self.work_queue.empty():
+            work: WorkUnit = await self.work_queue.get()
+            await self.do_work(work)
+        self.working = False
+
+    async def do_work(self, work: WorkUnit):
+        try:
+            logger.info(f"Working on '{work.work_id}'", extra={"work_id": work.work_id})
```
lferran, line 296: for logging purposes, maybe we want to have the slipt logged too? or the callback func name.

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,184 @@
...
+                event_id = errors.capture_exception(grpc_error)
+                logger.error(
+                    "An error ocurred on indexer worker while deleting a resource. "
+                    f"Check sentry for more details. Event id: {event_id}"
+                )
+            return None
+
+        else:
+            logger.info(f"...done (Deleted {rid} in {shard_id} otx:{pb.txid})")
+            return status
```
lferran, line 184: I think that the status is no longer used, so you can omit returning anything from the method (and the same for the set_resource)

nucliadb_node/nucliadb_node/listeners/gc_scheduler.py
```diff
@@ -0,0 +1,147 @@
...
+            self._change_count = 0
+            try:
+                with gc_observer():
+                    # NOTE: garbage collector may not run if the shard is busy. We currently don't do anything to retry.
+                    await self._writer.garbage_collector(self._shard_id)
+                    logger.info(
+                        "Garbage collection finished", extra={"shard": self._shard_id}
+                    )
+            except Exception:
+                logger.exception(
```
lferran, line 104: do we want to send a sentry event with errors.capture_exception() in this case?

nucliadb_node/nucliadb_node/pull.py
```diff
@@ -61,369 +51,65 @@
...
-            uuid=indexpb.resource,
-            kbid=indexpb.kbid,
-            action=Notification.INDEXED,
-        )
-
-        await self.pubsub.publish(
-            const.PubSubChannels.RESOURCE_NOTIFY.format(kbid=indexpb.kbid),
-            message.SerializeToString(),
-        )
+    @subscriber_observer.wrap()
```
lferran, line 113: this metrics observer does not make sense anymore here, right? We should have observers at the workers/processors now

nucliadb_node/nucliadb_node/pull.py
```diff
@@ -61,369 +51,65 @@
...
-            options["user_credentials"] = indexing_settings.index_jetstream_auth
-
-        if len(indexing_settings.index_jetstream_servers) > 0:
-            options["servers"] = indexing_settings.index_jetstream_servers
-
-        self.nc = await nats.connect(**options)
-        self.js = get_traced_jetstream(self.nc, SERVICE_NAME)
-        logger.info(f"Nats: Connected to {indexing_settings.index_jetstream_servers}")
-        await self.subscribe()
+        await self.indexer.finalize()
```
lferran, line 78: Typically you want to finalize in the reverse order of initialization.
Otherwise you may be pushing new messages to an indexed that has finished / is finishing and that may cause errors.

nucliadb_utils/nucliadb_utils/signals.py
lferran: also, can you add unit tests to this module? it would help on understanding its usage too

nucliadb_utils/nucliadb_utils/nats/connection_manager.py
vangheem: Dang, I didn't know this! We should use this setting everywhere else as well.

nucliadb_utils/nucliadb_utils/nats/demux.py
jotare: Oh, it is... Good catch!

I'm not sure I follow everything. We can discuss it later

nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -189,7 +189,13 @@ class AuditSettings(BaseSettings):
...
     # Read about message ordering:
     #   https://docs.nats.io/nats-concepts/subject_mapping#when-is-deterministic-partitioning-needed
-    nats_max_ack_pending: int = 1  # required for strict message ordering
+
+    # NATS MaxAckPending controls how many messages can we handle simultaneously
+    # (look at the implementation for how concurrent indexing is implemented)
+    #
+    # NOTE this parameter don't actually change already existing consumers! If
+    # you want to update the value, you should use nats-cli to do so
+    nats_max_ack_pending: int = 10
```
lferran, line 198: Note: 

If you want to play safer, you could leave the default to `1` as it was, and then tweak it to `10` via chart values in stage until we are sure it works fine. And then finally bump this to 10.

nucliadb_node/nucliadb_node/tests/unit/test_indexer.py
```diff
@@ -0,0 +1,286 @@
...
+        processor_work_unit: WorkUnit,
+        writer_work_unit: WorkUnit,
+        writer_seqid: int,
+        processor_seqid: int,
+    ):
+        processor_work_unit.nats_msg.reply = f"x.x.x.x.x.{processor_seqid}"
+        writer_work_unit.nats_msg.reply = f"x.x.x.x.x.{writer_seqid}"
+
+        assert writer_work_unit < processor_work_unit
+        assert processor_work_unit > writer_work_unit
```
lferran, line 93: duplicated

nucliadb_node/nucliadb_node/tests/unit/test_indexer.py
```diff
@@ -0,0 +1,286 @@
...
+        assert writer_work_unit != processor_work_unit
+        assert processor_work_unit == processor_work_unit
```
lferran, line 95: did you mean to assert something else?

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,313 @@
...
+                "subject": msg.subject,
+                "reply": msg.reply,
+                "seqid": work.seqid,
+                "storage_key": work.index_message.storage_key,
+            },
+        )
+
+        indexer, created = await self.get_or_create_indexer(work.shard_id)
+        await indexer.index_soon(work)
+        await self.clean_idle_indexers()
```
lferran, line 143: nit: should this be a background task managed by the class instance instead? I'm fine as it is now too, but we will only clean indexers if new messages arrive

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,313 @@
...
+                "Message indexing finished",
+                extra={
+                    "seqid": work.seqid,
+                    "shard": work.index_message.shard,
+                    "storage_key": work.index_message.storage_key,
+                    "time": time.time() - start,
+                },
+            )
+        finally:
+            await work.mpu.end()
```
lferran, line 252: should this be handled in a context manager in `index_message_soon`? otherwise it feels weird to start the mpu in a higher level function and ending it here.

nucliadb_node/nucliadb_node/pull.py
lferran: it was a comment from my review 2 weeks ago, yup

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+
+    async def push_work_to_indexer(self, work: WorkUnit):
+        exists = work.shard_id in self.indexers
+        if exists:
+            indexer, _ = self.indexers[work.shard_id]
+            await indexer.index_soon(work)
+        else:
+            indexer = PriorityIndexer(writer=self.writer, storage=self.storage)
+            await indexer.index_soon(work)
```
vangheem, line 186: Set `self.indexers[work.shard_id] = (indexer, task)` immediately after you create it here so that the next msg that comes in uses the cached in instead of potentially creating multiple at the same time before setting it on the dict

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,337 @@
...
+
+class PriorityIndexer:
+    def __init__(self, writer: Writer, storage: Storage):
+        self.writer = writer
+        self.storage = storage
+        self.work_queue: asyncio.PriorityQueue = asyncio.PriorityQueue()
+        self.working = False
+
+    async def index_soon(self, work: WorkUnit):
+        await self.work_queue.put(work)
```
vangheem, line 212: maybe make this use `put_nowait` and make the function not async.

We do not have a max size set on the queue so there isn't any reason to use async version here

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,337 @@
...
+                "An error happened on indexer, all messages for this queue will be flushed. "
+                f"Check sentry for more details. Event id: {event_id}",
+                extra={
+                    "seqid": work.seqid,
+                    "shard": work.index_message.shard,
+                    "storage_key": work.index_message.storage_key,
+                },
+                exc_info=exc,
+            )
+            while not self.work_queue.empty():
```
vangheem, line 235: Alright so we're just going to wait for it to redeliver and not bubble an error into hard failure.

I wonder if we should also hard fail somehow but am okay starting here

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,337 @@
...
+        logger.info(f"Deleting {rid} in {shard_id} otx:{pb.txid}")
+        try:
+            status = await self.writer.delete_resource(resource)
+
+        except AioRpcError as grpc_error:
+            if grpc_error.code() == StatusCode.NOT_FOUND:
+                logger.error(f"Shard does not exist {pb.shard}")
+            else:
+                event_id = errors.capture_exception(grpc_error)
+                logger.error(
```
vangheem, line 329: should this bubble  the error?

vangheem (changes requested):
Nice job!

nucliadb_node/nucliadb_node/indexer.py
jotare: Changed to sync

nucliadb_node/nucliadb_node/indexer.py
jotare: `index_soon` is now sync, so there's no risk on adding first the work

nucliadb_node/nucliadb_node/indexer.py
jotare: We weren't bubbling it before, but if it makes sense I can do it

nucliadb_node/nucliadb_node/indexer.py
vangheem: cause an exit with code 1 to show the process failed

jotare: I see. I don't have an strong opinion on this. If you prefer some approach, we can go for it

nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -677,3 +685,17 @@ async def delete_kb(self, kbid: str = "", slug: str = "") -> str:
...
+        (
+            message.source is writer_pb2.BrokerMessage.MessageSource.WRITER
+            for message in messages
+        )
+    )
+    if from_writer:
+        source = nodewriter_pb2.IndexMessageSource.WRITER
+    else:
+        source = nodewriter_pb2.IndexMessageSource.PROCESSOR
+    return source
```
lferran, line 701: What happens if we we get messages from different sources in the same txn?

I know it may never be the case, but to be safer, maybe we can raise an error?

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,341 @@
...
+            # for this queue. We flush the queue and wait for the messages to be
+            # redelivered
+            event_id = capture_exception(exc)
+            logger.error(
+                "An error happened on an indexer, all messages for its queue will be flushed. "
+                f"Check sentry for more details. Event id: {event_id}",
+                extra={
+                    "seqid": work.seqid,
+                    "shard": work.index_message.shard,
+                    "storage_key": work.index_message.storage_key,
```
lferran, line 230: is the kbid part of the storage key? if not, it would be nice to have the kbid in the log extra too.

nucliadb/nucliadb/ingest/orm/processor/__init__.py
jotare: Sure, I can raise an error. Anyway, it'd surprise me if we are using this feature haha

nucliadb_node/nucliadb_node/pull.py
lferran: I dont think its necessary. we have this metric already from the nucliadb_telemetry package

nucliadb_node/nucliadb_node/listeners/gc_scheduler.py
jotare: Added a `capture_exception`

nucliadb/nucliadb/ingest/orm/processor/__init__.py
jotare: I don't feel comfortable raising an exception. I'll capture one and log an error but default to PROCESSOR. If we don't support multi broker message transactions, we should remove all the code related.

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,341 @@
...
+            task.add_done_callback(
+                partial(
+                    ConcurrentShardIndexer._indexer_done_callback, self, work.shard_id
+                )
+            )
+            self.indexers[work.shard_id] = (indexer, task)
+
+    def _indexer_done_callback(self, shard_id: str, task: asyncio.Task):
+        # autoremove indexer when finished
+        self.indexers.pop(shard_id)
```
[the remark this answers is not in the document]
jotare: I don't think so, but maybe @vangheem can give us a third opinion

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,453 @@
...
+        tasks = []
+        reverse_indexers = {}
+        for shard_id, (_, task) in self.indexers.items():
+            tasks.append(task)
+            reverse_indexers[task] = shard_id
+
+        if tasks:
+            # REVIEW timeout. Is 10 a good timeout? Should we implement some
+            # indexer finalize policy? Nak everything and close, for example?
+            done, pending = await asyncio.wait(tasks, timeout=10)
```
lferran, line 156: Normally when finalize is called is because a new version of the pod is being deployed right?
If it's simpler, nacking everything should be good enough.

However, you could also leave the wait + timeout and just cancel the pending ones?
Make sure the cancellation ends up with nacking the message though.

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,453 @@
...
+            if pending:
+                logger.error(
+                    f"{len(pending)} pending tasks are still running after concurrent shard indexer finalizes!"
+                )
+
+            for task in done:
+                if isinstance(task, Exception):
```
lferran, line 163: I think you need to parse exceptions with the `task.exception()` method instead

jotare: That's true, thanks!

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,454 @@
...
+        self.storage = await get_storage(service_name=SERVICE_NAME)
+
+    async def finalize(self):
+        tasks = []
+        reverse_indexers = {}
+        for shard_id, (_, task) in self.indexers.items():
+            tasks.append(task)
+            reverse_indexers[task] = shard_id
+
+        if tasks:
```
vangheem, line 153: nit:
```suggestion
        if len(tasks) > 0:
```

nucliadb_node/nucliadb_node/indexer.py
```diff
@@ -0,0 +1,454 @@
...
+                        exc_info=error,
+                    )
+
+        self.indexers.clear()
+
+    def index_message_soon(self, msg: Msg):
+        work = WorkUnit.from_msg(msg)
+
+        logger.info(
+            f"Index message for shard {work.shard_id} is being enqueued",
```
vangheem, line 177: `work.shard_id` is already in extra

vangheem: The advantage of not having variables in log messages is that you can group by those values.

vangheem (approved):
I think everything is okay.

There are some details here with asyncio implementation that are difficult to review very well and I think that might be what is making this PR a bit more difficult to reason about and review.

PR #1572, opened 2023-11-09
Perf tests on standalone

lferran:
### Description
- Add performance tests molotov script for standalone deploy
- Add middleware to record server response times (only used in debug mode)
- The test produces a json file compatible with the performance benchmark system we are using right now

### How was this PR tested?
local tests


lferran:
[sc-7817]

nucliadb_performance/Makefile
```diff
@@ -73,6 +74,21 @@ test-search-kb: export KBID=$(kbid)
...
 	echo " *  Test Started at: $$(date -u)"
 
-	molotov ./nucliadb_performance/test_search_kb.py $(sizing) --sizing-tolerance=$(error_tolerance) -w $(max_workers) --ramp-up $(ramp_up) -d $(duration_s) --force-shutdown
+	molotov ./nucliadb_performance/hosted/test_search_kb.py $(sizing) --sizing-tolerance=$(error_tolerance) -w $(max_workers) --ramp-up $(ramp_up) -d $(duration_s) --force-shutdown
+
+	echo " *  Test Finished at: $$(date -u)"
+
+
+.PHONY: test-standalone-search
+test-standalone-search: export READER_API=http://localhost:8080/api
```
jotare, line 83: NIT: you could use a variable for the base URL to be able to test against other URLs without touching the makefile

jotare (approved):

nucliadb_performance/Makefile
lferran: yes, totally right. I think if I have to extend this again I will switch to use config files though, as there start to be too many options :)

PR #1574, opened 2023-11-10
Improve progress reporting on imports

lferran:
### Description
The progress of import tasks is reported with `total` and `processed` by setting the total size of the import blob and the bytes read from the stream so far.

### How was this PR tested?
Existing tests should cover it

### DIFF
--- a/nucliadb/nucliadb/writer/api/v1/export_import.py
+++ b/nucliadb/nucliadb/writer/api/v1/export_import.py
@@ -98,22 +98,21 @@ async def start_kb_import_endpoint(request: Request, kbid: str):
         )
         return CreateImportResponse(import_id=import_id)
     else:
-        # TODO: Implement range/resumable uploads to better suppor big exports.
-        await upload_import_to_blob_storage(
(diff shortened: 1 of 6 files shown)


nucliadb/nucliadb/writer/api/v1/export_import.py
```diff
@@ -138,10 +137,13 @@ async def start_export_task(context: ApplicationContext, kbid: str, export_id: s
         raise
 
 
-async def start_import_task(context: ApplicationContext, kbid: str, import_id: str):
+async def start_import_task(
+    context: ApplicationContext, kbid: str, import_id: str, import_size: int
```
jotare, line 141: Should be this `max_import_size`?

```suggestion
    context: ApplicationContext, kbid: str, import_id: str, max_import_size: int
```

jotare (approved):

nucliadb/nucliadb/writer/api/v1/export_import.py
lferran: no, it´s the actual size of the import blob

PR #1575, opened 2023-11-10
Fix file upload bug with pg storage driver

lferran:
### Description
- Plug GCS and PG storage driver in CI tests
- Fix bug that was making pg storage driver crash on regular uploads

### How was this PR tested?
Many new tests added!

### DIFF
--- a/nucliadb/nucliadb/tests/integration/test_upload.py
+++ b/nucliadb/nucliadb/tests/integration/test_upload.py
@@ -39,8 +39,9 @@ async def test_upload(
             "X-Filename": base64.b64encode(b"testfile").decode("utf-8"),
             "X-Synchronous": "true",
             "Content-Type": "text/plain",
+            "Content-Length": str(len(content)),
         },
-        content=base64.b64encode(content),
+        content=content,
     )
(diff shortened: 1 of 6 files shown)


nucliadb/nucliadb/writer/tus/dm.py
```diff
@@ -98,8 +98,8 @@ def content_type(self):
         return self.metadata.get("content_type")
 
     @property
-    def size(self):
-        return self._data.get("size", 0)
```
lferran, line 102: the bug was here: we were returning strings instead of integers

nucliadb/nucliadb/tests/integration/test_upload.py
```diff
@@ -39,8 +39,9 @@ async def test_upload(
             "X-Filename": base64.b64encode(b"testfile").decode("utf-8"),
             "X-Synchronous": "true",
             "Content-Type": "text/plain",
+            "Content-Length": str(len(content)),
```
lferran, line 42: this was needed to reproduce the issue

jotare (approved):

PR #1586, opened 2023-11-15
Enable NATS manual_ack by default

jotare:
### Description
By default, JS subscribers will wrap the subscription callback to always ack the messages. As we are always manually ack/nak messages, this should be enabled.

### How was this PR tested?
Existing tests


vangheem (approved):

jotare:
Changes will be done in another PR: https://github.com/nuclia/nucliadb/pull/1595

PR #1592, opened 2023-11-17
Fix filtering without needing to reindex

lferran:
### Description

### How was this PR tested?


nucliadb_protos/nodereader.proto
```diff
@@ -8,7 +8,8 @@ import public "nucliadb_protos/utils.proto";
 
 
 message Filter {
-    repeated string tags = 1;
+    repeated string labels = 1;
```
lferran, line 11: rename to field_labels

nucliadb_paragraphs/src/search_query.rs
```diff
@@ -444,25 +443,58 @@ pub fn search_query(
...
+    search
+        .filter
+        .iter()
+        .flat_map(|f| f.paragraph_labels.iter())
+        .flat_map(|facet_key| Facet::from_text(facet_key).ok().into_iter())
+        .for_each(|facet| {
+            let facet_term = Term::from_facet(schema.facets, &facet);
+            let facet_term_query = TermQuery::new(facet_term, IndexRecordOption::Basic);
+            fuzzies.push((Occur::Should, Box::new(facet_term_query.clone())));
+            originals.push((Occur::Should, Box::new(facet_term_query)));
```
lferran, line 468: turn into OR (pl1 AND pl2 AND ...)

vangheem (approved):

nucliadb_core/src/query_planner.rs
```diff
@@ -146,25 +174,71 @@ impl From<SearchRequest> for QueryPlan {
...
+        pre_filter_request
+            .timestamp_filters
+            .extend(timestamp_filters);
+    }
+
+    // Labels filters
+    let request_has_labels_filters = search_request.filter.is_some()
+        && !search_request
+            .filter
+            .clone()
```
hermeGarcia, line 195: This is clone is not needed since you only want to check if `field_labels` is empty. Also, maybe the expression is a bit too complex and has hidden logic (it needs lazy evaluation). How about this:
```rust
    let request_has_labels_filters = search_request
        .filter
        .as_ref()
        .map(|i| i.field_labels.is_empty())
        .unwrap_or_default();
```

nucliadb_core/src/query_planner.rs
```diff
@@ -146,25 +174,71 @@ impl From<SearchRequest> for QueryPlan {
...
+    let request_has_timestamp_filters = search_request.timestamps.as_ref().is_some();
+    if request_has_timestamp_filters {
+        let timestamp_filters = compute_timestamp_pre_filters(search_request);
+        pre_filter_request
+            .timestamp_filters
+            .extend(timestamp_filters);
+    }
+
+    // Labels filters
+    let request_has_labels_filters = search_request.filter.is_some()
```
hermeGarcia, line 192: This is clone is not needed since you only want to check if `field_labels` is empty. Also, maybe the expression is a bit too complex and has hidden logic (it needs lazy evaluation). How about this:
```rust
    let request_has_labels_filters = search_request
        .filter
        .as_ref()
        .map(|i| i.field_labels.is_empty())
        .unwrap_or_default();
```

hermeGarcia (changes requested):
just one change

PR #1595, opened 2023-11-20
Concurrent nats tasks consumption 

lferran:
### Description
- Tasks nats consumer spawns a new asyncio task for every message (with a configurable limit)
- Add manual ack to nats conn manager
- Add unsubscribe to nats conn manager

### How was this PR tested?
unit and integration tests


lferran:
[sc-7643]

nucliadb/nucliadb/tasks/consumer.py
```diff
@@ -61,29 +65,49 @@ async def initialize(self, context: ApplicationContext):
...
         logger.info(
             f"Subscribed to {subject} on stream {stream}",
             extra={"consumer_name": self.name},
         )
 
+    async def _subscription_worker_as_task(self, msg: Msg):
+        seqid = int(msg.reply.split(".")[5])
+        task_name = f"NatsTaskConsumer({self.name}, msg={seqid})"
+        task = asyncio.create_task(self.subscription_worker(msg), name=task_name)
+        self.running_tasks.append(task)
```
jotare, line 109: Memory leak in `running_tasks`: you're always adding tasks but never removing them. You could add a `add_done_callback` to the task to autoremove itself of check for finished tasks before creating new ones

nucliadb/nucliadb/tasks/consumer.py
```diff
@@ -113,6 +137,10 @@ async def subscription_worker(self, msg: Msg):
             )
             try:
                 await self.callback(self.context, task_msg)  # type: ignore
+            except asyncio.CancelledError:
```
jotare, line 140: Doubt: `CancelledError` can be raised here?

nucliadb/nucliadb/tasks/consumer.py
```diff
@@ -45,13 +45,17 @@ def __init__(
         stream: const.Streams,
         callback: TaskCallback,
         msg_type: MsgType,
+        max_concurrent_messages: Optional[int] = None,
```
jotare, line 48: Doubt: `None` is the same as 1 right?

nucliadb/nucliadb/tests/integration/tasks/test_tasks.py
```diff
@@ -31,6 +31,17 @@
 
 pytestmark = pytest.mark.asyncio
 
+WORK_DONE = None
```
jotare, line 34: Be careful, tests relying on global variables can't run in parallel. I'd rather prefer to use a fixture instead or something like the tests you're doing below

jotare (changes requested):

nucliadb/nucliadb/tasks/consumer.py
lferran: None is the nats default, which is no limit on max_ack_pending

nucliadb/nucliadb/tasks/consumer.py
lferran: yes! if the task nats consumer is finalized and there are running tasks, it will cancel them, and therefore we need to nack the msgs.

nucliadb/nucliadb/tasks/consumer.py
lferran: good catch! fixing

nucliadb/nucliadb/tests/integration/tasks/test_tasks.py
lferran: fixed

nucliadb/nucliadb/tasks/consumer.py
jotare: We have to be careful with too many simultaneous tasks though

lferran: agree! I've set a maximum of 10 for exports imports and defaulted to settings.max_ack_pending if not set (which right now is set to 1)

PR #1601, opened 2023-11-22
Fix user classifications not indexed on reindex

lferran:
### Description
Paragraph user annotations were missed on reindex

### How was this PR tested?
Integration tests. I chose not to add more unit tests on purpose since the orm brain / resource logic needs a refactor soon.


jotare (approved):

lferran:
[sc-7479]

PR #1610, opened 2023-11-27
Use node writer response on set/delete resource

jotare:
### Description
Right now, errors returned by node writer are ignored in the sidecar. That means handled errors on node writer are ignored on sidecar and messages are being acked. We should fail if node writer returns an error!

### How was this PR tested?
- Unit test


nucliadb_node/nucliadb_node/pull.py
```diff
@@ -307,10 +307,13 @@ async def subscription_worker(self, msg: Msg):
...
             try:
+                status = None
                 if pb.typemessage == TypeMessage.CREATION:
-                    await self.set_resource(pb)
+                    status = await self.set_resource(pb)
                 elif pb.typemessage == TypeMessage.DELETION:
-                    await self.delete_resource(pb)
+                    status = await self.delete_resource(pb)
+                if status is not None and status.status != OpStatus.OK:
+                    raise Exception(status.detail)
```
lferran, line 316: Instead of a generic exception, I think it'd be better to create a custom for instance `NodeIndexOpError` or something like that

PR #1611, opened 2023-11-27
Fix various leaks

lferran:
### Description
- Driver abort transaction task leaks
- Train servicer session not closed
- Cleaner usage of semaphores

### How was this PR tested?
Unit tests


nucliadb/nucliadb/common/maindb/driver.py
```diff
@@ -94,4 +101,5 @@ async def transaction(
                 if error or wait_for_abort:
                     await txn.abort()
                 else:
-                    asyncio.create_task(txn.abort())
+                    abort_task = asyncio.create_task(txn.abort())
+                    self._abort_tasks.append(abort_task)
```
vangheem, line 105: If you're going to add to `self._abort_tasks`, this value will forever increase and never get cleaned right?

Don't you need a done callback that will remove the task from this list?

This is to address tests complaining?

vangheem (changes requested):

nucliadb/nucliadb/common/maindb/driver.py
lferran: Ouch, yes I'm dumb. Changed!

>> This is to address tests complaining?
No, this is fixing a memory leak in how we use asyncio in this case. Every transaction context manager was leaving a lingering task for the abort, with a reference to the transaction object too.

PR #1635, opened 2023-12-05
decouple metadata from shard cache

vangheem:
### Description

### How was this PR tested?


nucliadb_node/src/shards/change_state.rs
```diff
@@ -0,0 +1,109 @@
...
+                return Some(Arc::clone(shards.get(&shard_id).unwrap()));
+            }
+        }
+        if let Ok(mut shards) = self.shards.write() {
+            let sm = Arc::new(ShardChangeStateManager::new(
+                self.shards_path.join(shard_id.clone()),
+            ));
+            shards.insert(shard_id.clone(), Arc::clone(&sm));
+            return Some(sm);
+        }
```
hermeGarcia, line 65: Is possible that between line 55 and jumping to line 59, the shard is added or state managers are not shared between threads? Also, do we wont for the manager to stop working in presence of a `PoisonError`?

nucliadb_node/src/shards/change_state.rs
```diff
@@ -0,0 +1,109 @@
...
+impl ShardChangeStateManager {
+    pub fn new(shard_path: PathBuf) -> Self {
+        Self {
+            shard_path,
+            generation_id: RwLock::new(None),
+        }
+    }
+
+    pub fn get_generation_id(&self) -> Option<String> {
+        let gen_id_read = self.generation_id.read().unwrap();
```
hermeGarcia, line 79: I dont know if unwrapping in a `PoisonError` is the way to go.. maybe:
```rust
  let gen_id_read = self.generation_id.read().unwrap_or_else(|e| e.into_inner());
```

hermeGarcia (changes requested):

nucliadb_node/src/shards/providers/unbounded_cache/unbounded_writer.rs
```diff
@@ -96,13 +99,21 @@ impl ShardWriterProvider for UnboundedShardWriterCache {
         let mut cache = self.write();
         for entry in std::fs::read_dir(&self.shards_path)? {
             let entry = entry?;
-            let file_name = entry.file_name().to_str().unwrap().to_string();
+            let shard_id = entry.file_name().to_str().unwrap().to_string();
             let shard_path = entry.path();
-            match ShardWriter::open(file_name.clone(), &shard_path) {
+            let metadata = self.metadata_manager.get(shard_id.clone());
```
hermeGarcia, line 104: Doing this is a problem if you serialise the metadata this way:
```rust
      let mut writer = BufWriter::new(File::create(metadata_path)?);
        serde_json::to_writer(
            &mut writer,
            &ShardMetadataFile {
                kbid: self.kbid.clone(),
                similarity: self.similarity,
                id: Some(self.id.clone()),
                channel: self.channel,
            },
        )?;
        writer.flush()?;
```
The problem is that you may read a partially written file, which can lead in undefined behaviour. I think the only way to this is by achieving atomicity in writing the metadata file.
A way of doing this is by using the `create and rename` strategy, basically, write the file a temp file and the rename it to the metadata file.

hermeGarcia (approved):
Only one point that I think need to be addressed! Happy to talk about it!

hermeGarcia (approved):

PR #1647, opened 2023-12-08
Resumable rollovers

vangheem:
### Description

The purpose of this PR is to allow rollovers to be resumable.

This utilizes our k/v store to keep a record of the current migration process. This means that we have to do quite a few writes to our k/v store during migrations.

It would be better to be able integrate a nats stream; however, we do not have that available to us for on prem.

### How was this PR tested?


nucliadb/nucliadb/common/datamanagers/rollover.py
```diff
@@ -0,0 +1,151 @@
...
+        start_key = KB_ROLLOVER_RESOURCES_INDEXED.format(kbid=kbid, resource="")
+        async with self.driver.transaction(wait_for_abort=False) as txn:
+            async for key in txn.keys(match=start_key, count=-1):
+                yield key.split("/")[-1]
+
+    async def _get_batch_indexed_data(
+        self, kbid, batch: list[str]
+    ) -> list[tuple[str, tuple[str, int]]]:
+        async with self.driver.transaction(wait_for_abort=False) as txn:
+            values = await txn.batch_get(
```
lferran, line 124: I wonder if this batch get can be problematic? Sometimes a single index message can be too big to fit into the max grpc payload of tikv, no?

lferran: ah, nevermind. I see it now. I thought you were storing the whole brain message in tikv. sorry

lferran (approved):

PR #1649, opened 2023-12-10
fix sdk async client

### LINKED ISSUE #1648: SDK Client throwing error when using the async client
**Describe the bug**
Using the SDK with the **Async** Client is throwing an exception, while the same code but using the Sync client works fine.

**Steps to reproduce (if applicable)**

Example code to reproduce the error:
```python
from nucliadb_sdk import NucliaDBAsync, Region

sdk = NucliaDBAsync(region=Region.ON_PREM, url="http://localhost:8080/api")

kb = await sdk.create_knowledge_box(slug="anything")
```

The previous code throws this exception:

```
  File "/home/david/Desktop/nuclia/nuclia/routers/kbs.py", line 23, in create_knowledge_box
    kb = await sdk.create_knowledge_box(slug=body.slug)
  File "/home/david/Desktop/nuclia/.venv/lib/python3.10/site-packages/nucliadb_sdk/v2/sdk.py", line 205, in _func
    resp = self._request(
TypeError: NucliaDBAsync._request() got an unexpected keyword argument 'content'
```

**Expected behavior**
It doesn't raise an exception and creates a knowledge box in the DB

**Configuration:**
Please provide:

1. Output of `pip list | grep nucliadb`

```
nucliadb-client                          3.0.0.post726
nucliadb-models                          2.34.0.post918
nucliadb-protos                          2.34.0.post918
nucliadb-sdk                             2.34.0.post918
nucliadb-telemetry                       2.35.0.post26
```

drf7:
### Description
This is just a proposal fix for the #1648 issue
Fixes #1648

### How was this PR tested?
Manually tested


nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -202,9 +202,14 @@ def _func(self: "NucliaDB", content: Optional[Any] = None, **kwargs):
             raise TypeError(f"Invalid arguments provided: {kwargs}")
 
         if not stream_response:
-            resp = self._request(
-                path, method, data=data, query_params=query_params, content=raw_content
-            )
+            if isinstance(self, NucliaDBAsync):
```
lferran, line 205: I think the real problem is that the `_request` method of the `NucliaDBAsync` class is missing the `content` parameter. See: https://github.com/nuclia/nucliadb/blob/main/nucliadb_sdk/nucliadb_sdk/v2/sdk.py#L843-L849

Can you add it? Then you don't need this conditional check

lferran (changes requested):
Hi @drf7! Thank you for taking the time on reporting this.

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
drf7: Thanks for the review @lferran. I have implemented your suggested change. I feel like most of the logic should be moved to the base class instead, but as a quick fix I think this is good enough

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -846,18 +846,22 @@ async def _request(
...
+        if content is not None:
+            opts["content"] = content
         if query_params is not None:
             opts["params"] = query_params
         response: httpx.Response = await getattr(self.session, method.lower())(
             url, **opts
         )
-        self._check_response(response)
-        return response
+        return self._check_response(response)
```
lferran, line 864: Careful here! 😜 the response should be returned

drf7: `self._check_response(response)` is also returning the response, I followed a similar syntax as the sync method 😁: https://github.com/nuclia/nucliadb/blob/main/nucliadb_sdk/nucliadb_sdk/v2/sdk.py#L747-L766

lferran: oh you're right! 👍🏼

lferran (approved):

nucliadb_sdk/nucliadb_sdk/v2/sdk.py
```diff
@@ -167,7 +167,7 @@ def _request_builder(
     stream_response: bool = False,
     docstring: Optional[docstrings.Docstring] = None,
 ):
-    def _func(self: "NucliaDB", content: Optional[Any] = None, **kwargs):
+    def _func(self: "NucliaDB | NucliaDBAsync", content: Optional[Any] = None, **kwargs):
```
lferran, line 170: sorry, I think the linter is complaining about this line in the CI.

Can you change it to 
```
    def _func(
        self: "NucliaDB | NucliaDBAsync", content: Optional[Any] = None, **kwargs
    ):
``` 
instead?

You can run `make python-code-lint` locally to make sure there are no formatting issues

drf7: Yes, my bad, updated the code 😃

lferran:
@drf7 yet another mypy complain. Feel free to add a `# type: ignore` comment to fix it.
We are planning on improving the type annotations of the sdk in the coming weeks, so we will address this too.

lferran:
Thank you @drf7 for your contribution 👏🏼 !

PR #1655, opened 2023-12-11
disable generation id cache

vangheem:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/replication/replicator.rs
+++ b/nucliadb_node/src/replication/replicator.rs
@@ -175,9 +175,7 @@ impl ReplicateWorkerPool {
     }
 
     pub async fn add<F>(&mut self, worker: F) -> NodeResult<()>
-    where
-        F: Future<Output = NodeResult<()>> + Send + 'static,
-    {
+    where F: Future<Output = NodeResult<()>> + Send + 'static {
         let work_lock = Arc::clone(&self.work_lock);
         let permit = work_lock.acquire_owned().await.unwrap();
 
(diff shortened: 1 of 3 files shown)


hermeGarcia (approved):
Approved, although I must say Im not a fan of commented code.

nucliadb_node/src/replication/replicator.rs
```diff
@@ -134,9 +134,14 @@ pub async fn replicate_shard(
     drop(file);
     drop(_gc_lock);
 
-    if generation_id.is_some() {
+    if let Some(gen_id) = generation_id {
```
hermeGarcia, line 137: just a comment, on line 87 you are doing this:
```rust
generation_id = Some(resp.generation_id);
```
which makes this code kind of unnecessary, maybe there is some logic you are not taking into account? If not, why not doing:
```rust
shard.metadata.set_generation_id(resp.generation_id);
```
and remove the mutable `gen_id` var that only seems to add noise.

hermeGarcia (changes requested):
Just found something that I think you should check

nucliadb_node/src/replication/replicator.rs
vangheem: oh, that is weird and needs to be cleaned up! Thanks :)

vangheem: oh, I was trying not to set the gen id in the while

hermeGarcia: But you have it already in `resp`, you dont need to declare anything. Specially because I see its not used after the loop.

vangheem: the response is the scope of the while loop. Technically, the stream could be empty and we have no value

hermeGarcia: I see

hermeGarcia (approved):

PR #1664, opened 2023-12-13
Garbage collection returns a status

jotare:
### Description
Really naive approach to GC returning a status via gRPC. It allows to control when garbage collection can't run due to the writer or a hard error

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/shards/shard_writer.rs
+++ b/nucliadb_node/src/shards/shard_writer.rs
@@ -28,6 +28,7 @@ use nucliadb_core::protos::{OpStatus, Resource, ResourceId, VectorSetId, VectorS
 use nucliadb_core::tracing::{self, *};
 use nucliadb_core::{thread, IndexFiles};
 use nucliadb_procs::measure;
+use nucliadb_vectors::VectorErr;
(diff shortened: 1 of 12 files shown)


vangheem (approved):

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -52,6 +53,11 @@ pub struct ShardWriter {
     write_lock: Mutex<()>,  // be able to lock writes on the shard
 }
 
+pub enum GarbageCollectorStatus {
+    GarbageCollected,
+    TryLater,
+}
+
```
hermeGarcia, line 60: if you implement:
```rust
impl From<GarbageCollectorStatus> for nucliadb_core::protos::garbage_collector_response::Status {
   .....
}
```
The implementation of `gc` gets less noisy.

hermeGarcia (approved):
Looks good, just one comment

nucliadb_node/src/shards/shard_writer.rs
jotare: Done

PR #1672, opened 2023-12-15
do not audit empty user

vangheem:
### Description
Do not audit when no user is provided.

### How was this PR tested?


nucliadb/nucliadb/search/search/chat/query.py
```diff
@@ -316,6 +316,10 @@ async def maybe_audit_chat(
     chat_history: List[ChatContextMessage],
     query_context: List[str],
 ):
+    if user_id != "_INTERNAL_":
```
lferran, line 319: should this be `==`?

PR #1687, opened 2023-12-21
Add sorted paragraphs in find results

lferran:
### Description

### How was this PR tested?


jotare (approved):
I feel kind of weird having to add this new `best_matches` to return a list of ids while we could be returning a list of ordered results instead of a dict

PR #1695, opened 2023-12-26
fix hnsw param use

hermeGarcia:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_vectors/src/data_point/params.rs
+++ b/nucliadb_vectors/src/data_point/params.rs
@@ -28,7 +28,7 @@ pub fn level_factor() -> f64 {
 
 /// Upper limit to the number of out-edges a embedding can have.
 pub const fn m_max() -> usize {
-    30
+    60
 }
 
 /// Number of bi-directional links created for every new element.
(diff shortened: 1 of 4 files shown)


nucliadb_vectors/src/data_point/params.rs
```diff
@@ -28,7 +28,7 @@ pub fn level_factor() -> f64 {
 
 /// Upper limit to the number of out-edges a embedding can have.
 pub const fn m_max() -> usize {
-    30
+    60
```
jotare, line 31: What implies increasing this number?

(doubt: why are we using a const fn instead of a const variable?)

hermeGarcia: It has several implications both in terms of space and  read-write speed. To learn more about its implications I suggest looking at [hnsw paper](https://arxiv.org/pdf/1603.09320.pdf), in particular to the insert algorithm. For our case, we had `m` and `m_max` set to the same number, which makes insertion slower (as can be inferred by looking at the algorithm). From what I saw defining `m_max` as `2*m` when `m` is small is a good approach.

As for why is a `const fn` and not a constant, I did it for uniformity's sake, since `level_factor` is a function. That said, I do not mind moving it to a constant.

lferran (approved):

PR #1702, opened 2024-01-04
Endpoint to get notifications of KB activity

lferran:
### Description
Adds the `activity` endpoint on the reader component that allows to get a stream of notifications for resource operations for a particular KB.

### How was this PR tested?
Unit and integration tests


nucliadb_models/nucliadb_models/activity.py
```diff
@@ -0,0 +1,60 @@
...
+    DELETED = "deleted"
+
+
+class NotificationAction(str, Enum):
+    COMMIT = "commit"
+    INDEXED = "indexed"
+    ABORTED = "aborted"
+
+
+class ResourceNotification(BaseModel):
```
lferran, line 37: TODO: re-model this to have a more generic notification stream (not only for resource activity, but potentially for other things)

TODO: Confirm with @ebrehault that the model and the UX is OK

vangheem (approved):

PR #1703, opened 2024-01-04
Adding security to index node search

lferran:
### Description
This PR adds security to search for the index node.

A new facet field `groups_with_access` has been added to the Tantivy texts index, where we store the list of group ids that have access to each field/document.
If a document does not have any group id provided, it is regarded as public -- all groups have access to it.

Search requests now have a `security` parameter:
```
message Security {
   repeated string access_groups = 1;
}
```
that allows to get search results on the subset of documents that the groups in `access_groups` have access to.
If no `access_groups` are specified, then the security lookup is omitted.

### How was this PR tested?
Rust integration tests


nucliadb_core/src/query_planner.rs
```diff
@@ -175,10 +176,17 @@ impl From<SearchRequest> for QueryPlan {
 
 fn compute_pre_filters(search_request: &SearchRequest) -> Option<PreFilterRequest> {
     let mut pre_filter_request = PreFilterRequest {
+        security: None,
         timestamp_filters: vec![],
         labels_filters: vec![],
     };
 
+    // Security filters
+    let request_has_security_filters = search_request.security.is_some();
```
lferran, line 185: make sure this is correct! should we check with something similar to python's `.HasField("security")`?

nucliadb_texts/src/writer.rs
```diff
@@ -296,6 +296,26 @@ impl TextWriterService {
             self.schema.status => resource.status as u64,
         );
 
+        if self.schema.schema.get_field("groups_with_access").is_some() {
```
lferran, line 299: I think this makes it possible to change the schema in a bw/compatible way.
I need to verify it though.

lferran: nah, old versions of the index break on loading. we'll need another version of the texts library

PR #1707, opened 2024-01-08
Add security search to index node

lferran:
### Description
Adds security search to the index node implementing it into nucliadb_texts2
Integrates nucliadb_texts2 to index node

### How was this PR tested?
Rust integration test


nucliadb_texts2/src/writer.rs
```diff
@@ -296,6 +296,23 @@ impl TextWriterService {
             self.schema.status => resource.status as u64,
         );
 
+        let public_to_all_groups = resource.security.is_none()
+            || resource.security.as_ref().unwrap().access_groups.is_empty();
```
hermeGarcia, line 300: Although this logic is safe I am not a huge fan of the behaviour it hides, basically it assumes lazy evaluation and that is why unwrapping in the right-hand side is safe. I would suggest the following:
```rust
        let public_to_all_groups = resource
            .security
            .as_ref()
            .map_or(true, |i| i.access_groups.is_empty());
```

nucliadb_texts2/src/writer.rs
```diff
@@ -296,6 +296,23 @@ impl TextWriterService {
...
+        if public_to_all_groups {
+            base_doc.add_u64(self.schema.groups_public, 1_u64);
+        } else {
+            base_doc.add_u64(self.schema.groups_public, 0_u64);
+            for group_id in resource.security.as_ref().unwrap().access_groups.iter() {
```
hermeGarcia, line 305: You are unwrapping here because the logic of the block ensures its safe, but there is a better way of encoding this that does not need that much logic tracking:
```rust
        let resource_security = resource.security.as_ref();
        if let Some(security_groups) = resource_security.filter(|i| !i.access_groups.is_empty()) {
            base_doc.add_u64(self.schema.groups_public, 0_u64);
            for group_id in security_groups.access_groups.iter() {
                let mut group_id_key = group_id.clone();
                if !group_id.starts_with('/') {
                    // Slash needs to be added to be compatible with tantivy facet fields
                    group_id_key = "/".to_string() + group_id;
                }
                let facet = Facet::from(group_id_key.as_str());
                base_doc.add_facet(self.schema.groups_with_access, facet)
            }
        } else {
            base_doc.add_u64(self.schema.groups_public, 1_u64);
        }
```

nucliadb_texts2/src/reader.rs
```diff
@@ -92,10 +92,25 @@ impl FieldReader for TextReaderService {
...
+                let mut group_id_key = group_id.clone();
+                if !group_id.starts_with('/') {
+                    // Slash needs to be added to be compatible with tantivy facet fields
+                    group_id_key = "/".to_string() + group_id;
+                }
+                let facet = Facet::from_text(&group_id_key).unwrap();
+                let term = Term::from_facet(self.schema.groups_with_access, &facet);
+                let term_query = TermQuery::new(term, IndexRecordOption::Basic);
+                access_groups_queries.push(Box::new(term_query));
+            }
```
hermeGarcia, line 111: Not really the "rustic" way of doing this, I would suggest:
```rust
if let Some(security) = request.security.as_ref() {
      for group_id in security.access_groups.iter() {
                let mut group_id_key = group_id.clone();
                if !group_id.starts_with('/') {
                    // Slash needs to be added to be compatible with tantivy facet fields
                    group_id_key = "/".to_string() + group_id;
                }
                let facet = Facet::from_text(&group_id_key).unwrap();
                let term = Term::from_facet(self.schema.groups_with_access, &facet);
                let term_query = TermQuery::new(term, IndexRecordOption::Basic);
                access_groups_queries.push(Box::new(term_query));
        }
}
```

hermeGarcia (approved):

PR #1709, opened 2024-01-09
Activity notification fixes

lferran:
### Description
- Extend the `Notification` proto to include the originator/source of the resource operation: this allows users to distinguish writer vs processor message notifications.
- Include resource uuid on resource indexed notifications

### How was this PR tested?
Local tests + existing tests should cover this


lferran:
[sc-8230]

PR #1714, opened 2024-01-10
Don't suggest resources

javitonino:
### Description
Avoid returning suggestions that match a prefix search on the resource ID (can happen under normal use for UUIDs & searches starting with a-f)

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/shards/shard_reader.rs
+++ b/nucliadb_node/src/shards/shard_reader.rs
@@ -344,24 +344,12 @@ impl ShardReader {
                     .filter(|prefix| prefix.len() >= MIN_VIABLE_PREFIX_SUGGEST)
                     .cloned()
                     .map(|prefix| RelationSearchRequest {
-                        shard_id: String::default(), // REVIEW: really?
                         prefix: Some(RelationPrefixSearchRequest {
                             prefix,
-                            // Any type that is not Resource (avoid suggesting resource UUIDs)
-                            node_filters: vec![
-                                RelationNodeFilter {
-                                    node_type: NodeType::Entity.into(),
-                                    ..Default::default()
-                                },
-                                RelationNodeFilter {
-                                    node_type: NodeType::Label.into(),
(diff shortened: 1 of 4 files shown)


nucliadb_node/src/shards/shard_reader.rs
```diff
@@ -345,7 +347,21 @@ impl ShardReader {
...
+                                },
+                                RelationNodeFilter {
+                                    node_type: NodeType::Label.into(),
+                                    ..Default::default()
+                                },
+                                RelationNodeFilter {
+                                    node_type: NodeType::User.into(),
+                                    ..Default::default()
+                                },
+                            ],
```
lferran, line 364: I think that the expectation for this feature is that only entities are suggested.

javitonino: :thinking: That makes sense for me as well, should we confirm it with someone else? Or just change it to filter only entities?

lferran: yup, good idea. let me confirm with @ebrehault

nucliadb_node/src/shards/shard_reader.rs
```diff
@@ -345,7 +347,21 @@ impl ShardReader {
                         shard_id: String::default(), // REVIEW: really?
```
hermeGarcia, line 347: Since you are modifying this part of the code, do you mind removing this line? Is not necessary since we use `..Default::default()` at the end.

nucliadb_node/src/shards/shard_reader.rs
javitonino: As another data point: the current tests check that Users are also suggested, but I think it might be a mistake since the comments explicitly mention entities. It's probably just copypasting what the function returned as the expected value without double-checking.

lferran (approved):
Nice first contribution to NucliaDB @javitonino 👏🏼  for many more 🍾

PR #1720, opened 2024-01-12
Remove async cache provider

javitonino:
### Description

- Removes async cache, unified into a single implementation of cache (sync).
- Copied safe-deletion logic from async to sync cache.
- Changed grpc servers and replication to use the sync cache, obtaining the shards from inside a blocking task (`spawn_blocking`).

### How was this PR tested?

Tests and basic tinkering. Will probably require some performance testing in stage.


nucliadb_node/src/shards/providers/unbounded_cache/unbounded_reader.rs
```diff
@@ -79,19 +85,20 @@ impl ShardReaderProvider for UnboundedShardReaderCache {
     fn load_all(&self) -> NodeResult<()> {
         let mut cache = self.write();
         let shards_path = self.shards_path.clone();
-        debug!("Recovering shards from {shards_path:?}...");
+
```
javitonino, line 88: :warning: Behaviour for `load_all` differed between the sync & async version:
- The async version loaded all shards and then grabbed the lock to insert all shards into cache. 
- The sync version held the lock while loading all shards.

This version behaves like the sync version and thus blocks the cache until everything is loaded. It seems safer but potentially a long block. The async version could be prone to race-conditions if a shard is opened while `load_all` is running. A third alternative would be to grab a write lock for each iteration of the loop, so other consumers get the chance to interact with the cache (albeit with high contention).

nucliadb_node/src/shards/providers/provider_traits.rs
```diff
@@ -55,17 +46,3 @@ pub trait ShardWriterProvider {
 
     fn get_metadata(&self, id: ShardId) -> Option<Arc<ShardMetadata>>;
 }
```
hermeGarcia, line 48: How about removing all the trait parafernalia? With this PR we will only have one provider (the unbounded one) so maybe is a little soon for this generalisations that at the moment only add overhead. 
What do you think?

hermeGarcia (approved):

PR #1724, opened 2024-01-12
Update Q&A dataset training

jotare:
### Description

### How was this PR tested?


nucliadb/nucliadb/train/generators/question_answer_streaming.py
```diff
@@ -106,17 +105,23 @@ async def iter_stream_items(
     question_answer_pb: QuestionAnswer,
 ) -> AsyncGenerator[QuestionAnswerStreamItem, None]:
     question_pb = question_answer_pb.question
+    question_paragraphs = [
+        await get_paragraph(kbid, paragraph_id)
```
lferran, line 109: just one thought: do we want to try/catch not found errors on the get_paragraph just in case the annotation points to a wrong paragraph id?

lferran (approved):

nucliadb/nucliadb/train/generators/question_answer_streaming.py
jotare: Sure, I'll change it

PR #1734, opened 2024-01-16
More idiomatic replication settings using Duration

jotare:
### Description
Use `std::time::Duration` instead of an `u64` for replication settings

### How was this PR tested?


hermeGarcia (approved):
Much better!

PR #1738, opened 2024-01-18
New LRU cache implementation

javitonino:
New cache implementation:
- Able to take shards out of the cache safely.
- LRU cache to control resource usage. For now, unbounded.
- Able to load multiple shards in parallel and avoid blocking other operations while doing so.

Future work:
- Bounded cache & settings
- Cache usage metrics


nucliadb_node/Cargo.toml
```diff
@@ -96,6 +96,7 @@ hostname = "0.3"
 username = "0.2"
 md5 = "0.7"
 lazy_static = "1.4.0"
+lru = "0.12.1"
```
javitonino, line 99: This was already in our dependency tree (via tantivy) so this should be pretty lightweigh.

nucliadb_node/src/shards/providers/shard_cache/reader_cache.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use super::resource_cache::{CacheResult, ResourceCache};
+use crate::disk_structure;
+use crate::settings::Settings;
+use crate::shards::errors::ShardNotFoundError;
+use crate::shards::metadata::ShardMetadata;
+use crate::shards::reader::ShardReader;
+use crate::shards::ShardId;
+
+pub struct ShardReaderCache {
```
javitonino, line 33: This could just be a couple of functions. I decided to keep it as an struct for symmetry with the ShardWriter, which works better with an struct to keep track of shards under deletion.

nucliadb_node/src/shards/providers/shard_cache/resource_cache.rs
```diff
@@ -0,0 +1,427 @@
...
+}
+
+impl<K> Drop for ResourceLoadGuard<K> {
+    fn drop(&mut self) {
+        self.waiter.notify();
+    }
+}
+
+// Use to wait until an entry is ready to be used
+pub struct ResourceWaiter(Arc<Waiter>);
```
javitonino, line 70: This, and the `Waiter` could be removed and just let the consumer loop/sleep until the resource is ready. This just makes the wait more efficient by notifying when it's actually ready.

nucliadb_node/src/shards/providers/shard_cache/resource_cache.rs
```diff
@@ -0,0 +1,427 @@
...
+    pub fn insert(&mut self, k: &K, v: &Arc<V>) {
+        if self.live.len() >= self.capacity.unwrap_or(NonZeroUsize::MAX).into() {
+            self.evict();
+        }
+        self.live.push(k.clone(), Arc::clone(v));
+    }
+
+    fn evict(&mut self) {
+        if let Some((evicted_k, evicted_v)) = self.live.pop_lru() {
+            println!("Evict {evicted_k:?}");
```
hermeGarcia, line 169: I dont think we want this println in the actual code. Right?

javitonino: Uh, it shouldn't. I missed it. Thanks! Good :eye:

hermeGarcia (approved):

PR #1739, opened 2024-01-18
Use monotonic timers for metrics

javitonino:
### Description
A lot of metrics were using `SystemTime` when it's more appropriate to use `Instant` for measuring elapsed time. This PR changes all of those to use `Instant`.

It does not change it in some places where it was persisted to disk and compared across executions.

Of particular note: https://github.com/nuclia/nucliadb/compare/monotonic-timers?expand=1#diff-9da5113fe7de5eb5858d773e4e90b81a2f8abeaff96517361fd3e2f3c9cc8fb4L352 which could panic the replication process when measuring time across a time-discontinuity (usually due to a NTP clock adjustment).

This PR _might_ just be the results of a regex search/replace and some manual review. :roll_eyes: 

### How was this PR tested?

### DIFF
--- a/nucliadb_texts2/src/writer.rs
+++ b/nucliadb_texts2/src/writer.rs
@@ -22,7 +22,7 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fs;
 use std::path::Path;
-use std::time::SystemTime;
+use std::time::Instant;
 
 use nucliadb_core::prelude::*;
 use nucliadb_core::protos::resource::ResourceStatus;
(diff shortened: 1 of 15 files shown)


nucliadb_texts2/src/writer.rs
```diff
@@ -80,35 +80,32 @@ impl WriterChild for TextWriterService {
...
-            debug!("{id:?} - Indexing document: starts at {v} ms");
-        }
         if resource.status != ResourceStatus::Delete as i32 {
             self.index_document(resource);
         }
-        if let Ok(v) = time.elapsed().map(|s| s.as_millis()) {
-            debug!("{id:?} - Indexing document: starts at {v} ms");
-        }
+        let v = time.elapsed().as_millis();
+        debug!("{id:?} - Indexing document: starts at {v} ms");
```
hermeGarcia, line 99: Should say "ends" not "starts", was already wrong before.

hermeGarcia (approved):
Looks good to me! Just one comment.

PR #1741, opened 2024-01-18
Upgrade to pydantic2

anastasds:
### Description
Nuclia uses Pydantic 1, which makes it incompatible with anything that uses Pydantic 2. This PR updates the codebase to Pydantic 2 and pulls in the new `pydantic2-argparse` library that I forked from the abandoned `pydantic-argparse` for Pydantic2 support after getting no response from its maintainer.

### How was this PR tested?
Hoping from some support from the team because:
* There is a circular dependency between `nucliadb` and `nuclia.py`
* There is an unsatisfiable dependency issue with `botocore` so I cannot build the project locally


nucliadb/nucliadb/common/cluster/settings.py
```diff
@@ -19,7 +19,10 @@
 #
 import enum
 
-from pydantic import BaseSettings, Field
+try:
+    from pydantic.v1 import BaseSettings, Field
+except ImportError:
+    from pydantic import BaseSettings, Field
```
lferran, line 25: Since nucliadb is not meant to be used as a library, can't we assume that we will have a pydantic version `>=2.0.0` and simply have `from pydantic.v1 import BaseSettings, Field` ?

This doesn't apply to the libraries like:
- nucliadb_telemetry
- nucliadb_models
- nucliadb_utils
- nucliadb_sdk

For the sake of this contribution, if you want you can limit it to migrate anything under the `nucliadb/` folder and replacing pydantic-argparse dependency to your fork. What do you think?

nucliadb/nucliadb/standalone/purge.py
```diff
@@ -20,7 +20,13 @@
 # Standalone purge command
 import asyncio
 
-import pydantic_argparse
+try:
+    import pydantic.v1 as pydantic
+    import pydantic2_argparse as pydantic_argparse
```
lferran, line 25: Here aswell: you can assume that the nucliadb package version will come with pydantic2_argparse from now on, no need to keep backwards compatibility.

Just make sure to set the right version dependencies in the `requirements.txt` and `requirements.lock.txt`

PR #1743, opened 2024-01-19
Fix storage uploads: use preferred chunk size

lferran:
### Description

When onprem, processed files need to be downloaded from NUA processing api and uploaded to NucliaDB's blob storage.
So far we had never hit this issue because we were using always PG and local file backends, which have no restrictions on chunk sizes.

However, when uploading to the GCS bucket we get these errors:
```
Giving up _append(...) after 4 tries (nucliadb_utils.storages.gcs.GoogleCloudException: 400: Invalid request.  The number of bytes uploaded is required to be equal or greater than 262144, except for the final request (it's recommended to be the exact multiple of 262144).  The received request contained 5813 bytes, which does not meet this requirement.)
```
This is something that I had fixed for the export/import, but it is hitting us now on prem.

### How was this PR tested?
Unit tests


lferran:
[sc-8583]

PR #1755, opened 2024-01-22
safe tantivy replication

hermeGarcia:
### Description
This PR uses a different approach to replication in Tantivy indexes. We were getting errors due to data race conditions, which are a result of us directly accessing Tantivy files without proper safeguards. 

### How was this PR tested?
Local tests.

### DIFF
--- a/nucliadb_core/src/tantivy_replica.rs
+++ b/nucliadb_core/src/tantivy_replica.rs
@@ -72,9 +72,12 @@ struct SafeMetadata {
     payload: Option<String>,
 }
 
+/// Needed by [`compute_safe_replica_state`]
 #[derive(Clone, Copy)]
 pub struct ReplicationParameters<'a> {
+    /// Where the index to be replicated is located on disk
     pub path: &'a Path,
+    /// IDs of the segments already present at the replica.
     pub on_replica: &'a [String],
 }
 
--- a/nucliadb_core/src/tantivy_replica.rs
+++ b/nucliadb_core/src/tantivy_replica.rs
@@ -112,12 +112,6 @@ pub fn compute_safe_replica_state(
             segment_id,
         });
 
-        let postings = PathBuf::from(format!("{raw_segment_id}.idx"));
-        let positions = PathBuf::from(format!("{raw_segment_id}.pos"));
-        let terms = PathBuf::from(format!("{raw_segment_id}.term"));
-        let store = PathBuf::from(format!("{raw_segment_id}.store"));
-        let fast_fields = PathBuf::from(format!("{raw_segment_id}.fast"));
-        let field_norms = PathBuf::from(format!("{raw_segment_id}.fieldnorm"));
         let deletes = delete_opstamp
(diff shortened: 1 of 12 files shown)


nucliadb_core/src/tantivy_replica.rs
```diff
@@ -0,0 +1,238 @@
...
+            path: workspace.path(),
+            on_replica: &[],
+        };
+        let metadata_as_json = serde_json::to_value(index.load_metas().unwrap()).unwrap();
+        let safe_replica_state = compute_safe_replica_state(replica_params, &index).unwrap();
+        assert_eq!(safe_replica_state.index_metadata, metadata_as_json);
+    }
+
+    #[test]
+    fn recreate_using_safe_state() {
```
javitonino, line 205: Cool test :heart:

nucliadb_node/src/replication/service.rs
```diff
@@ -117,6 +118,67 @@ async fn stream_data(
     Ok(())
 }
 
+async fn replicate_from_raw(
+    replica_state: RawReplicaState,
+    chunk_size: u64,
+    shard_path: &Path,
+    generation_id: &str,
+    _index_prefix: PathBuf,
```
javitonino, line 126: I find it a bit confusing that tantivy returns paths relative to the index and raw (vectors) returns paths relative to the shard. Would it be possible to change vectors so that it's also using paths relative to the index?

Feel free to say no, I saw something regarding `vectors` and `vectorset` which looks a bit tricky.

In any case, it would be nice to keep this in mind for `vectors2`

nucliadb_core/src/tantivy_replica.rs
```diff
@@ -0,0 +1,241 @@
...
+            max_doc,
+            segment_id,
+        });
+
+        let postings = PathBuf::from(format!("{raw_segment_id}.idx"));
+        let positions = PathBuf::from(format!("{raw_segment_id}.pos"));
+        let terms = PathBuf::from(format!("{raw_segment_id}.term"));
+        let store = PathBuf::from(format!("{raw_segment_id}.store"));
+        let fast_fields = PathBuf::from(format!("{raw_segment_id}.fast"));
+        let field_norms = PathBuf::from(format!("{raw_segment_id}.fieldnorm"));
```
javitonino, line 120: Maybe all of this could be pushed to after the `continue`, since it's not used in the case where the segment is on the replica.

javitonino (approved):
Left a few suggestions, nothing blocking.

As a general comment, this ties us a bit more onto the specific tantivy version, but it seems pretty much inevitable.

nucliadb_node/src/replication/service.rs
hermeGarcia: You are totally right, the only reason why I did not unify the approach was to not overload the PR. I was trying to add the tantivy thing while changing the surroundings as little as possible. If you want I can unify in the same PR!

nucliadb_core/src/tantivy_replica.rs
hermeGarcia: Yep, sorry OCD kicked in 😂

PR #1767, opened 2024-01-24
Offer different strategies for building RAG prompt context

lferran:
### Description

This PR adds a new feature to the chat endpoint that allows the users to choose different strategies for crafting the LLM context.

Right now, 2 strategies have been implemented:
- `full_resource`: if selected, the LLM context is done by pulling the complete extracted text of all fields of the matching resources.
- `extend_with_fields`: if selected, must specify a list of fields (e.g: `a/title`) that will be fetched from the matching paragraph's resource and included in the context.

I also added a way to "cap" the size of the context sent to the LLM

### How was this PR tested?
Unit and integration tests


nucliadb_utils/nucliadb_utils/concurrency.py
```diff
@@ -0,0 +1,78 @@
...
+from nucliadb_utils import logger
+
+
+class ConcurrentRunner:
+    """
+    Runs a list of coroutines concurrently, with a maximum number of tasks running.
+    Returns the results of the coroutines in the order they were scheduled.
+    """
+
+    def __init__(self, max_tasks: Optional[asyncio.Semaphore] = None):
```
jotare, line 33: NIT: I'd prefer an `int` here to hide the `asyncio.Semaphore` implementation detail

edit: you're making this in `run_concurrently`, so feel free to ignore this

nucliadb_models/nucliadb_models/search.py
```diff
@@ -776,6 +817,11 @@ class ChatRequest(BaseModel):
     security: Optional[
         RequestSecurity
     ] = SearchParamDefaults.security.to_pydantic_field()
+    rag: Optional[RAGOptions] = Field(
+        default=None,
+        title="RAG context options",
+        description="Options for tweaking how thecontext for the LLM model is crafted",
```
jotare, line 823:
```suggestion
        description="Options for tweaking how the context for the LLM model is crafted",
```

nucliadb/nucliadb/search/search/chat/prompt.py
```diff
@@ -154,3 +162,207 @@ async def get_chat_prompt_context(
...
+
+    async def _build_context(self):
+        if self.options is None or self.options.context_strategies is None:
+            return await default_prompt_context(self.kbid, self.find_results)
+
+        chosen_strategies = self.options.context_strategies
+        if ContextStrategy.FULL_RESOURCE in chosen_strategies:
+            return await full_resource_prompt_context(self.kbid, self.find_results)
+
+        return await composed_prompt_context(
```
jotare, line 341: Feels kind of weird how `extend_with_fields` implicitly activates the strategy even though could not be in your chosen strategies. Maybe a different structure in `RagOption` or to explicitly check for the strategy will solve it

nucliadb_models/nucliadb_models/search.py
```diff
@@ -776,6 +817,11 @@ class ChatRequest(BaseModel):
     security: Optional[
         RequestSecurity
     ] = SearchParamDefaults.security.to_pydantic_field()
+    rag: Optional[RAGOptions] = Field(
```
vangheem, line 820: This config feels a little complex. Is there a way to do it with less custom validation and less required None checks?

Maybe something like this?

```
class RagStrategy(pydantic.BaseModel):
    name: str

class SimpleStrategy(RagStrategy)
    name: str = Field("simple", const=True)

class ExtendeFieldStrategy(RagStrategy)
    name: str = Field("extended_fields", const=True)
    fields: list[str]

class ChatRequest(BaseModel):
    ...
    rag_strategies: list[RagStrategy] = [SimpleStrategy] # might need union here
```

nucliadb_utils/nucliadb_utils/concurrency.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
vangheem, line 1: Maybe rename to `asyncio_utils`?

nucliadb_models/nucliadb_models/search.py
```diff
@@ -728,6 +728,47 @@ class AskDocumentModel(BaseModel):
     user_id: str = Field(description="The id of the user associated to the request")
 
 
+class ContextStrategy(str, Enum):
+    FULL_RESOURCE = "full_resource"
+    EXTEND_WITH_FIELDS = "extend_with_fields"
```
vangheem, line 733: Should the conversation auto expansion that we currently do also be considered a context strategy with configuration on how many messages we inject?

Maybe not in this PR I guess though

nucliadb_models/nucliadb_models/search.py
lferran: hm, I like it more! I will try, but IIRC fastapi doesn't know how to properly choose which model of the union has to deserialize to

lferran: nice! I made it work, thanks for the tip!

nucliadb_models/nucliadb_models/search.py
lferran: I like the idea, we can do in a separate PR

vangheem (approved):

PR #1768, opened 2024-01-24
Prefer read replicas when choosing nodes to search

jotare:
### Description
Change logic around `choose_node` so read replicas are preferred over primary index nodes

### Potential improvements
With this logic, if we call `choose_node(shard, read_only=True)` and there's 1 read replica and 2 primary nodes, searches will always go to the secondary. Maybe a better load balancing can be thought.

### How was this PR tested?
Unit tests

### DIFF
--- a/nucliadb/nucliadb/common/cluster/manager.py
+++ b/nucliadb/nucliadb/common/cluster/manager.py
@@ -515,14 +515,14 @@ def choose_node(
 ) -> tuple[AbstractIndexNode, str, str]:
     """Choose an arbitrary node storing `shard` following these rules:
     - nodes containing a shard replica from `target_replicas` are the preferred
-    - if `read_only`, read_replicas are preferred over primaries
+    - when enabled, read replica nodes are preferred over primaries
     - if there's more than one option with the same score, a random choice will
-      be made
+      be made between them.
 
-    According to these rules and considering `read_only == True`, a secondary
-    node containing a shard replica from `target_replicas` is the most
-    preferent, while a primary node with a shard not in `target_replicas` is the
-    least preferent.
+    According to these rules and considering we use read replica nodes, a read
+    replica node containing a shard replica from `target_shard_replicas` is the
+    most preferent, while a primary node with a shard not in
+    `target_shard_replicas` is the least preferent.
 
     """
     target_shard_replicas = target_shard_replicas or []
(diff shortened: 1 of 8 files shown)


nucliadb/nucliadb/common/cluster/manager.py
```diff
@@ -480,45 +480,72 @@ async def add_resource(
...
+
+    According to these rules and considering `read_only == True`, a secondary
+    node containing a shard replica from `target_replicas` is the most
+    preferent, while a primary node with a shard not in `target_replicas` is the
+    least preferent.
+
+    """
+    target_replicas = target_replicas or []
+
+    shard_nodes = get_all_shard_nodes(shard, use_read_replicas=read_only)
```
lferran, line 528: Can we tweak the naming a bit here so that we don't get confused by shard replicas Vs nore read replicas?

Proposal: 
- use `use_secondaries`
- use `shard_replica_id` instead of `replica_id`

lferran (approved):
Looks great! Just left one comment!

Would be nice to catch errors when querying a secondary and fallback to a primary. But maybe in a separate PR

nucliadb/nucliadb/common/cluster/manager.py
jotare: I used *read replica* because I think it's the most used term but as you may have noticed, we are using *read replica* and *secondaries* in multiple places. IMO we should define which one do we prefer and be consistent. I don't mind which one.

I can change all mentions to `replica` for `shard_replica` though

nucliadb/nucliadb/common/cluster/manager.py
```diff
@@ -480,46 +480,75 @@ async def add_resource(
...
+
+def choose_node(
+    shard: writer_pb2.ShardObject,
+    *,
+    target_shard_replicas: Optional[list[str]] = None,
+    use_read_replica_nodes: bool = False,
+) -> tuple[AbstractIndexNode, str, str]:
+    """Choose an arbitrary node storing `shard` following these rules:
+    - nodes containing a shard replica from `target_replicas` are the preferred
+    - if `read_only`, read_replicas are preferred over primaries
```
lferran, line 518: forgot to update the docstring

lferran (approved):

nucliadb/nucliadb/common/cluster/manager.py
jotare: True, thanks!

PR #1776, opened 2024-01-26
More validation for rag strategies

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/search/search/chat/prompt.py
+++ b/nucliadb/nucliadb/search/search/chat/prompt.py
@@ -210,7 +210,11 @@ async def get_resource_field_extracted_text(
     if resource is None:
         return None
 
-    field_type, field_key = field_id.split("/")
+    try:
+        field_type, field_key = field_id.strip("/").split("/")
+    except ValueError:
+        logger.error(f"Invalid field id: {field_id}. Skipping getting extracted text.")
+        return None
     field = await resource.get_field(field_key, KB_REVERSE[field_type], load=False)
     if field is None:
         return None
--- a/nucliadb_models/nucliadb_models/search.py
+++ b/nucliadb_models/nucliadb_models/search.py
@@ -737,14 +737,52 @@ class RagStrategy(BaseModel):
     name: str
 
 
+ALLOWED_FIELD_TYPES: dict[str, str] = {
+    "l": "layout",
+    "t": "text",
+    "f": "file",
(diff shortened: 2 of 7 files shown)


nucliadb/nucliadb/search/search/chat/prompt.py
```diff
@@ -210,7 +210,11 @@ async def get_resource_field_extracted_text(
     if resource is None:
         return None
 
-    field_type, field_key = field_id.split("/")
+    try:
+        field_type, field_key = field_id.strip("/").split("/")
```
javitonino, line 214: We added `.strip("/")` here to remove leading/trailing `/`. Is it really needed? It looks like it makes the validation less strict, which we may want or not.

If this is the case we may want to have a test to ensure leading/trailing `/` are ignored.

nucliadb_models/nucliadb_models/search.py
```diff
@@ -737,15 +737,53 @@ class RagStrategy(BaseModel):
...
 
+    @root_validator(pre=True)
+    def fields_validator(cls, values):
+        if values.get("fields") is None:
+            return values
+
+        # Check that the fields are in the format {field_type}/{field_name}
+        for field in values.get("fields"):
+            try:
+                field_type, _ = field.strip("/").split("/")
```
javitonino, line 769: Same as above

nucliadb_models/nucliadb_models/search.py
```diff
@@ -737,15 +737,53 @@ class RagStrategy(BaseModel):
...
+            except ValueError:
+                raise ValueError(
+                    f"Field '{field}' is not in the format {{field_type}}/{{field_name}}"
+                )
+            if field_type not in ALLOWED_FIELD_TYPES:
+                allowed_field_types_part = ", ".join(
+                    [
+                        f"'{fid}' for '{fname}' fields"
```
javitonino, line 777: Nice detailed error message :sparkling_heart:

javitonino (approved):
Left a question, not sure if `.strip()` is right. If it is, maybe add a test? And then :rocket:

nucliadb/nucliadb/search/search/chat/prompt.py
lferran: Yes! I added the strip so that trailing/leading / were ignored. Will add the test!!

PR #1781, opened 2024-01-29
Install script also installs an upgrade script

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/scripts/install-vm.sh
+++ b/scripts/install-vm.sh
@@ -105,8 +105,18 @@ if (($EUID != 0)); then
   SUDO='sudo'
 fi
 
+PACKAGE=nucliadb
+if [ -n "$1" ]
+then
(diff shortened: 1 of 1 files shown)


scripts/install-vm.sh
```diff
@@ -88,4 +88,36 @@ EOF
...
+set -e
+INSTALL_DIR=/opt/nucliadb
+
+SUDO=''
+if (($EUID != 0)); then
+  SUDO='sudo'
+fi
+
+$SUDO systemctl stop nucliadb
+$SUDO $INSTALL_DIR/bin/pip install --upgrade nucliadb
```
lferran, line 109: WDYT about supporting specific version to upgrade to?
Otherwise this will pull latest, and we may not want this in all cases.

PR #1786, opened 2024-01-30
Do not do substitution when installing upgrade script

javitonino:
### Description
Avoid command substitution in install script.
Also remove duplicated config file installation

### How was this PR tested?

### DIFF
--- a/scripts/install-vm.sh
+++ b/scripts/install-vm.sh
@@ -91,7 +91,7 @@ $SUDO mkdir /var/lib/nucliadb
 $SUDO chown nucliadb:nucliadb /var/lib/nucliadb
 
 # Install upgrade script
-cat << EOF | $SUDO tee /usr/bin/upgrade-nucliadb > /dev/null
+cat << 'EOF' | $SUDO tee /usr/bin/upgrade-nucliadb > /dev/null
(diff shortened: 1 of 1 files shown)


scripts/install-vm.sh
```diff
@@ -91,7 +91,7 @@ $SUDO mkdir /var/lib/nucliadb
 $SUDO chown nucliadb:nucliadb /var/lib/nucliadb
 
 # Install upgrade script
-cat << EOF | $SUDO tee /usr/bin/upgrade-nucliadb > /dev/null
+cat << 'EOF' | $SUDO tee /usr/bin/upgrade-nucliadb > /dev/null
```
javitonino, line 94: The quotes here tell bash to avoid variable substitution in the following script.

scripts/install-vm.sh
```diff
@@ -121,13 +121,3 @@ $SUDO systemctl start nucliadb
 EOF
 
 $SUDO chmod +x /usr/bin/upgrade-nucliadb
-
-# Install config
```
javitonino, line 125: This was duplicated :facepalm:

jotare (approved):

PR #1788, opened 2024-01-30
use v2 processing for e2e test

vangheem:
### Description

### How was this PR tested?


.github/workflows/e2e.yml
```diff
@@ -156,4 +156,6 @@ jobs:
           DEBUG: "1"
           NUA_API_KEY: ${{ secrets.NUA_API_KEY }}
           NUCLIADB_DISABLE_ANALYTICS: "True"
+          ENVIRONMENT: "stage"
+          FLAG_SETTINGS_URL: https://raw.githubusercontent.com/nuclia/status/main/features-v2.json
```
lferran, line 160: nice! I wonder if we should be setting this as default too, so we can have a little bit more control of what happens in onprem deployments.

lferran (approved):

PR #1801, opened 2024-02-05
Support for more complex filtering expressions

lferran:
### Description

### How was this PR tested?


nucliadb_models/nucliadb_models/search.py
```diff
@@ -799,7 +815,11 @@ class FullResourceStrategy(RagStrategy):
 class ChatRequest(BaseModel):
     query: str = SearchParamDefaults.chat_query.to_pydantic_field()
     fields: List[str] = SearchParamDefaults.fields.to_pydantic_field()
-    filters: List[str] = SearchParamDefaults.filters.to_pydantic_field()
+    filters: Union[List[str], List[Filter]] = Field(
+        default=[],
+        title="Filters",
+        description="The list of filters to apply. Filtering examples can be found here: https://docs.nuclia.dev/docs/docs/using/search/#filters",  # noqa: E501
```
lferran, line 821: todo: add examples of more complex filtering in the docs

nucliadb_core/src/query_language/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+            Some(BooleanExpression::Operation(BooleanOperation {
+                operator: Operator::And,
+                operands: search_queries,
+            }))
+        },
+    })
+}
+
+#[cfg(test)]
+mod tests {
```
javitonino, line 235: I would like a test or two hitting `translate()`. All the smaller functions are covered, but no test really covers `translate`.

nucliadb_vectors/src/service/reader.rs
```diff
@@ -97,15 +94,24 @@ impl ReaderChild for VectorReaderService {
         let index_slock = self.index.get_slock()?;
 
         let key_filters = request.key_filters.iter().cloned().map(AtomClause::key_prefix);
-        let paragraph_labels = request.paragraph_labels.iter().cloned().map(AtomClause::label);
+        let field_labels = request.field_labels.iter().cloned().map(AtomClause::label);
         let mut formula = Formula::new();
-        request.field_labels.iter().cloned().map(AtomClause::label).for_each(|c| formula.extend(c));
+
+        if request.field_labels.len() > 0 {
```
javitonino, line 100: Obey clippy :P

nucliadb_core/src/query_language/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+        is_nested,
+        has_field_labels,
+        has_paragraph_labels,
+        expression: BooleanExpression::Operation(boolean_operation),
+    })
+}
+
+fn translate_json(query: Json, context: &QueryContext) -> NodeResult<Translated> {
+    let Json::Object(json_object) = query else {
+        #[rustfmt::skip] return Err(node_error!(
```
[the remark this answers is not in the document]
hermeGarcia: In my opinion this way error messages pop out in the code, which makes them easier to spot.  Also, even though the message is not that long I think is easier on the eyes and when making changes to have the message as isolated as possible.
I thought, since we have `rustfmt::skip` why not use it in this cases.

nucliadb_core/src/query_language/mod.rs
```diff
@@ -0,0 +1,442 @@
...
+    let mut search_queries = Vec::new();
+
+    for (root_id, expression) in subexpressions {
+        let as_json_expression = serde_json::json!({ root_id: expression });
+        let translation_report = translate_json(as_json_expression, &context)?;
+        if translation_report.has_field_labels && translation_report.has_paragraph_labels {
+            let splitted = split_mixed(translation_report.expression, &context)?;
+            prefilter_queries.push(splitted.fields_only);
+            search_queries.push(splitted.paragraphs_only);
+        } else if translation_report.is_nested || translation_report.has_field_labels {
```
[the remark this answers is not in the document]
hermeGarcia: It can not happen, but the code should not reflect that in this way. Fixing it

nucliadb_vectors/src/service/query_io.rs
```diff
@@ -0,0 +1,56 @@
...
+fn map_operation(operation: &BooleanOperation) -> Clause {
+    let operator = match operation.operator {
+        Operator::And => BooleanOperator::And,
+        Operator::Or => BooleanOperator::Or,
+    };
+    let mut operands = vec![];
+    for operand in operation.operands.iter() {
+        let operand = map_expression(operand);
+        operands.push(operand);
+    }
```
javitonino, line 45:
```suggestion
    let operands = operation.operands.iter().map(map_expression).collect();
```

I think a map here is better for expressing intent than a loop.

nucliadb_node/src/shards/shard_reader.rs
```diff
@@ -404,17 +407,16 @@ impl ShardReader {
...
         let search_id = uuid::Uuid::new_v4().to_string();
         let span = tracing::Span::current();
-        let pre_filter = query_plan.pre_filter;
         let mut index_queries = query_plan.index_queries;
 
         // Apply pre-filtering to the query plan
-        if let Some(pre_filter) = pre_filter {
-            let pre_filtered = self.text_reader.pre_filter(&pre_filter)?;
-            index_queries.apply_pre_filter(pre_filtered);
+        if let Some(pre_filter) = &query_plan.prefilter {
```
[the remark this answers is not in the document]
hermeGarcia: We are on the same boat here, fixing it 😂

nucliadb_paragraphs/src/query_io.rs
```diff
@@ -0,0 +1,68 @@
+// Copyright (C) 2021 Bosutech XXI S.L.
```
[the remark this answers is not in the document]
hermeGarcia: I thought this conversation could come up. Although now it would be really easy to generalise those modules into one receiving a `Field` instead of an `Schema` I think is too soon to generalise the filtering logic of paragraphs and texts.
This filtering strategy that uses a query language is not mature enough and I feel will be prone to changes soon, so I prefer to wait for a bit before building a general solution.

nucliadb_vectors/src/service/query_io.rs
hermeGarcia: Good catch!

PR #1806, opened 2024-02-06
Relax read replicas readiness probe times

jotare:
Info about modified parameters:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes


jotare:
The original problem was solved, this is no longer needed

PR #1813, opened 2024-02-08
Fix missing await on GCS debug info

jotare:
### Description

### How was this PR tested?


lferran (commented):
do we need some test that didn't catch it? 😛

PR #1825, opened 2024-02-12
Update python dependency versions

javitonino:
### Description

### How was this PR tested?


nucliadb/nucliadb/writer/api/v1/upload.py
```diff
@@ -294,7 +294,8 @@ async def _tus_post(
     await storage_manager.start(dm, path=path, kbid=kbid)
     await dm.save()
 
-    location = f"{request['path']}/{upload_id}"
+    api.url_path_for
+    location = api.url_path_for("Upload information", kbid=kbid, upload_id=upload_id)
```
javitonino, line 298: Starlette 0.35 (required by fastapi 0.109, required to fix dependabot warning) changes the behaviour of `request['path']`, the old behaviour was a bug. https://github.com/encode/starlette/pull/2400

Changing it to use a higher-level API which works with both versions and should always be correct.

As a note, our use of the `Location` header here is questionable, since we return something like `/kb/...` but we expect the client to send a request to `/api/v1/kb/...` which is non-standard. Of course, it works well with our clients, but still, it's a bit unexpected for people not using the SDK.

lferran: careful though, there are several endpoints with the "upload information" name

nucliadb/nucliadb/writer/tests/test_files.py
```diff
@@ -536,6 +535,99 @@ async def test_file_tus_upload_field_by_slug(writer_api, knowledgebox_writer, re
...
+    path = resp.headers["ndb-field"]
+    field = path.split("/")[-1]
+    assert path.split("/")[-3] == resource
+    assert writer.uuid == resource
+    assert writer.basic.icon == "image/jpg"
+    assert writer.basic.title == ""
+    assert writer.files[field].language == "ca"
+    assert writer.files[field].file.size == len(raw_bytes)
+    assert writer.files[field].file.filename == "image.jpg"
+    assert writer.files[field].file.md5 == "7af0916dba8b70e29d99e72941923529"
```
lferran, line 621: can you simplify the test to assert only what is being tested? i.e that the URL returned works

lferran (approved):

nucliadb/nucliadb/writer/api/v1/upload.py
javitonino: Of course, and that's on purpose :smile:

`url_path_for` matches on name and parameters. So this will match the correct endpoint of the three (it gets passed `**request.path_params` although that's not showing here :/.

lferran (approved):

PR #1831, opened 2024-02-13
Fix some global state in our tests

jotare:
### Description

### How was this PR tested?


nucliadb/nucliadb/common/maindb/redis.py
```diff
@@ -187,7 +187,7 @@ async def initialize(self):
 
     async def finalize(self):
         if self.initialized is True:
-            await self.redis.close()
+            await self.redis.aclose()
```
javitonino, line 190: I saw this deprecation warning in the update deps PR and decided not to apply it. This method was added in 5.0.1 where the `.close()` was marked as deprecated. If we want to do this change, we should change the requirement from `redis>=4.3.4` to `redis>=5.0.1`. Since 5.x is still fairly new, I decided to keep the old one for now, and not change this. This means we get a deprecation warning though.

If you use the `requirements.lock.txt` you should get `redis==5.0.0` which didn't produce the deprecation, at least for me.

jotare: Oh okay, I'll remove it from the PR as it was more a side thing

lferran (approved):

PR #1835, opened 2024-02-14
Fix GCS delete bucket status code

jotare:
### Description
DELETE a bucket returns a 204 in GCS, not a 200

### How was this PR tested?


lferran (approved):
nice catch!
Can you make sure we at least get a sentry event if the response status is not any of those that we handle here?

Also, can you check that this is also fixed on the writer tus gcs driver?

PR #1837, opened 2024-02-14
Fix error code parsing for AWS S3 responses

jotare:
### Description
Error status is sometimes a number and sometimes a string with a special code. I've created a simple function to parse it and return a number always. More mappings can be added if needed

### How was this PR tested?


nucliadb_utils/nucliadb_utils/storages/s3.py
```diff
@@ -514,3 +518,25 @@ async def create_bucket(
                 ]
             },
         )
+
+
+def parse_status_code(error: botocore.exceptions.ClientError) -> int:
```
lferran, line 523: would be nice to have some unit testing for this function, wdyt?

jotare: Sure, I'll add them

lferran (approved):

PR #1838, opened 2024-02-14
Purge orphan shards

jotare:
### Description
Orphan shards exists for multiple reasons:
- a node is not available when removing a KB and shard deletion is never retried
- old versions of rollover/migration code left orphan shards

This PR extends purge command to remove indexed shards that doesn't belong to any live KB

### How was this PR tested?
Integration tests


lferran (commented):
WDYT about providing this logic in the purge command but only really delete orphaned shards if a `--shards` argument is passed?

If it is not passed (which would be the default), we simply log which ones were found so we can investigate and then run the command manually with the `--shards` on.

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Purging a knowledgebox can lead to orphan shards if index nodes are not
+    available or fail. It's possible that other procedures in our database left
+    orphan shards too.
+
+    ATENTION!
+    While migrations are running, new shards are indexed and are
+    not yet stored as shards belonging to a KB, so they can be misclassified as
+    orphan. It is highly recommended to don't remove orphan shards that hasn't
+    exist for a long time.
```
lferran, line 178: Maybe you could extend the logic to detect them to check which shards are part of a reindex?

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    await setup_cluster()
+    driver = await setup_driver()
+
+    try:
+        orphan_shards = await detect_orphan_shards(driver)
+        logger.info(f"Orphan shards detect found {len(orphan_shards)} orphans")
+        if running_settings.debug:
+            await report_orphan_shards(orphan_shards, driver)
```
lferran, line 187: why not reporting always?

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    # To avoid detecting a new shard as orphan, query the index first and maindb
+    # afterwards
+    indexed_shards = await _get_indexed_shards()
+    stored_shards = await _get_stored_shards(driver)
+
+    # Log an error in case we found a shard stored but not indexed, this should
+    # never happen as shards are created in the index node and then stored in
+    # maindb
+    not_indexed_shards = stored_shards.keys() - indexed_shards.keys()
```
lferran, line 58: nit: what about `missing_shards`?

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+async def _get_indexed_shards() -> dict[str, ShardLocation]:
+    indexed_shards: dict[str, ShardLocation] = {}
+    available_nodes = manager.get_index_nodes()
+    for node in available_nodes:
+        node_shards = await node.list_shards()
+        for shard_id in node_shards:
+            try:
+                shard_pb = await node.get_shard(shard_id)
```
lferran, line 84: I'm worried about this, as this will bring the shard into the node memory cache right? is it needed?

Maybe we should only do this check for the shards that we detected as orphaned/ missing, which hopefully it is a smaller subset?

nucliadb/nucliadb/purge/orphan_shards.py
javitonino, line 58: You may want to use `.items()` instead of `.keys()` here. That way, it compares that (shard_id, node_id, kb_id) match and it will also catch the case where the maindb and index agree that a shard exist but disagree on which KB it belongs to or in which node it is present.

That shouldn't happen and it means a shard is very broken, but since I think it would be easy to add, it might be worth it. It it gets even a bit complicated, this is probably not worth it since we will learn about this from errors in other parts of the application.

jotare: It's not clear where is missing though. Anyway, I agree `not_indexed_shards` is not a great name

nucliadb/nucliadb/purge/orphan_shards.py
jotare: It's the only supported operation right now that returns the `kbid`. As it's not something you'll run often, I thought it wouldn't be a problem. But I'll change it to do it only for orphan shards.

nucliadb/nucliadb/purge/orphan_shards.py
jotare: If I change what Ferran pointed above, we won't have the kbid for each one and this won't be possible

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,252 @@
...
+
+    await setup_cluster()
+    driver = await setup_driver()
+
+    try:
+        orphan_shards = await detect_orphan_shards(driver)
+        logger.info(f"Orphan shards detect found {len(orphan_shards)} orphans")
+        await report_orphan_shards(orphan_shards, driver)
+        if args.purge:
+            await purge_orphan_shards(driver)
```
lferran, line 240: If `purge=True` we will be running `detect_orphan_shards`.
you could change it to something like:

```
if args.purge:
   await purge_orphan_shards()
else:
   await report_orphan_shards()
```
And each internall calls detect_orphan_shards

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,252 @@
...
+    orphan_shards: dict[str, ShardLocation] = {}
+    unavailable_nodes = set()
+    rollover_dm = RolloverDataManager(driver)
+    for shard_id in orphan_shard_ids:
+        node_id = indexed_shards[shard_id].node_id
+        node = manager.get_index_node(node_id)
+        kbid = UNKNOWN_KB
+
+        if node is None:
+            unavailable_nodes.add(node_id)
```
lferran, line 85: General style/aesthetic nit comment that feel free to ignore:

If you use `break` or `continue` on while/for loops the code is much easier to understand, as there is less nested branching.

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,252 @@
...
+    return stored_shards
+
+
+async def report_orphan_shards(
+    shards: dict[str, ShardLocation],
+    driver: Driver,
+):
+    async with driver.transaction(read_only=True) as txn:
+        for shard_id, location in shards.items():
+            kb_exists = await KnowledgeBox.exist_kb(txn, location.kbid)
```
lferran, line 159: you can skip the exists_kb lookup it the kbid is equal to "unknown"

nucliadb/nucliadb/purge/orphan_shards.py
```diff
@@ -0,0 +1,252 @@
...
+        if node is None:
+            now = datetime.now()
+            last_try = unavailable_nodes.setdefault(location.node_id, now)
+            if (now - last_try).seconds < 10:
+                logger.warning(
+                    "Node not available while purging orphan shards, skipping",
+                    extra={"node_id": location.node_id},
+                )
+                unavailable_nodes[location.node_id] = now
+            continue
```
lferran, line 190: the retry when the node is None logic seems a bit off, isn't it?
You could instead abstract into a `get_available_node` that handles the waiting and retry ?

nucliadb/nucliadb/purge/orphan_shards.py
jotare: I'm not sure I'm following you but this if is meant to skip the gRPC call when node is not available but still add this shard as orphan. Maybe I can do something to write it more clear

nucliadb/nucliadb/purge/orphan_shards.py
jotare: Simplified

lferran (approved):

PR #1842, opened 2024-02-15
New KB creation logic

lferran:
### Description
Tries to unify as much as possible both onprem and hosted scenarios

### How was this PR tested?
Starting a local nucliadb against stashify + unit tests + existing tests


e2e/test_e2e.py
```diff
@@ -193,6 +194,7 @@ def test_learning_config(kbid: str):
     )
     assert resp.status_code == 422
 
+    # Get config
```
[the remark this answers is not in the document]
jotare: (a couple of lines below this)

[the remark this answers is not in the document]
lferran: nop, /config is the internal endpoint we proxy the request to

nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -207,3 +207,7 @@ class NATSConsumerSettings(BaseSettings):
 
 
 nats_consumer_settings = NATSConsumerSettings()
+
+
+def is_onprem_nucliadb() -> bool:
```
jotare, line 212: Should we use `nucliadb_utils.settings.nuclia_settings.onprem` instead?

nucliadb_protos/knowledgebox.proto
```diff
@@ -48,11 +48,14 @@ message KnowledgeBoxNew {
...
     string forceuuid = 3;
-    utils.VectorSimilarity similarity = 4;
-    optional int32 vector_dimension = 5;
-    optional float default_min_score = 6;
     utils.ReleaseChannel release_channel = 7;
     string learning_config = 8;
+
+    /* Deprecated: these are no longer allowed to be set by the user
+       and will be fetched from the learning config api */
+    utils.VectorSimilarity similarity = 4 [deprecated = true];
```
jotare, line 56: Did you forgot to update protos or field deprecation doesn't change any file?

nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -220,41 +220,50 @@ async def NewKnowledgeBox(  # type: ignore
...
-            # Rollback KB that was just created
+            # Rollback learning config for the kbid that was just created
             try:
-                await self.proc.delete_kb(kbid=kbid)
+                if lconfig_created:
+                    await learning_config.delete_configuration(kbid)
             except Exception:
                 logger.warning(
-                    "Could not rollback KB", exc_info=True, extra={"kbid": kbid}
+                    "Could not rollback learning configuration",
```
jotare, line 261: Should we report to sentry or log an error? A failure here will leave orphan configuration on learning side

jotare (approved):

nucliadb_utils/nucliadb_utils/settings.py
lferran: I think this one is safer, as we can't control which env vars the user will set.
For sure if a nua_key is configured, it means we are in onprem.

PR #1848, opened 2024-02-19
Open grpc-stubs test requirement

jotare:
### Description
`grpc-stubs` was pinned in our test-requirements but not in our normal requirements. Upgrading this package allows us to remove most `type: ignore` while importing grpc

### How was this PR tested?
Running mypy

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -13,7 +13,7 @@ requests
 maturin
 pytest-lazy-fixtures
 pytest-mock
-grpc-stubs==1.24.7
+grpc-stubs>=1.24.7
 aioresponses==0.7.5
 asyncpg>=0.27.0
 # required for pytest-docker-fixtures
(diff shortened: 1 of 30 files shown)


test-requirements.txt
```diff
@@ -13,7 +13,7 @@ requests
 maturin
 pytest-lazy-fixtures
 pytest-mock
-grpc-stubs==1.24.7
+grpc-stubs>=1.24.7
```
javitonino, line 16: Maybe match the version of the grpc packages here?

I mean, `>=1.24.7` happens to work because in new envs or if you force the upgrade because it defaults to installing the latest version, but it's technically not correct.

javitonino: Yeah, something like `>=1.44.0`

javitonino (approved):

PR #1860, opened 2024-02-21
Add Merge gRPC method to index node writer

jotare:
### Description

### How was this PR tested?


hermeGarcia (approved):
Looks great!

lferran (commented):
@jotare can you add some tests to make sure all the wiring between layers is correct?
Doesn't need to be a functional test, just a integration one

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -436,6 +436,13 @@ impl ShardWriter {
         }
     }
 
+    #[tracing::instrument(skip_all)]
+    pub fn merge(&self) -> NodeResult<MergeMetrics> {
+        let _lock = self.gc_lock.blocking_lock();
```
javitonino, line 441: Why do you want this? This will prevent GC and replication while force_merging. GC is already prevented by the slock/elock, so taking this lock here basically just blocks replication while merging. Any particular reason why?

nucliadb_vectors/src/data_point_provider/mod.rs
```diff
@@ -254,6 +262,62 @@ impl Index {
...
+        let force_merge_capacity = 100;
+        let mut live_segments: Vec<_> = state.dpid_iter().collect();
+        let mut buffer = Vec::with_capacity(force_merge_capacity);
+
+        while buffer.len() < force_merge_capacity {
+            let Some(journal) = live_segments.pop() else {
+                break;
+            };
+            buffer.push((state.delete_log(journal), journal.id()));
+        }
```
javitonino, line 304: Maybe we should limit by number of nodes per segment?

```suggestion
        let force_merge_capacity = 50_000;
        let mut live_segments: Vec<_> = state.dpid_iter().collect();
        let mut buffer = Vec::with_capacity(force_merge_capacity);
        let mut nodes_to_merge = 0;

        while nodes_to_merge < force_merge_capacity {
            let Some(journal) = live_segments.pop() else {
                break;
            };
            if journal.no_nodes() > force_merge_capacity {
              live_segments.push(journal);
            } else {
              buffer.push((state.delete_log(journal), journal.id()));
              nodes_to_merge += journal.no_nodes();
            }
        }
```

hermeGarcia: Seems like another good metric, definitely useful, however to make the right decision we should get more metrics right? I went with this because testing is easier tbh...

hermeGarcia: If a segment is big enough this is an infinite loop though:
```rust
while nodes_to_merge < force_merge_capacity {
            let Some(journal) = live_segments.pop() else {
                break;
            };
            if journal.no_nodes() > force_merge_capacity {
              live_segments.push(journal);
            } else {
              buffer.push((state.delete_log(journal), journal.id()));
              nodes_to_merge += journal.no_nodes();
            }
}
```

javitonino: yeah, I somehow thought `live_segments` was two separate vecs! :sweat:

hermeGarcia: Threshold added!

javitonino (approved):

lferran (approved):

PR #1863, opened 2024-02-21
Do not scan_keys on getting labelsets

lferran:
### Description

Current implementation of getting labelsets uses tikv scan keys to get the list of labelsets.
Tikv doesn't behave very well when scanning keys.

This PR does:
- materializes the list of labelsets in a separate key in the KV
- makes sure to update the list on additions and deletions
- adds a migration step to materialize labelset lists for existing kbs
- runs backward compatible code to make sure the feature works while the migration hasn't run yet

### How was this PR tested?
Integration tests


nucliadb/nucliadb/common/datamanagers/labels.py
```diff
@@ -41,16 +43,87 @@ async def get_labels(self, kbid: str) -> kb_pb2.Labels:
...
+
+    @classmethod
+    async def _add_to_labelset_ids(
+        cls, kbid: str, txn: Transaction, labelsets: list[str]
+    ) -> None:
+        previous = await cls._get_labelset_ids(kbid, txn)
+        needs_set = False
+        if previous is None:
+            # TODO: Remove this after migration #11
+            needs_set = True
```
vangheem, line 93: This technically means on initial write, we're always doing a scan. I guess this is okay.

nucliadb/nucliadb/common/datamanagers/labels.py
```diff
@@ -68,13 +141,30 @@ async def set_labels(self, kbid: str, labels: kb_pb2.Labels) -> None:
         Set all labels for a knowledge box (may include multiple labelsets)
         """
         async with self.driver.transaction() as txn:
+            labelset_ids = list(labels.labelset.keys())
+            await LabelsDataManager._add_to_labelset_ids(kbid, txn, labelset_ids)
```
vangheem, line 145: What about removal? This is an overwrite right? Correct me if I'm wrong but only adding here might mean we aren't removing things?

vangheem (approved):

nucliadb/nucliadb/common/datamanagers/labels.py
lferran: yes, this will be like this until the migration has run. Afterwards, we will be able to remove the bw/compat code and skip the scan.

nucliadb/nucliadb/common/datamanagers/labels.py
lferran: good catch!

lferran:
[sc-8994]

PR #1867, opened 2024-02-22
Models management missing endpoint

lferran:
### Description
Adds an endpoint to get the kb config schema before the kb is created

### How was this PR tested?
Unit and local tests

### DIFF
--- a/nucliadb/nucliadb/reader/api/v1/learning_config.py
+++ b/nucliadb/nucliadb/reader/api/v1/learning_config.py
@@ -21,9 +21,11 @@ from fastapi import Request
 from fastapi_versioning import version
 
 from nucliadb import learning_config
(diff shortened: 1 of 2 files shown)


nucliadb/nucliadb/reader/api/v1/learning_config.py
```diff
@@ -32,7 +34,7 @@
     name="Download the Knowledege Box model",
     description="Download the trained model or any other generated file as a result of a training task on a Knowledge Box.",  # noqa
     response_model=None,
-    tags=["Training"],
+    tags=["Models"],
```
jotare, line 37: NIT: Can we use `LLM models` or something else to not get mixed up with pydantic/API models?

jotare (approved):

PR #1887, opened 2024-02-26
Better handle resumable uri not found errors on tus patch

lferran:
### Description

### How was this PR tested?


jotare (approved):
What about the other problem we were facing with s3? The root cause was the same but the error was different

PR #1864, opened 2024-02-21
Complete merges in a separate thread

javitonino:
### Description
Describe the proposed changes made in this PR.

### How was this PR tested?
Describe how you tested this PR.

PR #1894, opened 2024-02-27
Global merge scheduler for the index node

jotare:
### Description
Create a new thread in the index writer to automatically merge shards

Thanks to @javitonino for https://github.com/nuclia/nucliadb/pull/1864

### How was this PR tested?
Unit tests


nucliadb_node/src/bin/writer.rs
```diff
@@ -98,6 +95,9 @@ async fn main() -> NodeResult<()> {
     let (shutdown_notifier, shutdown_notified) = get_shutdown_notifier();
     let shard_cache = Arc::new(ShardWriterCache::new(settings.clone()));
 
+    lifecycle::initialize_writer(settings.clone())?;
+    lifecycle::initialize_merger(Arc::clone(&shard_cache), settings.clone())?;
```
javitonino, line 99: I think we also want to initialize the merger on the bindings initialization, don't we?

nucliadb_node/src/bin/writer.rs
```diff
@@ -98,6 +95,9 @@ async fn main() -> NodeResult<()> {
...
+
```
javitonino, line 100: I was expecting to see a hook at the end of the writer (after `wait_for_sigkill`) to cleanly shutdown the merger. Maybe send a message that sets a boolean that makes the thread exit as soon as it finished the current job.

Not critical, since an unfinished merge will eventually be GC'ed, but still, it's good practice to have clean shutdowns.

nucliadb_node/src/merge/work.rs
```diff
@@ -0,0 +1,142 @@
...
+        for (_, queue) in self.queues.iter_mut().sorted_by_key(|x| x.0).rev() {
+            if let Some(item) = queue.pop_front() {
+                return Some(item);
+            }
+        }
+        None
+    }
+
+    pub fn len(&self) -> usize {
+        self.queues.values().fold(0, |acc, queue| acc + queue.len())
```
javitonino, line 74:
```suggestion
        self.queues.values().reduce(|acc, queue| acc + queue.len())
```

Nitpick: I like using reduce when the initializer is the default, I find it a bit more explicit.

nucliadb_node/src/merge/work.rs
```diff
@@ -0,0 +1,142 @@
...
+
+        if !queue.contains(&item) {
+            queue.push_back(item);
+        }
+    }
+
+    /// Removes the greatest item from the queue and returns it, or `None` if
+    /// it's empty
+    pub fn pop(&mut self) -> Option<T> {
+        for (_, queue) in self.queues.iter_mut().sorted_by_key(|x| x.0).rev() {
```
javitonino, line 65: This is just an idea. I'm okay with the current implementation, so feel free to take this or leave it, as you prefer.

Maybe the constant `MERGE_PRIORITIES_COUNT` should be an array with the ordered priorities instead:

`const SORTED_MERGE_PRIORITIES = [High, Low, WhenFree]`.

This can be used here to iterate over that array (instead of sorting the keys in the hashmap each time). And no need to implement comparators for the enum.

You can also use it in the `WorkQueue` initialization (loop over this array to insert the queues onto the map).

nucliadb_node/src/lifecycle.rs
```diff
@@ -46,6 +50,14 @@ pub fn initialize_writer(settings: Settings) -> NodeResult<()> {
     Ok(())
 }
 
+/// Initialize the global merge scheduler. This function must be called if merge
+/// scheduler should run
+pub fn initialize_merger(shard_cache: Arc<ShardWriterCache>, settings: Settings) -> NodeResult<()> {
```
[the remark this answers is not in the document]
jotare: Deleted in another PR. We can rebase to main when https://github.com/nuclia/nucliadb/pull/1898 is merged

nucliadb_node/src/bin/writer.rs
jotare: I added a `Mutex<bool>` in the `MergeScheduler` and stop function. Do you think is okay?

nucliadb_node/src/merge/scheduler.rs
```diff
@@ -0,0 +1,294 @@
...
+
+    /// Runs the scheduler until stopped
+    pub fn run(&self) {
+        let mut shutdown = self.shutdown.lock().expect("Poisoned merger shutdown signal");
+        *shutdown = false;
+        drop(shutdown);
+
+        loop {
+            let merge = self.process();
+            if let Err(error) = merge {
```
javitonino, line 93: `shutdown` should be read somewhere around here to exit the loop if it's set to true.

nucliadb_node/src/bin/writer.rs
javitonino: Yes, `Mutex<bool>` works. An alternative would be to use `AtomicBool`. Both should work well for this.

I left a comment in the scheduler because I see `shutdown` being set, but I don't see it being read to exit the loop.

javitonino (commented):
One last thing and ready to go!

nucliadb_node/src/merge/scheduler.rs
```diff
@@ -0,0 +1,368 @@
...
+use super::work::WorkQueue;
+use crate::merge::{MergePriority, MergeRequest, MergeWaiter};
+
+// Time between scheduler being idle and scheduling all shards for merge.
+//
+// This value also affects shutdown reaction time
+#[cfg(not(test))]
+const SCHEDULE_ALL_SHARDS_DELAY: Duration = Duration::from_secs(10);
+#[cfg(test)]
+const SCHEDULE_ALL_SHARDS_DELAY: Duration = Duration::from_millis(50);
```
hermeGarcia, line 49: Maybe this should be an attribute of the scheduler, to avoid the conditional compilation.
Not important though, just if you think is worth it.

hermeGarcia (commented):
Just one comment for now! Is looking good!

nucliadb_node/src/merge/scheduler.rs
jotare: I can add as a setting and grab it from there too

nucliadb_node/src/grpc/grpc_writer.rs
```diff
@@ -381,6 +381,16 @@ impl NodeWriter for NodeWriterGRPCDriver {
 
     async fn merge(&self, request: Request<ShardId>) -> Result<Response<MergeResponse>, Status> {
         let shard_id = request.into_inner().id;
+        let shard_id_copy = shard_id.clone();
+        let shards_tasks_copy = Arc::clone(&self.shards);
+        let load_shard_task = || obtain_shard(shards_tasks_copy, shard_id_copy);
```
jotare, line 386: Could you add a comment explaining the load shard from cache hack?

nucliadb_vectors/src/data_point_provider/reader.rs
```diff
@@ -165,9 +165,9 @@ impl Reader {
         let state_path = self.path.join(STATE);
         let disk_version = last_modified(&state_path)?;
 
-        if disk_version == self.version {
-            return Ok(());
-        }
+        // if disk_version == self.version {
```
jotare, line 168: Did you forget to remove this?

hermeGarcia: I sure did 🙏

nucliadb_node/src/grpc/grpc_writer.rs
hermeGarcia: true!

nucliadb_node/src/replication/replicator.rs
```diff
@@ -146,7 +146,7 @@ pub async fn replicate_shard(
     // We only do this here because GC will not be done otherwise on a secondary
     // XXX this should be removed once we refactor gc/merging
     let sshard = Arc::clone(&shard); // moved shard for gc
-    tokio::task::spawn_blocking(move || sshard.gc()).await?.expect("GC failed");
+    tokio::task::spawn_blocking(move || sshard.collect_garbage()).await?.expect("GC failed");
```
javitonino, line 149: This will GC. A writer is running so GC will run. However, the writer has not reloaded the state so the GC will delete everything new.

javitonino (approved):

PR #1905, opened 2024-03-04
Parallelize KB migrations 

lferran:
### Description
Right now we're running only one at a time. This is a pain in production as we have very big KBs that have been running a shard rollover migration for weeks now.

### How was this PR tested?
Unit and integration tests


lferran:
[sc-9035]

vangheem (approved):

PR #1910, opened 2024-03-06
Better min score feature

lferran:
### Description
This PR does:

- Keeps backward compatibility with old api params
- Adds new options on the API to support both min score for bm25 and semantic results.

### How was this PR tested?
Unit and integration tests


hermeGarcia (commented):
Rust side looks good!

hermeGarcia (approved):

jotare (approved):

PR #1911, opened 2024-03-06
add css selector to linkfield

Ciniesta:
### Description
adds css selector to linkfield to support new link parsing functionality

### How was this PR tested?


nucliadb_protos/resources.proto
```diff
@@ -472,6 +473,7 @@ message FieldFile {
     string url = 5;
     map<string, string> headers = 6;
     map<string, string> cookies = 7;
+    string css_selector = 8;
```
lferran, line 476: I think this one is not necessary right? AFAIK we're only adding this to the link field

Ciniesta: right! fixed

lferran (approved):

nucliadb_models/nucliadb_models/link.py
```diff
@@ -63,6 +64,7 @@ class LinkField(BaseModel):
     uri: str
     language: Optional[str] = None
     localstorage: Optional[Dict[str, str]] = {}
+    css_selector: Optional[str] = None
```
lferran, line 67: @Ciniesta one last comment: Can you add pydantic field annotations so that the new parameter is documented on the api?

Ciniesta: done :)

PR #1914, opened 2024-03-07
Improve NER performance

lferran:
### Description
Several things have been done to improve performance/scalability of the entities feature, from more to less important:

- When getting the list entity groups, the relations index query was very inefficient. We use the document index now and just use a faceted query.
- Query read replica shards when possible
- Compose results asynchronously
- Do not allow `show_entities=true` on the listing endpoint

### How was this PR tested?
Existing tests and local tests against stage to double check that the documents index returns the same indexed entity group names


nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -66,10 +68,17 @@ def __init__(
         self,
         knowledgebox: KnowledgeBox,
         txn: Transaction,
+        use_read_replica_nodes: bool = False,
     ):
+        """
+        Args:
+        - use_read_replica_nodes: when True, use read replica nodes for querying the relation
+          index. Otherwise, primary nodes are used.
+        """
```
jotare, line 77: NIT: I feel this comment is kind of redundant

nucliadb/nucliadb/ingest/orm/entities.py
```diff
@@ -95,16 +104,28 @@ async def get_entities_groups(self) -> dict[str, EntitiesGroup]:
...
-        async for group in self.iterate_entities_groups_names(exclude_deleted=True):
-            stored = await self.get_stored_entities_group(group)
-            if stored is not None:
-                groups[group] = EntitiesGroupSummary(
-                    title=stored.title, color=stored.color, custom=stored.custom
+        max_simultaneous = asyncio.Semaphore(10)
+
+        async def _composition(group: str):
+            async with max_simultaneous:
+                stored = await self.get_stored_entities_group(group)
```
jotare, line 111: Is this parallelization really needed? The method only gets a value from maindb

jotare (approved):

nucliadb/nucliadb/ingest/orm/entities.py
lferran: it's doing a get for every group! we have 18 groups that learning detects, plus if the user enters their own, this could be a lot of gets in the end. 

IMO it doesn't harm

PR #1925, opened 2024-03-11
Merge metrics

javitonino:
### Description

Adds metrics:
- Merge time
- Merge input segment size (in # of embeddings)
- Merge output segment size (in # of embeddings)

All of those are histograms, and labeled by source (scheduler, on-commit), so we can also get things like:
- Number and speed of merge operations per source
- Total number of merged segments per operation

Also changes to always compile metrics, which simplifies a bit of the metrics code.


nucliadb_vectors2/src/service/writer.rs
```diff
@@ -59,6 +59,19 @@ impl Debug for VectorWriterService {
     }
 }
 
+fn record_merge_metrics(source: MergeSource, data: &crate::data_point_provider::writer::MergeMetrics) {
+    if data.merged == 0 {
```
hermeGarcia, line 63: Maybe knowing how often this happens could help us in some way? To tweak frequencies maybe?

hermeGarcia (approved):
LGTM! Just a comment

PR #1927, opened 2024-03-12
Update shards protobuf to represent multiple writable shards

jotare:
### Description

### How was this PR tested?


nucliadb/nucliadb/common/cluster/manager.py
```diff
@@ -268,9 +271,13 @@ async def create_shard_by_kbid(
 
         # Append the created shard and make `actual` point to it.
         kb_shards.shards.append(shard)
+        # B/c with Shards.actual
```
lferran, line 274: WDYT about adding unit tests to this part of the code? It's a pretty important part of nucliadb

nucliadb/nucliadb/ingest/consumer/shard_creator.py
```diff
@@ -89,9 +90,17 @@ async def handle_message(self, raw_data) -> None:
...
+        async with self.driver.transaction(read_only=True) as txn:
+            kb_shards = await shards_data_manager.get_kb_shards(txn, kbid)
+            current_shard = await self.shard_manager.get_current_active_shard(txn, kbid)
 
+        if kb_shards is None or current_shard is None:
+            msg = "Processing a notification for a nonexistent"
+            logger.error(msg, extra={"kbid": kbid})
+            raise Exception(msg)
+
+        # TODO: when multiple shards are allowed, this should either handle the
```
lferran, line 102: we could start simple for now and keep the same strategy: when the shard that has been written to goes over the max paragraph/fields size, we set it as read_only and create another one.

lferran: we should keep the other ones open though

nucliadb/nucliadb/ingest/consumer/shard_creator.py
```diff
@@ -89,9 +90,17 @@ async def handle_message(self, raw_data) -> None:
...
-        kb_shards = await self.shard_manager.get_shards_by_kbid_inner(kbid)
-        current_shard: writer_pb2.ShardObject = kb_shards.shards[kb_shards.actual]
+        async with self.driver.transaction(read_only=True) as txn:
+            kb_shards = await shards_data_manager.get_kb_shards(txn, kbid)
+            current_shard = await self.shard_manager.get_current_active_shard(txn, kbid)
 
+        if kb_shards is None or current_shard is None:
+            msg = "Processing a notification for a nonexistent"
+            logger.error(msg, extra={"kbid": kbid})
+            raise Exception(msg)
```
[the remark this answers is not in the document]
jotare: Not sure. This situations probably means either the KB has something wrong or it has been deleted. The first case is interesting to raise and exception, the second would mean the notification queue is a bit slow consuming but nothing weird.

Maybe we can ignore it

nucliadb/nucliadb/ingest/consumer/shard_creator.py
jotare: Why would you want to leave the old one open? The behavior right now in this PR should be the same as before (the TODO is just a pointer for the future). Currently, increasing the `actual` field when a new shard is added, the other is automaticall "closed"

nucliadb/nucliadb/common/cluster/manager.py
jotare: Added a unit test for shard creation

nucliadb/nucliadb/ingest/consumer/shard_creator.py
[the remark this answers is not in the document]
lferran: I would say if the KB was deleted, we need to make sure we ignore that message and ack

lferran (approved):

nucliadb/nucliadb/ingest/consumer/shard_creator.py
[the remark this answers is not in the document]
jotare: I'll left the log and changed the `raise` for a `return`

PR #1932, opened 2024-03-13
Faster hnsw merge

javitonino:
### Description

### How was this PR tested?


nucliadb_vectors2/src/data_point/mod.rs
```diff
@@ -202,17 +202,32 @@ where
...
+    if has_deletions {
+        index = RAMHnsw::new();
+        start_node_index = 0;
+    } else {
+        // If there are no deletions, we can reuse the first segment
+        // HNSW since its indexes will match the the ones in data_store
+        index = DiskHnsw::deserialize(&operants[0].1.index);
+        start_node_index = data_store::stored_elements(&operants[0].1.nodes);
+    }
+
```
hermeGarcia, line 226: Just a mental note / something good to have in mind: this only works because `data_store::merge` will maintain the order in which producers were given. Otherwise this would fail.
Would it be cleaner to have `data_store::merge` return the index of the producer that was written first?

hermeGarcia (approved):
Looks good just a comment that I think is worth considering before merging.

nucliadb_vectors2/src/data_point/mod.rs
javitonino: This is currently covered by tests (if you change the order, the tests break, the search does not return the correct results).

With that in mind, do you think it's worth adding this? I think with the current code it would look weird (it would return a hardcoded 0) and it's likely you would forget to check this anyway until the tests fail.

If we want to do this, I would suggest to move the iteration over datapoints from `data_store::merge` to `data_point::merge`. Something like this:

```rust
// In data_point::merge
data_store::reserve_space(nodes_file, total_nodes_count);
let mut has_deletions = false;
let mut pointer = 0; // To keep position in target nodes_file
for op in operants {
  if data_store::copy(nodes_file, op.nodes_file, op.delete_log, &mut pointer) {
    has_deletions = true
  }
}
```

That way, the `data_point::merge` algorithm can ensure control of the order of the data_store.

Does this make sense?

hermeGarcia: Oh I didn't realise the test was going to fail if that were to change. Even though what you suggested I now think how it is plus the test is actually the cleanest way! Thanks for pointing it out

PR #1933, opened 2024-03-13
Vectors replication from state

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/tests/common/node_services.rs
+++ b/nucliadb_node/tests/common/node_services.rs
@@ -277,7 +277,7 @@ impl NodeFixture {
             self.with_secondary_writer().await?;
         }
 
-        let settings = self.settings.clone();
+        let settings = self.secondary_settings.clone();
         let addr = self.secondary_reader_addr;
         let notifier = Arc::clone(&self.shutdown_notifier);
 
--- a/nucliadb_vectors2/src/service/writer.rs
+++ b/nucliadb_vectors2/src/service/writer.rs
@@ -30,14 +30,13 @@ use nucliadb_procs::measure;
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::fs;
-use std::io::Cursor;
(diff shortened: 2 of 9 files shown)


nucliadb_node/tests/common/node_services.rs
```diff
@@ -277,7 +277,7 @@ impl NodeFixture {
             self.with_secondary_writer().await?;
         }
 
-        let settings = self.settings.clone();
+        let settings = self.secondary_settings.clone();
```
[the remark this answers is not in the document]
hermeGarcia: 😱

nucliadb_vectors2/src/service/writer.rs
```diff
@@ -292,25 +292,29 @@ impl VectorWriter for VectorWriterService {
     }
 
     fn get_segment_ids(&self) -> NodeResult<Vec<String>> {
```
hermeGarcia, line 294: I think the approach here and in `get_index_files` is good, but I think is now obvious that this functions should not be part of `VectorWriter` right?
Maybe we should create a `replication` module with those methods as functions whose input is the location of the index in the FS.
What do you think?

hermeGarcia (changes requested):
Just code structure matters, the rest looks good!

nucliadb_vectors2/src/service/writer.rs
javitonino: This makes sense to me. The only reason it's not already done is to minimize changes in a single PR. But it might be worth doing it now.

hermeGarcia: I think the change could fit in this PR, but no problem if you prefer to leave it for another PR!

hermeGarcia (approved):
The PR looks good! We must remind that this is safe even without pining the segments because of the `gc_lock` in `ShardWriter` .

PR #1936, opened 2024-03-13
Fix prefiltering for paragaphs

lferran:
### Description
Prefilter works great when to detect ALL match or NO document match, as we know we can skip the filtering / lookup, correspondingly.

Anything in between, we would rather just run the filter query of the search on the paragraphs index, as otherwise the list of matching fields can be huge and turn into a huge set of OR queries for tantivy (one for each matching resource) -- which is very slow.

### How was this PR tested?
Existing tests cover us. See integration tests test_filters.py


nucliadb_core/src/query_planner.rs
```diff
@@ -175,8 +167,18 @@ pub fn build_query_plan(search_request: SearchRequest) -> NodeResult<QueryPlan>
...
+        (Some(search), Some(prefilter)) => Some(BooleanExpression::Operation(BooleanOperation {
+            operator: query_language::Operator::And,
+            operands: vec![search, prefilter],
+        })),
+    };
+
     let paragraphs_context = ParagraphsContext {
-        filtering_formula: search_query,
+        filtering_formula,
     };
```
hermeGarcia, line 182: This can only be done for paragraphs version 2, version 1 would need to use the prefiltered keys since it.

hermeGarcia (changes requested):

hermeGarcia (approved):

PR #1940, opened 2024-03-13
Filter HNSW during BFS in a single pass

javitonino:
### Description

### How was this PR tested?


nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -129,37 +128,67 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+        k_neighbours: usize,
         filter: NodeFilter<'a, DR>,
-    ) -> Option<(Address, f32)> {
+    ) -> Vec<(Address, f32)> {
         // We just need to perform BFS, the replacement is the closest node to the actual
         // best solution. This algorithm takes a lazy approach to computing the similarity of
         // candidates.
-        let mut visited_nodes = HashSet::new();
-        let mut candidates = VecDeque::from([x]);
+        let mut visited = HashSet::new();
```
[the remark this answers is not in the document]
javitonino: I just did it for funsies :stuck_out_tongue_closed_eyes:

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -29,6 +29,35 @@ use rand::rngs::SmallRng;
...
+    fn get(&self, pos: usize) -> bool {
+        let Some(elem) = self.0.get(pos / 32) else {
+            return false;
+        };
+        let bit = pos % 32;
+        let mask = 1 << bit;
+        elem & mask > 0
+    }
+}
+
```
hermeGarcia, line 60: Cool! Don't you think it would be better to use a crate lile [bit-set](https://crates.io/crates/bit-set) ? To not add extra complexity.

hermeGarcia (changes requested):
Just one thing to discuss!

nucliadb_vectors/src/data_point/ops_hnsw.rs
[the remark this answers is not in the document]
hermeGarcia: you can also use  [FxHashSet](https://docs.rs/rustc-hash/latest/rustc_hash/type.FxHashSet.html) which is not cryptographically secure, so its way faster.

nucliadb_vectors/src/data_point/ops_hnsw.rs
javitonino: Oh nice! Changed it. I wrote this in my last five minutes before leaving and the only thing I saw about bit-set is that not maintained banner :sweat_smile:. So it was faster to just write this. But it seems to work well enough and it's not like something like this needs much maintenance.

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -157,7 +157,9 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
                 // Our best candidate is worse than the min score
                 (Some(Cnx(_, cs)), _) if cs < self.tracker.min_score() => break,
                 // Our best candidate is worse than the worst result
-                (Some(Cnx(_, cs)), Some(Reverse(Cnx(_, ws)))) if cs < ws => break,
+                (Some(Cnx(_, cs)), Some(Reverse(Cnx(_, ws)))) if cs < ws && results.len() == k_neighbours => {
```
hermeGarcia, line 160: Since this is filtered BFS search, we can only stop here if we have all that we need.

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -291,14 +291,13 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
         let Some(entry_point) = hnsw.get_entry_point() else {
             return Neighbours::default();
         };
-        let mut crnt_layer = entry_point.layer;
+
         let mut neighbours = vec![(entry_point.node, 0.)];
-        while crnt_layer != 0 {
+        for crnt_layer in (0..=entry_point.layer).rev() {
```
hermeGarcia, line 296: we were skipping layer 0

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: Sorry  I was wrong, this is actually ok 🙏

nucliadb_vectors/src/data_point/ops_hnsw.rs
javitonino: hah, I thought it was on purpose. Should have asked!

PR #1946, opened 2024-03-13
Use current node image for testing

javitonino:
### Description

Tag the compiled node image with the same tag that is being loaded by the Python tests. Before this change, we were testing against an old node image.

### How was this PR tested?

### DIFF
--- a/nucliadb_node/nucliadb_node/tests/integration/test_indexing.py
+++ b/nucliadb_node/nucliadb_node/tests/integration/test_indexing.py
@@ -270,6 +270,8 @@ class TestConcurrentIndexingFailureRecovery:
             elapsed = datetime.now() - start
             assert elapsed.seconds < 60, "too much time waiting"
 
+        await asyncio.sleep(0.5)
+
         # check all messages have been indexed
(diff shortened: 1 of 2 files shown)


nucliadb_node/nucliadb_node/tests/integration/test_indexing.py
```diff
@@ -270,6 +270,8 @@ async def test(
             elapsed = datetime.now() - start
             assert elapsed.seconds < 60, "too much time waiting"
 
+        await asyncio.sleep(0.5)
```
javitonino, line 273: This test is very flaky, and adding this here helps a lot. Not sure if it defeats the purpose of the test though, so please review with care.

jotare: It doesn't change the purpose of the test, but it's kind of weird that the check after the loop fails if we are exiting the loop.

Totally unrelated, the test should be named `test_indexing_on_faulty_worker` or something more descriptive. I think we forgot to name it

jotare (approved):

PR #1952, opened 2024-03-14
single pass bfs

hermeGarcia:
### Description
Based on the idea shown by in [this PR](https://github.com/nuclia/nucliadb/pull/1940), but with a simpler implementation.

### How was this PR tested?
Local tests.

### DIFF
--- a/nucliadb_vectors/src/data_point/ops_hnsw.rs
+++ b/nucliadb_vectors/src/data_point/ops_hnsw.rs
@@ -130,7 +130,7 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
     }
     fn closest_up_nodes<L: Layer>(
         &'a self,
-        x: Address,
+        entry_points: Vec<Address>,
         query: Address,
         layer: L,
         number_of_results: usize,
@@ -141,8 +141,8 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
         // candidates.
 
         let mut results = Vec::new();
-        let mut visited_nodes = HashSet::new();
-        let mut candidates = VecDeque::from([x]);
+        let mut visited_nodes: HashSet<_> = HashSet::from_iter(entry_points.iter().copied());
+        let mut candidates = VecDeque::from(entry_points);
 
         loop {
             if results.len() == number_of_results {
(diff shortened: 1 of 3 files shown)


nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -129,37 +128,54 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+        mut filter: NodeFilter<'a, DR>,
+    ) -> Vec<(Address, f32)> {
         // We just need to perform BFS, the replacement is the closest node to the actual
         // best solution. This algorithm takes a lazy approach to computing the similarity of
         // candidates.
-        let mut visited_nodes = HashSet::new();
-        let mut candidates = VecDeque::from([x]);
+
+        let mut results = Vec::new();
+        let mut visited_nodes: HashSet<_> = HashSet::from_iter(entry_points.iter().copied());
```
javitonino, line 144: You might want to use a bit-set here. I tried an approach very similar to this as well and it had the same problem where the `HashSet` represented 30% of the runtime. So if this is now taking 2s, switching to a bitfield might lower it to 1.5s

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -129,37 +128,54 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+
+            if candidate_similarity < self.tracker.min_score() {
+                break;
             }
+
+            if filter.is_valid(candidate, candidate_similarity) && filter.passes_formula(candidate) {
+                let candidate_vector = self.tracker.get_vector(candidate);
+                filter.vec_counter.add(candidate_vector);
+                results.push((candidate, candidate_similarity));
+            }
```
javitonino, line 166: You might want to move the `if results.len() == number_of_results` to here? No need to expand the last nodes candidates

javitonino (approved):

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: You are right! This took down our benchmark results to 1.15 secs!! 🤯

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: Done! Thanks!

PR #1963, opened 2024-03-18
Add global migration to purge orphan kbslugs keys from maindb

jotare:
### Description
Remove orphan kbslugs from maindb using a global migration

### How was this PR tested?


nucliadb/migrations/0017_purge_orphan_kbslugs.py
```diff
@@ -0,0 +1,55 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def migrate(context: ExecutionContext) -> None:
+    async with context.kv_driver.transaction() as txn:
+        async for key in txn.keys(KB_SLUGS_BASE, count=-1):
+            slug = key.replace(KB_SLUGS_BASE, "")
+            value = await txn.get(key)
+            if value is None:
```
lferran, line 40: shouldn't we delete the key in this case too?

jotare: Sure!

nucliadb/migrations/0017_purge_orphan_kbslugs.py
```diff
@@ -0,0 +1,55 @@
...
+                logger.error(f"KB with slug ({slug}) but without uuid?")
+                continue
+
+            kbid = value.decode()
+            if not KnowledgeBox.exist_kb(txn, kbid):
```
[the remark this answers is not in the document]
jotare: Thanks! I think I'll add a test just in case

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -119,7 +119,6 @@ async def exist_kb(cls, txn: Transaction, uuid: str) -> bool:
     async def delete_kb(cls, txn: Transaction, slug: str = "", kbid: str = ""):
         # Mark storage to be deleted
         # Mark keys to be deleted
-        logger.info(f"Deleting KB kbid={kbid} slug={slug}")
```
lferran, line 122: I wonder if we should return an error in the event of the passed but not not found?
For if this happens again in the future

jotare: I think we should rethink the entire KB slugs thing

lferran (approved):

PR #1971, opened 2024-03-20
parametric merge

hermeGarcia:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/merge/scheduler.rs
+++ b/nucliadb_node/src/merge/scheduler.rs
@@ -35,15 +35,21 @@ use tokio::sync::oneshot::{Receiver, Sender};
 use crate::{settings::Settings, shards::providers::shard_cache::ShardWriterCache};
 use nucliadb_core::metrics::vectors::MergeSource;
 use nucliadb_core::tracing::warn;
+use nucliadb_core::vectors::MergeContext;
 use nucliadb_core::NodeResult;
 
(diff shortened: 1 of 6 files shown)


nucliadb_node/src/merge/scheduler.rs
```diff
@@ -54,11 +60,19 @@ pub struct MergeScheduler {
...
         Self {
-            work_queue: Mutex::new(WorkQueue::new()),
             shard_cache,
             settings,
+            work_queue: Mutex::new(WorkQueue::new()),
             condvar: Condvar::default(),
             shutdown: AtomicBool::new(false),
+            max_nodes_in_merge: match std::env::var(MAX_NODES_IN_MERGE) {
+                Ok(v) => v.parse().unwrap_or(50_000),
+                Err(_) => 50_000,
```
javitonino, line 70: We might want to set a higher default here, so things start happening right away. Maybe it's enough with 100k? Anyway, we can always tweak via env variable later.

javitonino (approved):

nucliadb_node/src/merge/scheduler.rs
hermeGarcia: I didn't want to increase that limit right away, doing it through the env vars life seemed safer. But if you think is not worth I dont mind changing it.

PR #2021, opened 2024-04-02
Fixes SEARCH-K1

bloodbare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/search/predict.py
+++ b/nucliadb/nucliadb/search/predict.py
@@ -350,14 +350,17 @@ class PredictEngine:
             logger.warning(error)
             raise SendToPredictError(error)
 
+        params = {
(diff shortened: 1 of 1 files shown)


lferran (approved):

nucliadb/nucliadb/search/predict.py
```diff
@@ -350,14 +350,17 @@ async def query(
             logger.warning(error)
             raise SendToPredictError(error)
 
+        params = {
+            "text": sentence,
+            "rephrase": str(rephrase),
```
vangheem, line 355: this can be `None`

bloodbare: I will switch it needs to be False or True and False by default

PR #2026, opened 2024-04-03
Throttle OpenTelemetry logs

javitonino:
### Description

### How was this PR tested?


nucliadb_node/src/telemetry.rs
```diff
@@ -35,6 +38,24 @@ use crate::settings::Settings;
 use crate::utils::ALL_TARGETS;
 
 const TRACE_ID: &str = "trace-id";
+const TELEMETRY_ERROR_INTERVAL: Duration = Duration::from_secs(5);
+
+fn telemetry_error_handler(error: global::Error) {
+    static LAST_ERROR: RwLock<Option<Instant>> = RwLock::new(None);
```
hermeGarcia, line 44: Have you considered `Option<Instant>` instead of just `Instant` and initialising to `Instant::now()`? I know we may need to wait a bit to start reporting the first time, but we avoid the need to match if its the first report every time, I don't know, what do you thing?

hermeGarcia (approved):
Just a comment to consider but nothing important! Looks good to me!

PR #2035, opened 2024-04-04
Fallback to generic json error response for unexpected errors

lferran:
### Description
This prevents showing the full error traceback when there are unexpected errors. Instead it will return a HTTP500 with 
```json
{"detail": "Something went wrong, please contact your administrator."}
```
[sc-9583]

### How was this PR tested?
Integration test


nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -25,7 +25,7 @@
 
 
 class RunningSettings(BaseSettings):
-    debug: bool = True
+    debug: bool = False
```
lferran, line 28: This was the actual culprit. We're using this value to set the `debug` attribute of our fastapi apps, which, if set to True, makes the exception handlers be ignored and the full error traceback is returned

jotare (approved):

PR #2049, opened 2024-04-09
Chat with a resource (without retrieval)

lferran:
### Description
When calling the chat endpoint on a particular resource, if the full_resource strategy is used, we skip the retrieval and simply send the whole extracted text to the LLM

### How was this PR tested?
Integration and local tests


lferran:
[sc-9658]

PR #2061, opened 2024-04-15
Validate that export is compatible when importing

lferran:
### Description
Right now, if you export a KB with semantic model A and import it into a KB with semantic model B, the import process will run smoothly but effectively the KB will not work because the model does not match with the imported data.

This PR aims to validate compatibility at the beginning of an import

### How was this PR tested?
Integration tests


lferran:
[sc-7295]

jotare (approved):

PR #2062, opened 2024-04-15
Do not fail if a field does not contain vectors

lferran:
### Description

### How was this PR tested?


vangheem (approved):

vangheem:
I think you can close this issue. I don't think this is actually happening in practice and was a test scenario failure.

PR #2066, opened 2024-04-16
Hnsw algorithm

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_vectors/src/data_point/mod.rs
+++ b/nucliadb_vectors/src/data_point/mod.rs
@@ -688,3 +688,180 @@ impl OpenDataPoint {
         neighbours.into_iter().map(|(address, dist)| (Neighbour::new(address, &self.nodes, dist))).take(results)
     }
 }
+
+#[cfg(test)]
+mod test {
+    use std::collections::BTreeMap;
+
+    use nucliadb_core::NodeResult;
+    use rand::{rngs::SmallRng, Rng, SeedableRng};
+    use tempfile::tempdir;
+
+    use crate::data_types::vector::{dot_similarity, encode_vector};
+
+    use super::{create, DataPointPin, Elem, NoDLog, SearchParams, Similarity};
+
+    const DIMENSION: usize = 1536;
+
+    fn random_vector(rng: &mut impl Rng) -> Vec<f32> {
+        let v: Vec<f32> = (0..DIMENSION).map(|_| rng.gen_range(-1.0..1.0)).collect();
+        normalize(v)
+    }
+
+    fn normalize(v: Vec<f32>) -> Vec<f32> {
+        let mut modulus = 0.0;
+        for w in &v {
+            modulus += w * w;
+        }
+        modulus = modulus.powf(0.5);
+
+        v.into_iter().map(|w| w / modulus).collect()
+    }
+
+    fn random_nearby_vector(rng: &mut impl Rng, close_to: &[f32], distance: f32) -> Vec<f32> {
+        // Create a random vector of low modulus
+        let fuzz = random_vector(rng);
+        let v = close_to.iter().zip(fuzz.iter()).map(|(v, fuzz)| v + fuzz * distance).collect();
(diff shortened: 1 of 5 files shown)


nucliadb_vectors/src/data_point/mod.rs
```diff
@@ -688,3 +688,180 @@ impl OpenDataPoint {
         neighbours.into_iter().map(|(address, dist)| (Neighbour::new(address, &self.nodes, dist))).take(results)
     }
 }
+
+#[cfg(test)]
+mod test {
```
javitonino, line 693: These tests are too slow to run in debug mode (tens of minutes), they only really work in release mode where they take a few seconds.

They are very synthetic, so not the most reliable, but I found them useful and do not involve shipping big datasets. Maybe they should be moved to the vectors_benchmark crate?

nucliadb_vectors/src/data_point/ops_hnsw.rs
```diff
@@ -121,6 +122,38 @@ impl<'a, DR: DataRetriever> HnswOps<'a, DR> {
...
+                results.push((*x, *sim));
+            } else {
+                discarded.push((*x, *sim));
+            }
+        }
+
+        // keepPrunedConnections: keep some other connections to fill M
+        if results.len() < k_neighbours {
+            discarded.sort_unstable_by_key(|(n, d)| std::cmp::Reverse(Cnx(*n, *d)));
+            results.extend(discarded.into_iter().take(k_neighbours - results.len()));
```
hermeGarcia, line 152: If `discarded` becomes a BinaryHeap of `std::cmp::Reverse(Cnx(*n, *d))` we could replace the if by:
```rust
while results.len() < k_neighbours {
     let Some(std::cmp::Reverse(Cnx(n, d))) = discarded.pop() else {
            return results;
     };
     results.push((n, d));
}
```
which should be better.

nucliadb_vectors/src/data_point/mod.rs
hermeGarcia: yes I think that if that is the case as benchmarks may be better, but is not a strong opinion!

nucliadb_vectors/src/data_point/ops_hnsw.rs
javitonino: I tested this. Actually, it should not have that `Reverse` in there, because this is a max-heap, so `pop()` returns the maximum value, which is already what we want, select the maximum similarity.

I micro-benchmarked both approaches and the vec + sorting is actually faster (by 1.5% or so). I think this makes sense, I expected them to be mostly equivalent, since `heap.pop()` is `O(log(n))` so popping a few elements is `O(m log(n))`. If I increase the efConstruction to 1000, it inverts, and BinaryHeap is faster (since m is smaller than n).

Times:
```
Bench               |  Vec + sort   |  BinaryHeap   |  Delta
Build full HNSW     | 24.608.256 µs | 24.993.731 µs | 98.46 %
micro: per select() |    298.419 ns |    303.408 ns | 98.35 %
```

Anyway, I'll change it to a binary heap since it's less sensitive to changes in parameters, but never underestimate a simple vector with a sort :grimacing:

nucliadb_vectors/src/data_point/mod.rs
javitonino: I moved the full code to benchmarks but left a smaller check here, as a canary in case we mess up recall too badly doing some changes. This way we get at least some warning if we inadvertently break things, but doesn't slow tests too much.

nucliadb_vectors/src/data_point/ops_hnsw.rs
hermeGarcia: Oh no, for sure in terms of efficiency is more or less the same, the input is the list of neighbors which will be small.
I just thought the code would be easier to maintain this way

hermeGarcia (approved):

PR #2068, opened 2024-04-16
[Proposal] Vectorsets contract

lferran:
### Description

My proposal is:
- make the embedding model ids match the vectorset_id.
- Include the vectorset_id in the broker message's ExtractedVectorsWrapper
- Send the list of embeddings models to calculate vectors for the resource at every push request to processing.

### How was this PR tested?


nucliadb_protos/resources.proto
```diff
@@ -201,6 +201,7 @@ message ExtractedVectorsWrapper {
         CloudFile file = 2;
     }
     FieldID field = 3;
+    string vectorset_id = 4;
```
lferran, line 204: Since `ExtractedVectorsWrapper` message is already prepared to be in a cloud file if it's too big, I think it'd be enough if we add the `vectorset_id` to it to be able to distinguish the different embedding models.

nucliadb/nucliadb/ingest/processing.py
```diff
@@ -110,6 +110,12 @@ class PushPayload(BaseModel):
         default_factory=models.PushProcessingOptions
     )
 
+    embedding_models: list[models.Embedding] = Field(
+        title="Embedding models",
+        description="List of embedding models to calculate vectors for this resource",
+        default=[],
+    )
```
lferran, line 117: This list will include all embedding models enabled by the KB at every processing request. This way learning knows which vectors needs to calculate for this resource.

lferran:
> - make the embedding model ids match the vectorset_id.
I think this is not a good idea as we may want to support same matryoshka model with different dimensions on different vectorsets.

PR #2071, opened 2024-04-17
Query shard replicas as primary-secondary

lferran:
### Description
This will reduce the possibility of showing inconsistent results across two consecutive requests, as we'll always be querying the same shard replicas as long as their corresponding node is up. Otherwise we'll fall back to the other replica.

### How was this PR tested?
Existing tests


jotare (approved):

lferran:
[sc-9809]

PR #2073, opened 2024-04-18
Term query set

javitonino:
### Description

### How was this PR tested?


nucliadb_node/state.bincode
hermeGarcia, line 1: should this be in the PR?

nucliadb_paragraphs3/src/reader.rs
```diff
@@ -132,6 +132,7 @@ impl ParagraphReader for ParagraphReaderService {
     #[measure(actor = "paragraphs", metric = "search")]
     #[tracing::instrument(skip_all)]
     fn search(&self, request: &ProtosRequest, context: &ParagraphsContext) -> NodeResult<ProtosResponse> {
+        println!("{request:#?}");
```
hermeGarcia, line 135: Should this be in the PR?

hermeGarcia (changes requested):
Looks good! Just some things that should be removed before merging

nucliadb_vectors/src/formula/mod.rs
```diff
@@ -18,37 +18,41 @@
...
-        match self.kind {
-            AtomKind::KeyPrefix => retriever.get_key(x).starts_with(self.value.as_bytes()),
-            AtomKind::Label => retriever.has_label(x, self.value.as_bytes()),
+        match self {
+            Self::KeyPrefix(value) => retriever.get_key(x).starts_with(value.as_bytes()),
+
+            Self::Label(value) => retriever.has_label(x, value.as_bytes()),
+            Self::KeyPrefixSet(set) => {
+                let key = retriever.get_key(x);
+                let kk = String::from_utf8_lossy(key);
```
hermeGarcia, line 49: maybe this is not the best name..

nucliadb_vectors/src/formula/mod.rs
```diff
@@ -18,37 +18,41 @@
...
+                let parts: Vec<_> = kk.split('/').collect();
+                if parts.len() < 3 {
+                    return false;
+                }
+                set.contains(&parts[0..3].join("/"))
```
hermeGarcia, line 54: Maybe im wrong, but in line `55` you are splitting the whole string by '/' to then take the first three chunks and joining them again by `/`, wouldn't it be better to do this in a single iteration through the key that builds the string you want?

hermeGarcia (changes requested):

nucliadb_vectors/src/formula/mod.rs
javitonino: :laughing:

nucliadb_vectors/src/formula/mod.rs
javitonino: Indeed. I just copied this piece of code from the old code at the paragraphs index. I cleaned that one up, but this one remained. I'll make it better.

nucliadb_node/state.bincode
javitonino: WTF? Not sure how this ended up outside the data directory :confused: Thanks!

nucliadb_paragraphs3/src/reader.rs
javitonino: :fire:

hermeGarcia (approved):

PR #2074, opened 2024-04-18
Vectorsets - CRUD API

lferran:
### Description
- Removed unnecessary ingest layer
- Implemented in a separate top level nucliadb sub-package
- Node part still under feature flag -- not ready yet

### How was this PR tested?
e2e tests with the sdk


lferran:
[sc-9618] [sc-9617] [sc-9613]

nucliadb/nucliadb/ingest/processing.py
```diff
@@ -110,6 +115,12 @@ class PushPayload(BaseModel):
         default_factory=models.PushProcessingOptions
     )
 
+    vectorsets: List[RequestedVectorSets] = Field(
```
lferran, line 118: @nuclia/learning this is a proposal on how nucliadb will tell you which embedding models you need to compute for the resource being processed.

nucliadb_protos/writer.proto
```diff
@@ -123,6 +124,9 @@ message BrokerMessage {
     repeated resources.FieldQuestionAnswerWrapper question_answers = 38;
 
     utils.Security security = 39;
+
+    repeated resources.ExtractedVectorsWrapper field_vectorsets = 40;
```
lferran, line 128: @nuclia/learning this is a proposal for how you can send to nucliadb the vectors computed for different vectorsets.

nucliadb_protos/resources.proto
```diff
@@ -201,6 +201,9 @@ message ExtractedVectorsWrapper {
         CloudFile file = 2;
     }
     FieldID field = 3;
+
+    // Vectorset id. If empty, it will be the default vector index
+    string vectorset_id = 4;
```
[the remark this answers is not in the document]
jotare: I don't think we should have a "main index". If users are able to create and remove any vectorset, a default makes no sense

nucliadb/nucliadb/vectorsets/datamanager.py
```diff
@@ -0,0 +1,104 @@
...
+    *, kbid: str
+) -> AsyncIterator[tuple[AbstractIndexNode, str]]:
+    async with datamanagers.with_transaction(read_only=True) as txn:
+        shards_obj = await datamanagers.cluster.get_kb_shards(txn, kbid=kbid)
+        if shards_obj is None:
+            raise ShardsNotFound(kbid)
+        for shard in shards_obj.shards:
+            for replica in shard.replicas:
+                node = get_index_node_retried(replica.node)
+                yield node, replica.shard.id
```
jotare, line 104: This functionality feels like it should be in `KBShardsManager` (from common/cluster/manager.py)

nucliadb/nucliadb/vectorsets/datamanager.py
```diff
@@ -0,0 +1,104 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
jotare, line 1: Creating a module vectorset with a datamanager feels weird as we already have a datamanagers module

nucliadb/nucliadb/vectorsets/models.py
```diff
@@ -0,0 +1,45 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+
+from pydantic import BaseModel
+
+
+class VectorSetHNSWIndex(BaseModel):
```
jotare, line 25: NIT: HNSW is an implementation detail, I don't think this name should be used outside nucliadb_vectors (as model or as variable name)

nucliadb_protos/resources.proto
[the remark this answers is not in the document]
lferran: I know, but that would require refactoring the existing code and makes it harder to deliver the feature in isolation.
I would prefer keeping it like this and unify it after the fact.

nucliadb/nucliadb/vectorsets/datamanager.py
lferran: 👍🏼 will put it in common/datamanagers.
I did it this way to start to avoid banch conflicts with the matryoshka PR

lferran:
This was just a POC. Seems like the implementation steered into another direction where most of the logic is nuclia internal.

PR #2080, opened 2024-04-22
Cleanup ingest gRPC and id/slugs

jotare:
### Description
- Remove `GetResourceId` ingest call and use datamanagers directly from writer
- Remove `ResourceFieldExists` from ingest gRPC
- Simplify how we call writer endpoints by slug or by resource id. Functions get a rid and endpoints by slug convert it first, to avoid conditional code with optional rid/slug
- KB deletion only by kbid. Deletion by slug is not an option in the REST API and having both in ingest makes no sense

### How was this PR tested?
Existing tests


nucliadb/nucliadb/writer/api/v1/resource.py
```diff
@@ -592,19 +590,37 @@ async def _reindex_resource(
...
     ingest = get_ingest()
-    pbrequest = ResourceIdRequest()
+    pbrequest = ResourceFieldId()
     pbrequest.kbid = kbid
-    pbrequest.slug = slug
+    pbrequest.rid = rid
+
     try:
-        response: ResourceIdResponse = await ingest.GetResourceId(pbrequest)  # type: ignore
+        response: ResourceFieldExistsResponse = await ingest.ResourceFieldExists(pbrequest)  # type: ignore
```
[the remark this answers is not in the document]
jotare: Because this call access ORM too and I preferred not to spread ORM outside ingest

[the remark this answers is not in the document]
lferran: Maybe you won't be able to remove the grpc endpoint, but in this particular case, the grpc endpoint is used to check if the resource exists or not. I think you could do a quick datamanager lookup like getting slug for uuid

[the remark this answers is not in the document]
jotare: I did a bit more refactoring and I achieved to remove the gRPC endpoint :)

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -91,25 +91,14 @@ async def get_config(self) -> Optional[KnowledgeBoxConfig]:
             return self._config
 
     @classmethod
-    async def delete_kb(cls, txn: Transaction, slug: str = "", kbid: str = ""):
+    async def delete_kb(cls, txn: Transaction, kbid: str):
```
lferran, line 94: Can you check that backend doesn't need to delete by slug?

nucliadb/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -91,25 +91,14 @@ async def get_config(self) -> Optional[KnowledgeBoxConfig]:
...
-            kbconfig_bytes = await txn.get(datamanagers.kb.KB_UUID.format(kbid=kbid))
-            if kbconfig_bytes is None:
-                raise datamanagers.exceptions.KnowledgeBoxNotFound()
-            pbconfig = KnowledgeBoxConfig()
-            pbconfig.ParseFromString(kbconfig_bytes)
-            slug = pbconfig.slug
+        kb_config = await datamanagers.kb.get_config(txn, kbid=kbid)
+        if kb_config is None:
+            # consider KB as deleted
+            return
```
lferran, line 100: Before we were raising an exception and now we don't. Is this OK? As long as all the callers of this method are adapted.

jotare: The caller was the ingest gRPC endpoint and it wasn't adapted, it'd make the call fail with some internal error. We could manage the exception if you consider it a better approach

nucliadb/nucliadb/ingest/orm/knowledgebox.py
jotare: Checked, they only use the kbid

nucliadb/nucliadb/ingest/orm/knowledgebox.py
jotare: I see backend has a big try-except for the delete, raising an unknown error if the gRPC call fails. Anyway, I think not raising an exception here is fine, we check KB existence using this maindb key

nucliadb/nucliadb/common/datamanagers/resources.py
```diff
@@ -219,3 +225,29 @@ async def modify_slug(txn: Transaction, *, kbid: str, rid: str, new_slug: str) -
...
+async def set_all_field_ids(
+    txn: Transaction, *, kbid: str, rid: str, allfields: resources_pb2.AllFieldIDs
+):
+    key = KB_RESOURCE_ALL_FIELDS.format(kbid=kbid, uuid=rid)
+    await txn.set(key, allfields.SerializeToString())
+
+
+async def has_field(
+    txn: Transaction, *, kbid: str, rid: str, field_id: resources_pb2.FieldID
+) -> bool:
```
lferran, line 246: Sorry, last comment: It would be nice to have tests for those methods, as they are meant to be used across all other parts of the codebase

lferran (approved):

nucliadb/nucliadb/common/datamanagers/resources.py
jotare: Added a simple test for these functions. Existing tests cover orm resource accessing and updating this key

PR #2094, opened 2024-04-24
Be able to index a resource by storage referece

lferran:
### Description
This way we can have sidecar indexing on more KBs at the same time than now because memory is a problem as we have to load the resources in memory.

### How was this PR tested?


lferran:
[sc-9954]

hermeGarcia (changes requested):
Looks good! Just one thing to look at!

hermeGarcia (approved):

nucliadb_node/src/settings.rs
```diff
@@ -86,16 +91,44 @@ const DEFAULT_ENV: &str = "stage";
 #[derive(Clone)]
 pub struct Settings {
     env: Arc<EnvSettings>,
+    pub object_store: Arc<dyn ObjectStore>,
```
jotare, line 94: I think adding `object_store` into `Settings` pollutes the struct. Maybe a `Context` or `GlobalContext` containing both settings and object storage would be cleaner?

jotare (commented):
LGTM!

nucliadb_node/src/settings.rs
lferran: Agree! In fact, we looked into it with Javi but the changes went far out of the scope of the task at hand.
Feel free to ship a proposal for this!

PR #2100, opened 2024-04-26
Cleanup optionals from metadata file

jotare:
### Description
- Pass release channel to secondaries (it was always passing None)
- Remove `Option` from all fields in shards metadata file. As all shards are in the same migration version, this change is safe

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/lib.rs
+++ b/nucliadb_core/src/lib.rs
@@ -72,6 +72,15 @@ impl From<i32> for Channel {
     }
 }
 
+impl From<Channel> for i32 {
+    fn from(value: Channel) -> Self {
+        match value {
+            Channel::STABLE => 0,
+            Channel::EXPERIMENTAL => 1,
+        }
+    }
+}
+
 #[derive(Debug, Default)]
 pub struct RawReplicaState {
     pub metadata_files: HashMap<String, Vec<u8>>,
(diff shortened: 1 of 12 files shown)


nucliadb_core/src/lib.rs
```diff
@@ -72,6 +72,15 @@ impl From<i32> for Channel {
...
 
+impl From<Channel> for i32 {
+    fn from(value: Channel) -> Self {
+        match value {
+            Channel::STABLE => 0,
+            Channel::EXPERIMENTAL => 1,
+        }
+    }
+}
+
```
hermeGarcia, line 83: If this is to keep compatibility with the channel defined in the protos maybe we should do it in a safer way..
An approach I am thinking is to implement `From<ReleaseChannel>` for this `Channel` and then implement `From<Channel> for i32` using the `ProtosChannel` as intermediary. What do you think? 
Is just because I find concerning having a hardcoded mapping that depends in some library implementation.

hermeGarcia (changes requested):
Great! Only one thing to discuss

nucliadb_core/src/lib.rs
jotare: Changed

hermeGarcia (approved):

PR #2103, opened 2024-04-30
Add shard indexes metadata file

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/shards/indexes.rs
+++ b/nucliadb_node/src/shards/indexes.rs
@@ -17,22 +17,21 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-use std::collections::HashMap;
-use std::io::{BufReader, BufWriter};
-use std::fs::File;
-use std::path::{Path, PathBuf};
-use nucliadb_core::NodeResult;
-use serde::{Serialize, Deserialize};
-use uuid::Uuid;
 use nucliadb_core::node_error;
+use nucliadb_core::NodeResult;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fs::File;
+use std::io::{BufReader, BufWriter};
+use std::path::{Path, PathBuf};
+use uuid::Uuid;
 
 use crate::disk_structure;
-
 
 pub const DEFAULT_VECTOR_INDEX_NAME: &str = "__default__";
 pub const SHARD_INDEXES_FILENAME: &str = "indexes.json";
 
-#[cfg_attr(test, derive(Debug))]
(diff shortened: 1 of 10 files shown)


nucliadb_node/src/shards/indexes.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    // Vectorsets
+
+    #[allow(dead_code)]
+    /// Add a new vectorset to the index and returns it's path
+    pub fn add_vectorset(&mut self, name: String) -> NodeResult<PathBuf> {
+        if name == DEFAULT_VECTOR_INDEX_NAME {
+            return Err(node_error!(format!("Vectorset id {DEFAULT_VECTOR_INDEX_NAME} is reserved for internal use")));
+        }
```
hermeGarcia, line 88: Maybe we should also check if the name is already in use?

nucliadb_node/src/shards/metadata.rs
```diff
@@ -83,9 +83,12 @@ pub struct ShardMetadata {
     shard_path: PathBuf,
     id: String,
     kbid: String,
-    similarity: Similarity,
     channel: Channel,
+
+    // redundant configuration
+    similarity: Similarity,
     normalize_vectors: bool,
```
hermeGarcia, line 90: Shouldn't this be removed?

nucliadb_node/src/shards/indexes.rs
```diff
@@ -0,0 +1,268 @@
...
+}
+
+impl Default for ShardIndexesFile {
+    fn default() -> Self {
+        let mut vectorsets = HashMap::new();
+        vectorsets.insert(DEFAULT_VECTOR_INDEX_NAME.to_string(), disk_structure::VECTORS_DIR.into());
+        Self {
+            texts: disk_structure::TEXTS_DIR.into(),
+            paragraphs: disk_structure::PARAGRAPHS_DIR.into(),
+            vectorsets,
```
hermeGarcia, line 144: if you want you can avoid the insert step by doing:
```rust
...
vectorsets: HashMap::from([(DEFAULT_VECTOR_INDEX_NAME.to_string(), disk_structure::VECTORS_DIR.into())]),
...
```

hermeGarcia (changes requested):
Looks good! Some things to look at!

nucliadb_node/src/shards/metadata.rs
jotare: Removing this affects replication and I preferred to do it separately. As replicator creates the shard, it needs all metadata and if it's not stored in shard metadata anymore, we need a way to ask the subindexes for it's metadata.

hermeGarcia (approved):

PR #2108, opened 2024-05-02
Added limits to pagination

lferran:
### Description

### How was this PR tested?


lferran:
[sc-10003]

nucliadb/nucliadb/search/search/query.py
```diff
@@ -68,6 +68,8 @@
     SortField.MODIFIED,
 ]
 
+MAX_HNWS_RESULTS_ALLOWED = 2000
```
jotare, line 71: NIT: there's a typo on HNSW and as it's an implementation detail on the node, I'd prefer something like this
```suggestion
MAX_VECTOR_RESULTS_ALLOWED = 2000
```

or `.._SEMANTIC_..`

jotare (approved):

PR #2111, opened 2024-05-02
Set resource from storage tweaks

lferran:
### Description
- More resilient env var parsing
- Set shard id on downloaded resource

### How was this PR tested?
Rust integration and unit tests


nucliadb_node/src/settings.rs
```diff
@@ -105,17 +105,15 @@ impl From<EnvSettings> for Settings {
 }
 
 pub fn build_object_store_driver(settings: &EnvSettings) -> Arc<dyn ObjectStore> {
-    eprintln!("File backend: {:?}", settings.file_backend);
+    println!("File backend: {:?}", settings.file_backend);
```
jotare, line 108: Why not an `info` log?

jotare (approved):

PR #2117, opened 2024-05-06
Less locking in ShardWriter

jotare:
### Description

### How was this PR tested?


hermeGarcia (approved):
Cool! But, to avoid the discrepancy, I think we should do the same thing for the reader as soon as possible.

PR #2119, opened 2024-05-07
Bump grpcio and protobuf stack

lferran:
### Description
Our current version reqs are giving pip resolution issues

### How was this PR tested?


lferran:
We can't bump to protobuf > 5 because otel libraries still don't support it...

PR #2120, opened 2024-05-07
Store indexes file by write and rename

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_node/src/shards/indexes.rs
+++ b/nucliadb_node/src/shards/indexes.rs
@@ -22,6 +22,7 @@ use nucliadb_core::NodeResult;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fs::File;
+use std::io::Write;
 use std::io::{BufReader, BufWriter};
 use std::path::{Path, PathBuf};
 use uuid::Uuid;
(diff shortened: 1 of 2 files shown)


nucliadb_node/src/shards/indexes.rs
```diff
@@ -131,8 +132,11 @@ impl ShardIndexesFile {
     }
 
     pub fn store(&self, shard_path: &Path) -> NodeResult<()> {
-        let mut writer = BufWriter::new(File::create(shard_path.join(SHARD_INDEXES_FILENAME))?);
+        let filename = shard_path.join(SHARD_INDEXES_FILENAME);
+        let temp = shard_path.join(TEMP_SHARD_INDEXES_FILENAME);
+        let mut writer = BufWriter::new(File::create(temp.clone())?);
         serde_json::to_writer(&mut writer, &self)?;
+        std::fs::rename(temp, filename)?;
```
javitonino, line 139: I think you need to flush the writer before the rename in order for things to be 100% consistent.

As it is now, between the rename and the BufWritter dropping out of scope (when it flushes automatically), the file can be empty.

javitonino (changes requested):

nucliadb_node/src/shards/indexes.rs
jotare: Oh, thanks! I'll add i

javitonino (approved):

PR #2121, opened 2024-05-07
CI wip

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   # Pre-checks
   format-rust:
-    name: Rust code Format
+    name: "Pre-checks: Rust code Format"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -29,7 +29,7 @@ jobs:
         run: cargo +nightly fmt --check
 
   clippy-rust:
-    name: Clippy lint
+    name: "Pre-checks: Clippy lint"
     runs-on: ubuntu-latest
     permissions: write-all
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,167 @@
...
+      - uses: actions/checkout@v3
+      - run: rustup component add clippy
+      - uses: Swatinem/rust-cache@v2
+      - uses: actions-rs/clippy-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --all-targets
+
+  # Builds
+  build-node-binaries:
```
lferran, line 42: another shortcut we could do is build conditionally iif there are changes in the node code /libs

javitonino: It's hard to fit in this model, since it would mean downloading the node from another cache. And compiling the node is actually faster now than installing all python dependencies, so it basically costs no additional time (they run in parallel).

Also, the alternative to avoid compiling the node is downloading the binaries from the main branch. This is tricky and we already had a problem in the past with it, since current main can be different from the point this pr branched off of main, which is what we would actually want.

I think this is better for simplicity. As a bonus, it avoids needing to maintain that list of which directories belong to the node vs python :sweat_smile: .

lferran: Sounds good. In fact, we ended up in the situation we are right now trying to optimize these kind of things too early.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,170 @@
...
+    - uses: actions-rs/cargo@v1
+      env:
+        RUST_BACKTRACE: 1
+      name: cargo test
+      with:
+        command: test
+        args: --workspace --all-features
+  
+  
+  nucliadb-tests:
```
lferran, line 128: Can you add pre-checks for the python part too?

lferran: if you want on a separate pr ofcourse

lferran (approved):

.github/workflows/ci.yml
javitonino: I'll add them here so it's easier to track what has been implemented in the new pipeline, this way it's the full nucliadb_node.yml

.github/workflows/ci.yml
```diff
@@ -41,6 +41,67 @@ jobs:
...
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Run pre-checks
+        run: |
+          pip install -r code-requirements.txt
+          isort -c --profile black nucliadb 
+          cd nucliadb
+          ruff check nucliadb --config=../ruff.toml
+          black --check .
```
lferran, line 66: nit: would be nice to have these abstracted in a make target in fact

PR #2125, opened 2024-05-07
Add generated files on links

bloodbare:
### Description

### How was this PR tested?


nucliadb_protos/resources.proto
```diff
@@ -184,6 +184,7 @@ message LinkExtractedData {
     string embed = 12;
 
     PageStructure pdf_structure = 13;
+    map<string, CloudFile> file_generated = 14;
```
lferran, line 187: Can you add a comment here indicating that the key is the field id?

PR #2129, opened 2024-05-07
Support vectorsets at shard level

jotare:
### Description

### How was this PR tested?


nucliadb_node/src/grpc/grpc_writer.rs
```diff
@@ -280,16 +280,123 @@ impl NodeWriter for NodeWriterGRPCDriver {
...
+                detail: "Vectorset ID must be provided".to_string(),
+                ..Default::default()
+            }));
+        };
+
+        let shards = Arc::clone(&self.shards);
+        let task = move || {
+            run_with_telemetry(info_span!(parent: &span, "Add a vectorset"), move || {
+                let shard = obtain_shard(shards, shard_id.clone())?;
+                shard.create_vectors_index(NewVectorsIndex {
```
javitonino, line 307: We talked about including the dimension in the create vectorset request. Are we not doing that for any particular reason?

After today's talk we would also need to eventually add more info like the datatype, etc.

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -236,135 +240,196 @@ impl ShardWriter {
...
+        let paragraph_task = || run_with_telemetry(info, paragraph_task);
+
+        let mut vector_tasks = vec![];
+        for (name, path) in indexes.iter_vectors_indexes() {
+            vector_tasks.push(|| {
+                // redefining here allows to get a copy of path without moving
+                // metadata
+                #[allow(clippy::redundant_locals)]
+                let path = path;
+                Some((name, open_vectors_writer(versions.vectors, &path, metadata.id())))
```
javitonino, line 277: An alternative (if you prefer it) is to extract the id from metadata and move that in:
```suggestion
            let id = metadata.id();
            vector_tasks.push(move || Some((name, open_vectors_writer(versions.vectors, &path, id))))
```

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -236,135 +240,196 @@ impl ShardWriter {
...
         };
 
-        let vector_task = || {
-            debug!("Vector service starts set_resource");
-            let result = indexes.vector_writer.set_resource(&resource);
-            debug!("Vector service ends set_resource");
-            result
-        };
+        let mut vector_tasks = vec![];
+        for (_, vector_writer) in indexes.vectors_indexes.iter_mut() {
```
javitonino, line 391: Won't this write the same vector to all vectorsets?

I guess it's still a placeholder until we change the SetResource message?

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -420,59 +502,106 @@ impl ShardWriter {
...
 
     #[tracing::instrument(skip_all)]
     pub fn merge(&self, context: MergeContext) -> NodeResult<MergeMetrics> {
-        let runner = read_rw_lock(&self.indexes).vector_writer.prepare_merge(context.parameters)?;
-        let Some(mut runner) = runner else {
-            return Ok(MergeMetrics {
-                merged: 0,
-                left: 0,
-            });
+        // TODO: return metrics by vectorset, not only the deafult one
```
javitonino, line 548: This is only running a merge on the default index, unless I missed something. I'd change this TODO to indicate that not only it returns default metrics, but also that it's only merging the default index. Or even better, actually merge all indexes.

nucliadb_node/src/grpc/grpc_writer.rs
jotare: No, I probably missed the conversation. I'll add it

jotare: No, I probably missed the conversation. Anyway, changing this would imply changing the new shard request too. Do we want to mix it in this PR?

nucliadb_node/src/shards/shard_writer.rs
jotare: Yes, maybe that's a good opportunity to clean protobuffers and pass a custom struct instead of the whole `Resource` to `nucliadb_vectors`

nucliadb_node/src/shards/shard_writer.rs
jotare: Yes, I know. I didn't know if I wanted to change merge protos too

javitonino (approved):

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -540,30 +540,56 @@ impl ShardWriter {
...
-                    merged: 0,
-                    left: 0,
-                });
-            };
-            runner.run()?
-        };
+            for (name, vectors_index) in indexes.vectors_indexes.iter() {
+                let runner = vectors_index.prepare_merge(context.parameters);
+                if let Ok(Some(mut runner)) = runner {
+                    let result = runner.run();
```
javitonino, line 550: `runner.run()` must be outside of any locks

jotare: Fixed

javitonino (approved):

PR #2131, opened 2024-05-08
Ask endpoint

lferran:
### Description
New version of the chat endpoint, streaming structured chunks of data that allows easier extensibility.

### How was this PR tested?
Local tests for now. 
Will add unit/integration testing once we're OK with the draft

### DIFF
--- a/nucliadb/nucliadb/search/api/v1/resource/ask.py
+++ b/nucliadb/nucliadb/search/api/v1/resource/ask.py
@@ -27,7 +27,7 @@ from nucliadb.common import datamanagers
 from nucliadb.models.responses import HTTPClientError
 from nucliadb.search.api.v1.router import KB_PREFIX, RESOURCE_SLUG_PREFIX, api
 from nucliadb_models.resource import NucliaDBRoles
-from nucliadb_models.search import AskRequest, NucliaDBClientType
+from nucliadb_models.search import AskRequest, NucliaDBClientType, SyncAskResponse
 from nucliadb_utils.authentication import requires
 
 from ..ask import create_ask_response
@@ -39,7 +39,9 @@ from ..ask import create_ask_response
     summary="Ask with a resource (by id)",
     description="Ask questions to a resource",
     tags=["Search"],
-    response_model=None,
+    response_model=SyncAskResponse,
+    # Add this to OpenAPI schema when endpoint is not in beta anymore
+    include_in_schema=False,
 )
 @requires(NucliaDBRoles.READER)
 @version(1)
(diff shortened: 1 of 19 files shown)


nucliadb_models/nucliadb_models/search.py
```diff
@@ -1291,3 +1302,77 @@ def validate_facets(facets):
                 )
         facet = next_facet
     return facets
+
```
lferran, line 1305: Here's the public models for the endpoint and the stream structured chunks

nucliadb/nucliadb/search/predict.py
```diff
@@ -115,6 +116,41 @@ class AnswerStatusCode(str, Enum):
     NO_CONTEXT = "-2"
 
 
+class TextGenerativeResponse(BaseModel):
```
lferran, line 119: these are the models corresponding with the data coming from perdict api

lferran:
[sc-9956]

jotare (commented):
Looks much cleaner! You can go ahead with the PR IMO

nucliadb/nucliadb/search/predict.py
```diff
@@ -304,6 +340,32 @@ async def chat_query(
         ident = resp.headers.get(NUCLIA_LEARNING_ID_HEADER)
         return ident, get_answer_generator(resp)
 
+    @predict_observer.wrap({"type": "chat_v2"})
+    async def chat_query_v2(
```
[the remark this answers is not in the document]
lferran: I think I'll keep it as `chat_ndjson`. Even if it's the same predict endpoint, I prefer to duplicate the most of the method so we can have usage metrics for the old and the new formats, to be sure the old one is not being used when we want to remove its support

nucliadb/nucliadb/search/search/chat/query.py
```diff
@@ -407,10 +409,13 @@ async def maybe_audit_chat(
         audit_pb2.ChatContext(author=message.author, text=message.text)
         for message in chat_history
     ]
+    sorted_query_context = sorted_prompt_context_list(
```
lferran, line 412: note: check if we're actually auditing the context text blocks, or just references to them

nucliadb/nucliadb/search/api/v1/resource/ask.py
```diff
@@ -0,0 +1,143 @@
...
+async def resource_ask_endpoint(
+    request: Request,
+    kbid: str,
+    item: AskRequest,
+    x_ndb_client: NucliaDBClientType,
+    x_nucliadb_user: str,
+    x_forwarded_for: str,
+    x_synchronous: bool,
+    resource_id: Optional[str] = None,
+    resource_slug: Optional[str] = None,
```
jotare, line 117: To avoid id/slug conditionals, I'd prefer you to always expect a resource id and convert the slug to id in the specific function. I fixed other writer methods to follow this pattern

jotare (approved):

PR #2136, opened 2024-05-09
Pdm

javitonino:
# mypy fixes

This PR intends to make mypy more consistent. It turns out mypy was not checking packages installed as editable, depending on how exactly they were installed (PEP-517 or old style). I added the corresponding paths to mypy.ini to ensure it always loads all packages.

It turns out that nucliadb_protos had some type errors. I fixed one of them by removing deprecated fields and the others are decided to ignore since they have no practical impact an they are due to a mypy-protobuf limitation (see comment for more details).

Finally, I updated to the latest version of mypy and fixed a couple of issues that the new version detects that the older one didn't.

# PDM: Dependency management

This is a bit unrelated but the PRs got a bit mixed up and I am too lazy to separate.

PDM is a dependency management tool, similar to Poetry. In our case, it's more convenient because it has better support for workspaces/monorepos and I found it easier to integrate into an existing project. In this case, I'm not changing anything in our published packages so everything should continue to work exactly the same.

The changes will be useful to have a consistent environment between our local machines, CI and the docker images. PDM locks the dependencies (in `pdm.lock`) and can restore an exact copy of an environment simply by running `pdm sync -d` (`-d` for develop, so it install testing and linting tools as well) which will install everything under a virtualenv `nucliadb/.venv`).

I deployed this in the new CI pipelines to give it a try, but it could be extended to all other CI pipelines and building docker images if it works for us.

# Dropping Python 3.8 support

We were already depending on packages that have dropped 3.8 support, so nucliadb doesn't really work in 3.8 anyway. MyPy was configured for 3.9, so we also wrote some syntax of our own which is not 3.9 compatible. It's also unsupported (not updated) and will even lose any security support in a couple of months. I think it's ok to drop it now.

This triggered a bunch of linting as black saw the minimum version changed and it could afford to do some syntax changes that were unsupported in 3.8. That's why you see random indentation changes here and there.

# Significant stuff

Since there's a bit too much noise in the PR, here's what's actually interesting to review:
- `.github/workflows/ci.yml` -> Using PDM in the new CI pipeline
- `.github/workflows` -> Dropping 3.8
- `mypy.ini` -> mypy fixes as described above
- `*/setup.py` -> Dropping 3.8
- `pyproject.toml` -> Configuration for PDM
- `nucliadb/nucliadb/search/requesters/utils.py` -> mypy fixes
- `nucliadb_protos/resources.proto` -> Removing deprecated fields
- `nucliadb_telemetry/nucliadb_telemetry/errors.py` and its test -> mypy fixes for sentry


.github/workflows/nucliadb_dataset.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   pre-checks:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11"]
```
lferran, line 19: I wonder if it's safe to just check the lowest and highest versions we support? To alleviate a bit CI.
In this case: removing 3.10 from the list

lferran (approved):

.github/workflows/nucliadb_dataset.yml
javitonino: Maybe. Honestly, I think running the pre-checks across multiple versions does very little, except maybe the oldest version to check for unsupported old syntax. Ideally we would also run the actual tests across all versions, but that is very slow :sweat_smile:

PR #2142, opened 2024-05-13
Update to Pydantic 2

javitonino:
### Description

### How was this PR tested?


nucliadb/nucliadb/reader/api/models.py
```diff
@@ -55,8 +55,8 @@
 class ResourceField(BaseModel):
     field_type: FieldTypeName
     field_id: str
-    value: ValueType
-    extracted: ExtractedDataType
+    value: ValueType = None
```
lferran, line 58: This should be `Optional[ValueType]` instead no?

nucliadb_models/nucliadb_models/common.py
```diff
@@ -88,15 +88,16 @@ class FieldType(Enum):
...
     content_type: str = "application/octet-stream"
-    payload: Optional[str] = Field(description="Base64 encoded file content")
+    payload: Optional[str] = Field(None, description="Base64 encoded file content")
     md5: Optional[str] = None
     # These are to be used for external files
     uri: Optional[str] = None
     extra_headers: Dict[str, str] = {}
 
-    @root_validator(pre=False)
+    @model_validator(mode="before")
```
lferran, line 99: doesn't `pre=False` equal to `mode="after"`?

nucliadb_models/nucliadb_models/text.py
```diff
@@ -83,9 +83,10 @@ class TextField(BaseModel):
     body: str
     format: TextFormat = TextFormat.PLAIN
 
-    @root_validator(pre=False, skip_on_failure=True)
+    @model_validator(mode="before")
```
lferran, line 86: `mode="after"` ? how is the skip on failure done in pydantic 2?

nucliadb_models/nucliadb_models/utils.py
```diff
@@ -18,49 +18,46 @@
...
-    code = "str.slug"
-    msg_template = (
-        "Invalid slug: '{value}'. Slug must be a string with only "
-        "letters, numbers, underscores, colons and dashes."
-    )
 
+def validate_slug(value, handler, info):
+    try:
+        return handler(value)
+    except:
```
lferran, line 49: It'd be safer if we could catch a more specific exception here

nucliadb_models/nucliadb_models/search.py
```diff
@@ -864,14 +867,14 @@ class ImageRagStrategy(BaseModel):
...
+    fields: Set[str] = Field(
         title="Fields",
         description="List of field ids to extend the context with. It will try to extend the retrieval context with the specified fields in the matching resources. The field ids have to be in the format `{field_type}/{field_name}`, like 'a/title', 'a/summary' for title and summary fields or 't/amend' for a text field named 'amend'.",  # noqa
-        min_items=1,
-        unique_items=True,
+        min_length=1,
     )
 
-    @root_validator(pre=True)
+    @model_validator(mode="before")
```
jotare, line 876: This change is different than the others (here `pre=True` is `mode="before"`)

nucliadb/nucliadb/reader/api/models.py
javitonino: `ValueType` is defined above as `Optional[Union[...]]` so it's already an Optional

nucliadb_models/nucliadb_models/text.py
javitonino: Yeah, it should be mode after. I changed it now. It also means changing the validator, so that's why I didn't change all of them at the first pass and I forgot some, thanks for noticing! :heart:.

`skip_on_failure` is not applicable to `mode=after` since what it does is skip pydantic own validators, which have already run when using mode=after.

lferran:
@javitonino can you modify the VERSION file in this PR to 4.0.0?
This way we make sure this changes are published in a major version of all libraries, and consumers can safely do `<4` in the requirements until they migrate

lferran (approved):

PR #2145, opened 2024-05-14
Sc 9354/idp regional merge nucliadb predict and processing

drf7:
### Description

### How was this PR tested?


nucliadb_utils/nucliadb_utils/audit/stream.py
```diff
@@ -178,6 +196,16 @@ async def report(
         if kb_counter:
             auditrequest.kb_counter.CopyFrom(kb_counter)
 
+            await self.kb_usage_utility.send_kb_usage(
```
sunbit, line 199: send_kb_usage is not async right?

sunbit (approved):

nucliadb_utils/nucliadb_utils/nuclia_usage/utils/kb_usage_report.py
```diff
@@ -0,0 +1,137 @@
...
+                usage.storage.fields = storage.fields
+            if storage.HasField("paragraphs"):
+                usage.storage.paragraphs = storage.paragraphs
+            if storage.HasField("resources"):
+                usage.storage.resources = storage.resources
+
+        self.send(usage)
+
+
+class DummyKbUsageReportUtility(KbUsageReportUtility):
```
lferran, line 124: Is this implementation used somewhere? if not, maybe it should be removed?

since the queue is never consumed, all events will be stored in memory.

nucliadb_utils/nucliadb_utils/nuclia_usage/utils/kb_usage_report.py
```diff
@@ -0,0 +1,137 @@
...
+            with suppress(asyncio.CancelledError, asyncio.exceptions.TimeoutError):
+                await asyncio.wait_for(self.task, timeout=2)
+
+    async def run(self) -> None:
+        while True:
+            message: KbUsage = await self.queue.get()
+            try:
+                await self._send(message)
+            except Exception:
+                logger.exception("Could not send KbUsage message")
```
lferran, line 73: should cancelled errors be catched more especifically to finalize gracefully?

lferran: actually, there should be a bigger try catch handling the cancelled error that includes the `while True` I think.

nucliadb_utils/nucliadb_utils/nuclia_usage/utils/kb_usage_report.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def send_kb_usage(
+        self,
+        service: Service,
+        account_id: Optional[str],
+        kb_id: Optional[str],
+        kb_source: KBSource,
+        processes: list[Process] = [],
+        predicts: list[Predict] = [],
+        searches: list[Search] = [],
```
lferran, line 98: IIRC it's dangerous to have parameter defaulting to empty dicts/lists.
Safer to use 
```
processes: Optional[list[Process]] = None,
```
and then below
```
processes = processes or []
```

nucliadb_utils/nucliadb_utils/audit/stream.py
```diff
@@ -99,9 +113,17 @@ async def initialize(self):
 
         self.js = get_traced_jetstream(self.nc, self.service)
         self.task = asyncio.create_task(self.run())
+
+        self.kb_usage_utility = KbUsageReportUtility(
```
lferran, line 117: Feels weird to have a utility inside another one -- utilities are meant to be singletons in nucliadb.
Proposal: why not calling it `KbUsageReportClient` instead?

lferran: no strong opinion though, feel free to disregard the comment !

nucliadb_utils/nucliadb_utils/nuclia_usage/utils/kb_usage_report.py
lferran: Oh, I see the suppress now! 👍🏼

nucliadb_utils/nucliadb_utils/nuclia_usage/__init__.py
lferran, line 1: Would be nice to have some unit tests of the `nuclia_usage` package, that test the initialize, finalize, task cancellation etc.

nucliadb_utils/nucliadb_utils/nuclia_usage/utils/kb_usage_report.py
drf7: Just for some context this `nuclia_usage` package is published in our private artifactory repository since it is used also by learning, so this `DummyKbUsageReportUtility` is used by learning in this case. 
Since nucliadb doesn't have access to the private artifactory we copied it manually into this nucliadb repository.

But of course we can manually modify some code that is not needed here, I have removed the `DummyKbUsageReportUtility` that was not used here.

nucliadb_utils/nucliadb_utils/nuclia_usage/__init__.py
drf7: I have added a test 👍

nucliadb_utils/nucliadb_utils/audit/stream.py
```diff
@@ -229,6 +283,22 @@ async def search(
         auditrequest.trace_id = get_trace_id()
         await self.send(auditrequest)
 
+        self.kb_usage_utility.send_kb_usage(
+            service=Service.NUCLIA_DB,
+            account_id=None,
+            kb_id=kbid,
+            kb_source=KBSource.HOSTED,
+            # TODO unify AuditRequest client type and Nuclia Usage client type
```
lferran, line 291: should this comment be removed? or are you planning on unifying some time soon?

nucliadb_utils/nucliadb_utils/nuclia_usage/protos/__init__.py
lferran, line 1: if we need to modify the protos later on, how do we do it?
We're missing the .proto file no?

nucliadb_utils/nucliadb_utils/audit/stream.py
```diff
@@ -229,6 +283,22 @@ async def search(
...
+            searches=[
+                Search(
+                    client=ClientTypeKbUsage.Value(ClientType.Name(client_type)),  # type: ignore
+                    type=SearchType.SEARCH,
+                    tokens=2000,
```
lferran, line 296: Out of curiosity: why 2000?
It could be a constant instead

nucliadb_utils/nucliadb_utils/audit/stream.py
drf7: I will unify it soon

nucliadb_utils/nucliadb_utils/nuclia_usage/protos/__init__.py
drf7: The proto file is found in another repo, here: https://github.com/nuclia/backend/blob/master/libraries/nuclia_usage/protos/kb_usage.proto
So if we need to modify the protos later one we should modify it in this other repo and copy it to nucliadb, this is only because nucliadb is a public repository that doesn't have access to our google cloud artifactory

nucliadb_utils/nucliadb_utils/audit/stream.py
drf7: I didn't decide it, probably it will be adjusted later when they know which is the best number of tokens per search, and regarding a constant since it is only used here not sure if it is worth to add a constant, but as soon as this number is reused elsewhere then yes

nucliadb_utils/nucliadb_utils/nuclia_usage/protos/__init__.py
lferran: gotcha! understood!

lferran (approved):

PR #2149, opened 2024-05-15
Delete sentences only if we're updating the vectors -- not always

lferran:
### Description

### How was this PR tested?


lferran:
[sc-10167]

PR #2152, opened 2024-05-16
Add a vectors index resource wrapper to set resources in the proper vectorset

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_core/src/vectors.rs
+++ b/nucliadb_core/src/vectors.rs
@@ -18,22 +18,25 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 //
 
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::path::PathBuf;
 
+use nucliadb_protos::nodereader;
+use nucliadb_protos::utils;
 use uuid::Uuid;
 
 use crate::metrics::vectors::MergeSource;
 use crate::prelude::*;
-use crate::protos::*;
 use crate::query_language::BooleanExpression;
 use crate::Channel;
 use crate::IndexFiles;
+use nucliadb_protos::noderesources;
 
 pub type VectorsReaderPointer = Box<dyn VectorReader>;
 pub type VectorsWriterPointer = Box<dyn VectorWriter>;
-pub type ProtosRequest = VectorSearchRequest;
-pub type ProtosResponse = VectorSearchResponse;
+pub type ProtosRequest = nodereader::VectorSearchRequest;
+pub type ProtosResponse = nodereader::VectorSearchResponse;
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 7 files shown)


nucliadb_core/src/vectors.rs
```diff
@@ -93,9 +96,76 @@ pub trait VectorWriter: std::fmt::Debug + Send + Sync {
...
+                            paragraph.vectorsets_sentences.get(vectorset).map(|vectorset_sentences| ParagraphVectors {
+                                vectors: &vectorset_sentences.sentences,
+                                labels: &paragraph.labels,
+                            })
+                        }
+                    }
+                });
+                (field_id, sentences_iterator)
+            })
+    }
```
javitonino, line 161: Maybe something like this is more legible?

```rust
    pub fn fields(&self) -> impl Iterator<Item = (&String, impl Iterator<Item = ParagraphVectors>)> {
        self.resource.paragraphs.iter().map(|(field_id, paragraphs_wrapper)| {
            let sentences_iterator = paragraphs_wrapper.paragraphs.iter().filter_map(|(_paragraph_id, paragraph)| {
                let sentences = if let Some(vectorset) = &self.vectorset {
                    // indexing a vectorset, we should return only paragraphs from this vectorset.
                    // If vectorset is not found, we'll skip this paragraph
                    paragraph.vectorsets_sentences.get(vectorset).map(|x| &x.sentences)
                } else {
                    // Default vectors index (no vectorset)
                    Some(&paragraph.sentences)
                };
                sentences.map(|s| ParagraphVectors {
                    vectors: s,
                    labels: &paragraph.labels,
                })
            });
            (field_id, sentences_iterator)
        })
    }
```

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -383,11 +383,17 @@ impl ShardWriter {
...
 
         let mut vector_tasks = vec![];
-        for (_, vector_writer) in indexes.vectors_indexes.iter_mut() {
+        for (vectorset, vector_writer) in indexes.vectors_indexes.iter_mut() {
             vector_tasks.push(|| {
                 run_with_telemetry(info_span!(parent: &span, "vector set_resource"), || {
                     debug!("Vector service starts set_resource");
-                    let result = vector_writer.set_resource(&resource);
+                    let vectorset_resource = match vectorset.as_str() {
+                        "" | DEFAULT_VECTORS_INDEX_NAME => nucliadb_core::vectors::ResourceWrapper::from(&resource),
```
javitonino, line 391: Usually we use `.into()` to call the `From` implementation. Also applies to some tests below

```suggestion
                        "" | DEFAULT_VECTORS_INDEX_NAME => (&resource).into(),,
```

javitonino (approved):

PR #2155, opened 2024-05-16
Split shard & vector configuration

javitonino:
### Description

### How was this PR tested?


nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -153,7 +153,7 @@ impl ShardWriter {
     }
 
     #[measure(actor = "shard", metric = "new")]
-    pub fn new(metadata: Arc<ShardMetadata>) -> NodeResult<ShardWriter> {
+    pub fn new(metadata: Arc<ShardMetadata>, vector_config: VectorConfig) -> NodeResult<ShardWriter> {
```
jotare, line 156: Idea (feel free to ignore)
I've always thought it's a bit weird how we create a `ShardMetadata` in the cache and then use it to create a shard. I think it'd be cleaner to create a shard with a custom struct and then, either return a tuple `(ShardWriter, ShardMetadata)` or let the caller get the metadata from the writer. This struct will include this new vector config too.

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -321,18 +320,12 @@ impl ShardWriter {
         })
     }
 
-    pub fn create_vectors_index(&self, new: NewVectorsIndex) -> NodeResult<()> {
+    pub fn create_vectors_index(&self, shard_id: String, name: String, new: VectorConfig) -> NodeResult<()> {
```
jotare, line 323: `shard_id` is redundant, it should be obtained from `self` (or `self.metadata`)

nucliadb_node_binding/src/writer.rs
```diff
@@ -96,16 +96,23 @@ impl NodeWriter {
...
+        let vector_config = if let Some(req_config) = request.config {
+            VectorConfig::try_from(req_config).map_err(|err| IndexNodeException::new_err(err.to_string()))?
+        } else {
+            #[allow(deprecated)]
+            VectorConfig {
+                similarity: request.similarity().into(),
+                normalize_vectors: request.normalize_vectors,
+                ..Default::default()
+            }
+        };
```
jotare, line 114: You could encapsulate this in the TryFrom implementation if you don't want a repeat yourself (I've seen the same code a couple of times)

javitonino: It's mostly the same but the request is different in each case (NewShard, NewVectorSet, Replication), so I'd have to write a common trait for these and it all gets messy.

nucliadb_node/src/shards/shard_writer.rs
javitonino: Yeah, I saw this and I agree. It's also weird that we create the shard inside the WriterCache instead of creating it in the writer and just setting it in the case.

I think they are changes worth doing, but belong to a separate PR.

nucliadb_node/src/shards/shard_writer.rs
javitonino: True! changing! :pray:

jotare (approved):

PR #2165, opened 2024-05-20
Remove broker message from pubsub notifications

lferran:
### Description

This PR moves the logic of calculating field auditing data to the processor to avoid sending the whole message on pubsub notifications.
This was dangerous as broker messages can be up to a few hundred Mb of size.

### How was this PR tested?
Existing tests


nucliadb/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -515,6 +516,19 @@ async def maybe_update_resource_basic(
...
 
+    async def get_extended_audit_data(
+        self, message: writer_pb2.BrokerMessage
+    ) -> writer_pb2.Audit:
+        message_audit = writer_pb2.Audit()
+        message_audit.CopyFrom(message.audit)
+        message_audit.kbid = message.kbid
+        message_audit.uuid = message.uuid
+        message_audit.message_source = message.source
+        message_audit.field_metadata.extend([fi.field for fi in message.field_metadata])
```
jotare, line 527: Isn't the field metadata still too big to send in every audit?

jotare (approved):

nucliadb/nucliadb/ingest/orm/processor/__init__.py
lferran: We're only sending the field ids though

jotare:
[sc-8767]

PR #2177, opened 2024-05-21
Include necessary files in sdist

javitonino:
Include the necessary files to be able to build bindings from the source distribution of the package.


nucliadb_node/Cargo.toml
```diff
@@ -110,9 +110,6 @@ tonic-build = "0.11.0"
 ci-test = []
 release-feature-set = []
 
-[profile.release]
-lto = true
-
```
javitonino, line 115: This was being ignored, so I removed it.
I tried adding it back to `.cargo/config.toml` where it wouldn't be ignored, but it was VERY slow (I got bored after 3 minutes, without it takes 5 seconds or so), so removing it for now.

jotare (approved):

PR #2179, opened 2024-05-21
SimSimd with aligned vectors

javitonino:
### Description

### How was this PR tested?


nucliadb_utils/nucliadb_utils/settings.py
```diff
@@ -218,4 +218,4 @@ class NATSConsumerSettings(BaseSettings):
 
 
 def is_onprem_nucliadb() -> bool:
-    return nuclia_settings.nuclia_service_account is not None
```
lferran, line 221: careful with this change. I know using nuclia_service_account is not explicit, but...

Is onprem set to false by default? If so, Is it set to true on all forms of onprem deployments?
Otherwise -- if it's True by default -- is it set to false via values on all hosted deployments?

nucliadb/nucliadb/tests/integration/test_purge.py
```diff
@@ -205,9 +205,13 @@ async def test_purge_orphan_shard_detection(
     node = random.choice(available_nodes)
     orphan_shard = await node.new_shard(
         kbid="deleted-kb",
-        similarity=utils_pb2.VectorSimilarity.COSINE,
         release_channel=utils_pb2.ReleaseChannel.STABLE,
-        normalize_vectors=False,
+        config=nodewriter_pb2.VectorIndexConfig(
```
lferran, line 209: I know right now we don't have config other than the vector index config, but just in case: should we name the param `vector_index_config`?

nucliadb/nucliadb/common/cluster/rollover.py
```diff
@@ -105,12 +105,17 @@ async def create_rollover_shards(
...
+                        vector_dimension=kb_shards.model.vector_dimension,
+                        normalize_vectors=is_matryoshka,
+                    )
                     try:
                         shard_created = await node.new_shard(
                             kbid,
-                            similarity=kb_shards.similarity,
                             release_channel=kb_shards.release_channel,
-                            normalize_vectors=is_matryoshka,
+                            config=config,
```
lferran, line 118: WDYT about `vector_index_config` instead?

jotare (approved):

PR #2182, opened 2024-05-22
Resources orm datamanager cleanup

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/nucliadb/writer/resource/slug.py
+++ b/nucliadb/nucliadb/writer/resource/slug.py
@@ -18,11 +18,9 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
-from nucliadb.common.maindb.utils import get_driver
-from nucliadb.ingest.orm.knowledgebox import KnowledgeBox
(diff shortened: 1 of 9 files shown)


nucliadb/nucliadb/writer/resource/slug.py
```diff
@@ -18,11 +18,9 @@
...
-from nucliadb.common.maindb.utils import get_driver
-from nucliadb.ingest.orm.knowledgebox import KnowledgeBox
+from nucliadb.common import datamanagers
 
 
 async def resource_slug_exists(kbid: str, slug: str) -> bool:
-    driver = get_driver()
-    async with driver.transaction() as txn:
-        return await KnowledgeBox.resource_slug_exists(txn, kbid, slug)
+    async with datamanagers.with_transaction(read_only=True) as txn:
```
javitonino, line 25: A bit of further cleanup:
This is a function that just calls another function and it's used in a single place... Maybe we could remove this function altogether? Then, we can also delete this file, which just contains this

javitonino (approved):
Just a suggestion

PR #2198, opened 2024-05-28
Improve storage interface

lferran:
### Description
Main change is improving typing of `exists` and `iterate_objects` methods, which were returning arbitrary dictionaries

### How was this PR tested?


jotare (approved):
Looks nicer!
What do you think about renaming the `StorageField` to `StorageObject`. If we start to talk about objects in the methods (like the `iterate_objects`), I feel it's a more appropiate name and won't overlap with the resource field concept we have in nucliadb

PR #2202, opened 2024-05-29
Add ingest gRPC methods to create and delete vectorsets

jotare:
### Description

### How was this PR tested?


nucliadb/nucliadb/ingest/service/writer.py
```diff
@@ -687,6 +692,59 @@ async def generate_buffer(
...
+                normalize_vectors=request.normalize_vectors,
+                vector_type=request.vector_type,
+                vector_dimension=request.vector_dimension,
+            ),
+            matryoshka_dimensions=request.matryoshka_dimensions,
+        )
+        response = NewVectorSetResponse()
+        try:
+            async with self.driver.transaction() as txn:
+                kbobj = KnowledgeBoxORM(txn, self.storage, request.kbid)
```
lferran, line 711: wdyt about not using the old orm pattern and simply use datamanagers here?

jotare: I don't know... I feel datamanagers should be only a layer on top of maindb, not a new ORM. But I also feel we miss the data layer that manages any storage (maindb and object storage). 

So I prefer keep it here and when we have this layer ready, move it. Wdyt?

lferran: Alternatives:
- Why not having the logic here, in the writer? (maybe as a private method)
- And if we ever see the need to reuse the above logic in other parts of the code: why not adding a top level `vectorsets` module inside `nucliadb` that will have this logic?

lferran: I am fine with keeping it in the orm layer aswell, no strong opinion

lferran (approved):

PR #2208, opened 2024-05-31
Replicate TiKV transactions to a journal on psql

javitonino:
### Description

### How was this PR tested?


nucliadb/nucliadb/common/maindb/tikv.py
```diff
@@ -262,21 +264,33 @@ async def count(self, match: str) -> int:
...
         await self.data_layer.abort()
+        if self.replication_tx:
+            await self.replication_tx.abort()
         self.open = False
 
     async def commit(self):
         assert self.open
         await self.data_layer.commit()
+        if self.replication_tx:
+            await self.replication_tx.create_replication_commit(self.replication_ops)
```
lferran, line 292: I would check if self.replication_ops has some element, just in case?

nucliadb/nucliadb/common/maindb/tikv.py
```diff
@@ -262,21 +264,33 @@ async def count(self, match: str) -> int:
...
         self.open = True
+        self.replication_tx = replication_tx
+        self.replication_ops: List[int] = []
 
     async def abort(self):
         if not self.open:
             return
         await self.data_layer.abort()
+        if self.replication_tx:
+            await self.replication_tx.abort()
```
lferran, line 285: maybe replication_ops should be cleared here. to be safer

nucliadb/nucliadb/common/maindb/tikv.py
javitonino: Adding an empty array is safe, but a waste. I'll add the check.

lferran (approved):

PR #2211, opened 2024-06-03
Delete vectors on label change

javitonino:
### Description

### How was this PR tested?


nucliadb/nucliadb/ingest/orm/resource.py
```diff
@@ -358,10 +358,16 @@ async def generate_index_message(self) -> ResourceBrain:
...
+                    replace_field = [
+                        f"{p.start}-{p.end}" for p in field_metadata.metadata.paragraphs
+                    ]
+                else:
+                    replace_field = []
                 brain.apply_field_metadata(
                     field_key,
                     field_metadata,
-                    replace_field=[],
+                    replace_field=replace_field,
```
lferran, line 370: rename to `paragraphs_to_replace` ?
Do we really need this though? AFAIK we only need to know the key prefix up to the field part

javitonino: We actually need it, the node expect a paragraph id. Changing this to work by field makes sense but it's a bit more involved, so I wanted to avoid it in this PR. I'll change the name!

nucliadb/nucliadb/ingest/orm/resource.py
```diff
@@ -327,7 +327,7 @@ async def set_relations(self, payload: list[PBRelation]):
         self.relations = relations
 
     @processor_observer.wrap({"type": "generate_index_message"})
-    async def generate_index_message(self) -> ResourceBrain:
+    async def generate_index_message(self, reindex: bool = False) -> ResourceBrain:
```
lferran, line 330: I wonder if reindex should always be true.
is there any occurrence of it being false?

javitonino: Yes, it's set to false during rollover and rebalance, where new resources are created without the need to delete old data.

lferran (approved):

PR #2218, opened 2024-06-04
Storage cleanup and improvements

lferran:
### Description
- Remove pg storage/tus implementations -- we no longer plan to support it.
- Remove unused methods from tus implementations
- Refactored storage's `download` and `iter_data` method to be more explicit about start and end ranges.
- Add integration tests for tus uploads and range downloads for every supported file backend driver

### How was this PR tested?
Integration tests


.github/workflows/nucliadb_standalone.yml
```diff
@@ -75,8 +75,10 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
+      max-parallel: 2
       matrix:
         python-version: ["3.12"]
+        storage_backend: ["gcs", "s3", "local"]
```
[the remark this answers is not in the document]
javitonino: Coming back, since we are using lazy fixtures, why not just do a single run of this with `TESTING_STORAGE_BACKEND=gcs,s3,local`? I guess it would be too slow since most tests depend on this fixture.

Anyway, feel free to disregard all this, mainly thinking out loud about CI.

nucliadb/tests/standalone/fixtures.py
```diff
@@ -28,7 +28,7 @@
 async def knowledgebox_one(nucliadb_manager):
     kbslug = str(uuid.uuid4())
     data = {"slug": kbslug}
-    resp = await nucliadb_manager.post(f"/{KBS_PREFIX}", json=data)
+    resp = await nucliadb_manager.post(f"/{KBS_PREFIX}", json=data, timeout=None)
```
javitonino, line 31: Random change?

nucliadb/src/nucliadb/reader/api/v1/download.py
```diff
@@ -342,14 +345,16 @@ async def download_api(sf: StorageField, headers: Headers, inline: bool = False)
...
         extra_headers["Content-Length"] = f"{range_size}"
         extra_headers["Content-Range"] = f"bytes {start}-{end}/{file_size}"
-        download_headers["Range"] = range_request
+        range_start = start
+        range_end = end
 
     return StreamingResponse(
-        sf.storage.download(sf.bucket, sf.key, headers=download_headers),  # type: ignore
+        sf.storage.download(
+            sf.bucket, sf.key, range_start=range_start, range_end=range_end
```
javitonino, line 356: I'd be tempted to write a small `Range` class that wraps `range_start` and `range_end` since it's passed along so much. It can also have a few helper functions since some operations are repeated here and there (converting to string, checking if any of start/end is None, etc.).

On the other hand, it might be a big change for not too much gain, and being explicit is also nice. So just leavign this as a suggestion, up to you!

nucliadb_utils/src/nucliadb_utils/storages/gcs.py
```diff
@@ -185,35 +189,37 @@ async def iter_data(self, headers=None):
...
+        self, range_start: Optional[int] = None, range_end: Optional[int] = None
+    ):
+        """
+        Iterate through object data.
+        """
+        assert self.storage.session is not None
 
+        headers = await self.storage.get_access_headers()
+        if range_start is not None or range_end is not None:
+            headers["Range"] = "bytes={start}-{end}".format(
```
javitonino, line 202: You might want to use the same code as the S3 driver, which is more compact and should do the same: `f"bytes={range_start or 0}-{range_end or ''}"`

javitonino (approved):
Looks good! Just left some suggestions here and there.

Thank you for your service, the code looks much better now :tada:

.github/workflows/nucliadb_standalone.yml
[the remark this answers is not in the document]
lferran: I actually agree with you, it's overkill in this case.
Plus I am realizing now that some of the tests are written assuming local storage only, so I would need to refactor them too...

I will parametrize the upload/download test instead 👍🏼

nucliadb/tests/standalone/fixtures.py
lferran: oh yes, this helps debugging. When you put a breakpoint, if timeout is not none eventually the request handling task is cancelled and you get kicked out

nucliadb/src/nucliadb/reader/api/v1/download.py
lferran: I like it!

PR #2229, opened 2024-06-06
Azure blob storage support

lferran:
### Description

### How was this PR tested?


pyproject.toml
```diff
@@ -2,6 +2,9 @@
 name = "nucliadb_workspace"
 requires-python = ">=3.9"
 version = "0.0.0"
+dependencies = [
+    "azure-storage-blob>=12.20.0",
```
javitonino, line 6: This dependency should be put in `nucliadb_utils`. Otherwise it won't work for people that download the package from pypi (these are the dependencies for the workspace, which we use for development but it's not published anywhere).

javitonino (changes requested):
Looking good. Just move the dependency to the right place and we are ready to go

javitonino (approved):

PR #2232, opened 2024-06-07
Mypy typing party

jotare:
### Description
- Remove unused `type: ignore` annotations
- Properly type metrics decorator, which was causing all decorated functions to not be type checked
- (bonus) Add `ruff check --fix` where missing

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/ask.py
+++ b/nucliadb/src/nucliadb/search/search/chat/ask.py
@@ -463,7 +463,7 @@ async def ask(
         ask_request=ask_request,
         find_results=find_results,
         nuclia_learning_id=nuclia_learning_id,
-        predict_answer_stream=predict_answer_stream,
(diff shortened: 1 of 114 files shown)


nucliadb/src/nucliadb/search/search/chat/ask.py
```diff
@@ -463,7 +463,7 @@ async def ask(
         ask_request=ask_request,
         find_results=find_results,
         nuclia_learning_id=nuclia_learning_id,
-        predict_answer_stream=predict_answer_stream,
+        predict_answer_stream=predict_answer_stream,  # type: ignore
```
jotare, line 466: @lferran this was the one I mentioned you. `predict_answer_stream` is an `AsyncIterator[GenerativeChunk]` but it expects an `AsyncGenerator[GenerativeChunk, None]`

lferran (approved):

PR #2233, opened 2024-06-07
Split kb ops by onprem hosted

jotare:
### Description
KnowledgeBox creation logic is more complex than needed. Currently, HTTP methods on writer forward operations to ingest gRPC. This gRPC endpoints are also called by backend on hosted deployments. As both have slightly different needs, the protobuffer have more things than necessary and learning_config creation/destruction logic is there too depending on the onprem configuration.

This PR:
- Splits the logic, making the HTTP endpoints only allowed for onprem and gRPC only for hosted. Each side handles it's errors avoiding duplication. 
- Deprecates `learning_config` from `NewKnowledgeBoxRequest` as it was only used by onprem
- Makes `KnowledgeBox` the central piece that everyone call, removing some wrapper functions in `Processor` that made little sense there 

### How was this PR tested?


nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -69,52 +72,23 @@ def __init__(self, txn: Transaction, storage: Storage, kbid: str):
...
+    @processor_observer.wrap({"type": "create_kb"})
     async def create(
         cls,
         txn: Transaction,
         slug: str,
         semantic_model: SemanticModelMetadata,
         uuid: Optional[str] = None,
         config: Optional[KnowledgeBoxConfig] = None,
-        release_channel: ReleaseChannel.ValueType = ReleaseChannel.STABLE,
+        release_channel: Optional[ReleaseChannel.ValueType] = ReleaseChannel.STABLE,
```
lferran, line 88: Optional is not needed when there is a default value no?

nucliadb/src/nucliadb/ingest/service/writer.py
```diff
@@ -101,12 +97,40 @@ async def finalize(self): ...
...
+
+        kbid = request.forceuuid or KnowledgeBoxORM.new_unique_kbid()
+        release_channel = request.release_channel
+        semantic_model = parse_model_metadata_from_request(request)
+
         try:
-            kbid = await self.create_kb(request)
-            logger.info("KB created successfully", extra={"kbid": kbid})
+            async with self.driver.transaction() as txn:
+                kbid, failed = await KnowledgeBoxORM.create(
```
lferran, line 118: now that we are at it, can we simply raise exceptions here and not return the `failed`?

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
```diff
@@ -32,61 +38,101 @@
...
+        # learning will choose the default values
+        user_learning_config = {}
+
+    # we rely on learning to return the updated configuration with defaults and
+    # any other needed values (e.g. matryoshka settings if available)
+    learning_config = await learning_proxy.set_configuration(kbid, config=user_learning_config)
+
+    # if KB creation fails, we'll have to delete its learning config
+    async def _rollback_learning_config(kbid: str):
+        try:
```
lferran, line 92: general comment that will make it easier to test:

WDYT about abstracting all the internal logic into another function and have only http-related logic in the endpoint (error handling and response serialization)?

lferran, line 92: general comment that will make it easier to test:

WDYT about abstracting all the internal logic into another function and have only http-related logic in the endpoint (error handling and response serialization)?

lferran: will make it more readable too

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
jotare: It was to avoid conditionals on `KnowledgeBox.create` callers if they may not have the `release_channel`, but I can move the condition to the POST /kbs endpoint

nucliadb/src/nucliadb/ingest/service/writer.py
jotare: Sure!

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
```diff
@@ -32,61 +38,101 @@
...
-    NewKnowledgeBoxResponse,
-    UpdateKnowledgeBoxResponse,
-)
+from nucliadb_protos import knowledgebox_pb2
 from nucliadb_utils.authentication import requires
-from nucliadb_utils.utilities import get_ingest
+from nucliadb_utils.settings import is_onprem_nucliadb
+
+
+def only_for_onprem(fun):
```
lferran, line 46: NIT: decorators are cool but is it worth the complexity here? Maybe it's easier and more obvious to the reader if you add this in every endpoiint --  like you did for the grpc interface.
Feel free to disregard the comment though

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
```diff
@@ -32,61 +38,101 @@
...
+        title=item.title or "",
+        description=item.description or "",
+    )
+    semantic_model = learning_config.into_semantic_model_metadata()
+    release_channel = item.release_channel.to_pb() if item.release_channel is not None else None
+    try:
+        async with driver.transaction() as txn:
+            kbid, failed = await KnowledgeBox.create(
+                txn,
+                slug=item.slug or "",  # empty slugs will be changed on KB creation
```
lferran, line 114: Should this param be optional then?

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
```diff
@@ -32,61 +38,101 @@
...
+    async def _rollback_learning_config(kbid: str):
+        try:
+            await learning_proxy.delete_configuration(kbid)
+        except Exception:
+            logger.warning(
+                "Could not rollback learning configuration",
+                exc_info=True,
+                extra={"kbid": kbid},
+            )
+            raise
```
lferran, line 100: Do we want to raise here? This will probably make the endpoint finish with a 500 "contact your administrator"

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
jotare: Sure, I can be explicit and add the conditionals in each function. I thought it was better to not duplicate the code, but if it's harder to read, I can change it

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
jotare: Probably, I didn't touch too much the signature of `create` to keep it simpler, but IMO we shouldn't be passing `config` either and we should have something like:
```
async def create(
    kbid: str,
    *,
    slug: Optional[str] = None,
    title: Optional[str] = None,
    description: Optional[str] = None,
    semantic_model: knowledgebox_pb2.SemanticModelMetadata,
    release_channel: Optional[
        utils_pb2.ReleaseChannel.ValueType
    ] = utils_pb2.ReleaseChannel.STABLE,
) -> tuple[str, str]:
```
returning `(kbid, slug)`.

WDYT?

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
jotare: We'll swallow the error. Is a warning enough for the log?

nucliadb/src/nucliadb/writer/api/v1/knowledgebox.py
jotare: Is this duplicated? I'll answer in the one below

jotare: Done

nucliadb/tests/ingest/fixtures.py
```diff
@@ -174,17 +173,12 @@ async def fake_node(indexing_utility, shard_manager):
...
-        semantic_threshold=None,
-        semantic_vector_size=len(V1),
-        semantic_vector_similarity="cosine",
-    )
-    with patch("nucliadb.ingest.service.writer.learning_proxy") as mocked:
-        mocked.set_configuration = AsyncMock(return_value=None)
-        mocked.get_configuration = AsyncMock(return_value=lconfig)
-        mocked.delete_configuration = AsyncMock(return_value=None)
-        yield mocked
+    from nucliadb_utils.settings import nuclia_settings
```
javitonino, line 176: Since this changes the behaviour of the fixture, maybe it needs a different name. THe most obvious would be `dummy_learning_services` or something like that.

javitonino (approved):

PR #2248, opened 2024-06-11
PG pool timeout

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/maindb/pg.py
+++ b/nucliadb/src/nucliadb/common/maindb/pg.py
@@ -282,9 +282,17 @@ class ReadOnlyPGTransaction(Transaction):
 class PGDriver(Driver):
     pool: asyncpg.Pool
 
-    def __init__(self, url: str, connection_pool_max_size: int = 10):
+    def __init__(
+        self,
+        url: str,
+        connection_pool_min_size: int = 10,
+        connection_pool_max_size: int = 10,
+        acquire_timeout_ms: int = 200,
+    ):
         self.url = url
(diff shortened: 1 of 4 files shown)


lferran (approved):

nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -239,8 +242,7 @@ async def count(self, match: str) -> int:
 class ReadOnlyPGTransaction(Transaction):
     driver: PGDriver
 
-    def __init__(self, pool: asyncpg.Pool, driver: PGDriver):
-        self.pool = pool
+    def __init__(self, driver: PGDriver):
```
jotare, line 245: Why are the read only transactions getting a driver and getting a connection every time? Shouldn't they behave as the write transactions? Get a connection, operate and finally close?

javitonino: I decided not to change the behaviour so close to the migration. But, some reasons:
- This is not running in a transaction, so there is no need to keep the connection.
- The read-only transaction may run for a while, e.g: for the entire duration of a read request. This would block the connection for this entire duration of the request, so in practice we would need one connection to PG per concurrent read request.
- Checking out and returning a connection to the pool is pretty fast (it doesn't involve networking unless the connection is inside a transaction when it is returned to the pool, in which case you send a rollback). So, checkout+release is not to costly.

I think this approach works better in the face of long-running read transactions with little DB activity, and we have a clear case of that in search() (waiting for the index to reply). So by keeping to do this, we would use up less of our open connections (which are pretty limited).

PR #2258, opened 2024-06-19
More read-only transactions when getting kb metadata at processing time

lferran:
### Description
Otherwise those keys have contention

### How was this PR tested?
Existing tests should cover


bloodbare (approved):
Shouldn't it be cached ?

PR #2274, opened 2024-06-26
Fix export/import of externally hosted file

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/export_import/datamanager.py
+++ b/nucliadb/src/nucliadb/export_import/datamanager.py
@@ -17,6 +17,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import json
 from datetime import datetime
 from typing import AsyncGenerator, Union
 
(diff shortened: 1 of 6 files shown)


nucliadb/src/nucliadb/export_import/datamanager.py
```diff
@@ -58,20 +59,33 @@ async def get_metadata(self, type: str, kbid: str, id: str) -> Metadata:
...
+        else:
+            raise ValueError(f"Invalid type: {type}")
+        json_decoded = json.loads(decoded)
+
+        # For some reason, the total and processed fields are not always present in the metadata.
+        # This is to unblock already created exports that hit this bug.
+        if json_decoded.get("total") is None:
+            json_decoded["total"] = 0
+        if json_decoded.get("processed") is None:
+            json_decoded["processed"] = 0
```
jotare, line 75: NIT: you could use `setdefault` here
```suggestion
        json_decoded.setdefault("total", 0)
        json_decoded.setdefault("processed", 0)
```

jotare (approved):

nucliadb/src/nucliadb/export_import/datamanager.py
lferran: The problem is that the keys are present but they come set to `None`

PR #2275, opened 2024-06-26
In memory cache with redis invalidations

lferran:
### Description

### How was this PR tested?


lferran (commented):
Make sure that if redis connection is lost, cache is invalidated .
Also, add a TTL cache instead so we make sure that keys are eventually invalidated

lferran:
not needed for now

PR #2278, opened 2024-07-01
Less select for updates

lferran:
### Description
- Reduce number of gets that do select for update

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/datamanagers/vectorsets.py
+++ b/nucliadb/src/nucliadb/common/datamanagers/vectorsets.py
@@ -68,7 +68,7 @@ async def set(txn: Transaction, *, kbid: str, config: knowledgebox_pb2.VectorSet
 
 
 async def delete(txn: Transaction, *, kbid: str, vectorset_id: str):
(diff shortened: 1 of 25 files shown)


nucliadb/src/nucliadb/common/datamanagers/vectorsets.py
```diff
@@ -68,7 +68,7 @@ async def set(txn: Transaction, *, kbid: str, config: knowledgebox_pb2.VectorSet
 
 
 async def delete(txn: Transaction, *, kbid: str, vectorset_id: str):
-    kb_vectorsets = await _get_or_default(txn, kbid=kbid)
+    kb_vectorsets = await _get_or_default(txn, kbid=kbid, for_update=False)
```
javitonino, line 71:
```suggestion
    kb_vectorsets = await _get_or_default(txn, kbid=kbid, for_update=True)
```

This modifies the vectorset list key so it needs for_update even if it's a deletion. I think we talked about this and I gave you the wrong answer, sorry.

javitonino (approved):

PR #2287, opened 2024-07-03
Vectorsets everywhere

jotare:
### Description
More things around vectorsets :)
- New KB creation gRPC endpoint
- Extended new shard to accept multiple vectorsets
- Start removing default vectorset concept

### How was this PR tested?


nucliadb_protos/writer.proto
```diff
@@ -444,9 +445,29 @@ message DelVectorSetResponse {
...
+    string title = 3;
+    string description = 4;
+
+    message VectorSet {
+        string vectorset_id = 1;
+        utils.VectorSimilarity similarity = 2;
+        uint32 vector_dimension = 3;
+        repeated uint32 matryoshka_dimensions = 4;
+    }
+    repeated VectorSet vectorsets = 5;
```
javitonino, line 460: Why not reuse the existing message `VectorSetConfig`?

nucliadb_node/src/shards/shard_writer.rs
```diff
@@ -151,88 +105,111 @@ impl ShardWriter {
...
+            return Err(node_error!("Shards must be created with at least one vector index"));
+        }
+
+        let shard_id = new.shard_id;
+        let shard_path = disk_structure::shard_path_by_id(shards_path, &shard_id);
+        let metadata = Arc::new(ShardMetadata::new(shard_path.clone(), shard_id.clone(), new.kbid, new.channel));
+        let mut indexes = ShardIndexes::new(&shard_path);
 
-        std::fs::create_dir(shard_path)?;
+        std::fs::create_dir(shard_path.clone())?;
```
javitonino, line 120:
```suggestion
        std::fs::create_dir(&shard_path)?;
```

nucliadb_node/src/shards/indexes.rs
```diff
@@ -305,7 +313,7 @@ mod tests {
 
         let mut indexes = ShardIndexes::new(shard_path);
 
-        for i in 0..(MAX_ALLOWED_VECTORSETS - 1) {
+        for i in 0..(MAX_ALLOWED_VECTORSETS) {
```
javitonino, line 316:
```suggestion
        for i in 0..MAX_ALLOWED_VECTORSETS {
```

nucliadb_protos/writer.proto
jotare: Because there's a bunch of params that won't be populated (normalize and vector type), but I can change it if you consider it cleaner

javitonino: Ah, I misread that message. Indeed, it makes sense to have it be a separate thing. :+1:  Thanks!

nucliadb/src/nucliadb/common/cluster/base.py
```diff
@@ -108,6 +108,21 @@ async def new_shard(
...
+        release_channel: utils_pb2.ReleaseChannel.ValueType,
+        vectorsets_configs: dict[str, VectorIndexConfig],
+    ) -> noderesources_pb2.ShardCreated:
+        req = NewShardRequest(
+            kbid=kbid,
+            release_channel=release_channel,
+            vectorsets_configs=vectorsets_configs,
+        )
+
+        resp = await self.writer.NewShard(req)  # type: ignore
```
lferran, line 123: We should be validating that the response was successful no?
Or will the node return a grpc error in the event of an error?

nucliadb/src/nucliadb/common/cluster/manager.py
```diff
@@ -223,43 +223,41 @@ async def create_shard_by_kbid(
...
                 except Exception as e:
                     errors.capture_exception(e)
-                    logger.exception(
-                        "Error creating vectorset '{vectorset_id}' new shard at {node}: {details}".format(
-                            vectorset_id=vectorset_config.vectorset_id,
-                            node=node,
-                            details=e,
-                        )
-                    )
+                    logger.exception(f"Error creating new shard at {node}: {e}")
```
lferran, line 260: you can use the `extra` to have a cleaner error log without the need of string formatting

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -80,78 +85,140 @@ def new_unique_kbid() -> str:
...
+            )
+        if semantic_model is None:
+            if semantic_models is None or len(semantic_models) == 0:
+                raise KnowledgeBoxCreationError(
+                    "KB must be created with at least one semantic model spec"
+                )
+        else:
+            if semantic_models is not None and len(semantic_models) > 0:
+                raise KnowledgeBoxCreationError(
+                    "KB must be created with at least one semantic model spec"
```
lferran, line 112: wrong error message: shoud be something like "KB must only define semantic_model or semantic_models, not both". But aren't you checking the same thing twice?

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -80,78 +85,140 @@ def new_unique_kbid() -> str:
...
+                    # only one vectorset
+
+                    # if this KB uses a matryoshka model, we can choose a different
+                    # dimension
+                    if len(semantic_model.matryoshka_dimensions) > 0:
+                        semantic_model.vector_dimension = choose_matryoshka_dimension(
+                            semantic_model.matryoshka_dimensions  # type: ignore
+                        )
+                    kb_shards.similarity = semantic_model.similarity_function
+                    kb_shards.model.CopyFrom(semantic_model)
```
lferran, line 163: shouldn't we set the vectorsets here aswell, with the `__default__` vectorset_id?

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -80,78 +85,140 @@ def new_unique_kbid() -> str:
...
+                # txn to be open!
+                await shard_manager.create_shard_by_kbid(txn, kbid)
+                # shards don't need a rollback as they will be eventually purged
+
+                await txn.commit()
+
+        except Exception:
+            # rollback all changes on the db and raise the exception
+            for op in reversed(rollback_ops):
+                await op()
```
lferran, line 218: wdyt about guarding each `await op()` inside a try/except with a logg.error + continue? so we attempt to rollback as much as possible but in the end raise the error we wanted to raise

nucliadb/src/nucliadb/ingest/service/writer.py
```diff
@@ -136,11 +189,11 @@ async def NewKnowledgeBox(  # type: ignore
                 exc_info=True,
                 extra={"slug": request.slug},
             )
-            return NewKnowledgeBoxResponse(status=KnowledgeBoxResponseStatus.ERROR)
+            return writer_pb2.NewKnowledgeBoxV2Response(status=KnowledgeBoxResponseStatus.ERROR)
 
         else:
             logger.info("KB created successfully", extra={"kbid": kbid})
-            return NewKnowledgeBoxResponse(status=KnowledgeBoxResponseStatus.OK, uuid=kbid)
+            return writer_pb2.NewKnowledgeBoxV2Response(status=KnowledgeBoxResponseStatus.OK)
```
lferran, line 196: you're not passing the kbid anymore to backend after success KB creation. This is needed by backend no?

nucliadb/tests/ingest/fixtures.py
```diff
@@ -185,14 +185,12 @@ def learning_config():
...
-            similarity_function=upb.VectorSimilarity.COSINE, vector_dimension=len(V1)
-        )
-        await KnowledgeBox.create(txn, kbslug, model, uuid=kbid)
-        await txn.commit()
+    kbid = KnowledgeBox.new_unique_kbid()
+    kbslug = "slug-" + str(uuid.uuid4())
+    model = SemanticModelMetadata(
+        similarity_function=upb.VectorSimilarity.COSINE, vector_dimension=len(V1)
+    )
+    await KnowledgeBox.create(maindb_driver, kbid=kbid, slug=kbslug, semantic_model=model)
```
lferran, line 193: Are you still using the deprecated creation on purpose, so we still test that code path?

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
jotare: Yes, I left an unnecessary check. Cleaning it up

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
jotare: No, I'm not populating the /vectorsets key on KBs with only the default on purpose, so migration will be easier later. At this point, we don't know the model name (vectorset_id), so setting `__default__` here would require another check in the migration (KBs without the key, KBs with __default__ and KBs with vectorsets)

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
jotare: Sure, good idea!

nucliadb/tests/ingest/fixtures.py
jotare: Yes, but I can use the new one if you prefer

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -114,7 +114,9 @@ async def create(
 
         try:
             async with driver.transaction() as txn:
-                exists = await datamanagers.kb.get_kb_uuid(txn, slug=slug)
+                exists = await datamanagers.kb.get_kb_uuid(
+                    txn, slug=slug
+                ) or await datamanagers.kb.exists_kb(txn, kbid=kbid)
```
lferran, line 119: shouldn't it simply be `exists = await datamanagers.kb.exists_kb(txn, kbid=kbid)` ?

PR #2288, opened 2024-07-05
Be able to create a KnowledgeBox with a Pinecone index

lferran:
### Description
This PR adds wiring to the knowledgebox creation logic to be able to pass a pinecone api key so that a pinecone index is used instead of our index node.

The index is not populated yet with vectors nor queried. These will come on future PRs.

### How was this PR tested?
Integration and unit tests


lferran:
[sc-10590]

PR #2294, opened 2024-07-09
Refactor audit sending to send one message per request

sunbit:
### Description
The issue was that on the activity log we were receiving two separate events for internal search operations and chat operations. Aside of causing two entries per chat request on a database that may get pretty big, the information we want together when we do generative question answers was splitted.

The idea now is that the audit request is created in the context of a request, and each audit operation adds to that AuditRequest. Once the request is done, if theres any audit data collected, it's queued to send. The last audit overwrites any shared information, so is the last one the sets the type, which matches what we want right now, this may be improved in the future.

Not 100% happy on how i needed to rewrap the streaming responses, it's the only way i found to overcome the fact that the middleware hook happens before the stream generator is consumed. I did that because it seemed the less intrusive. The other option was to explicitly return a custom StreamingResponse class with some end action, but that would require using that in all parts that requires auditing.

### How was this PR tested?
Added a couple of integration tests for the use cases that previously sent two messages, testing that now send one with all the information


nucliadb_utils/src/nucliadb_utils/audit/basic.py
```diff
@@ -50,20 +50,19 @@ async def report(
...
-        resources: int,
-    ):
+    def report_resources(self, *, kbid: str, resources: int):
         logger.debug(f"REPORT RESOURCES {kbid} {resources}")
 
     async def visited(self, kbid: str, uuid: str, user: str, origin: str):
         logger.debug(f"VISITED {kbid} {uuid} {user} {origin}")
 
+    def send(self, msg: AuditRequest):
+        pass
```
jotare, line 71: You could add a `logger.debug` for testing (as the other methods have)

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -54,6 +58,77 @@
...
+            response = self.wrap_streaming_response(response, context)
+        else:
+            self.enqueue_pending_audit_request(context.audit_request)
+
+        request_context_var.reset(token)
+
+        return response
+
+    def enqueue_pending_audit_request(self, audit_request: AuditRequest):
+        # Circular dependency, open to ideas to get rid of this
```
jotare, line 98: As middleware could be used with any audit utility, could be in another file or require the audit utility to be passed in middleware creation

nucliadb_utils/src/nucliadb_utils/audit/audit.py
```diff
@@ -61,28 +57,29 @@ async def initialize(self):
...
         self,
         kbid: str,
+        uuid: str,
         user: str,
-        client: int,
         origin: str,
-        search: SearchRequest,
-        timeit: float,
-        resources: int,
+        send: bool = False,
```
jotare, line 66: This boolean deciding whether to call or not another public method, mixed with all the audit parameters, feels kind of weird.
As the only user is ingest, how do you feel calling `report` and `send` or a "`report_and_send`" method?

jotare (commented):
Few small comments, but looks good!

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -54,6 +59,78 @@
...
+            response = self.wrap_streaming_response(response, context)
+        else:
+            self.enqueue_pending(context)
+
+        request_context_var.reset(token)
+
+        return response
+
+    def enqueue_pending(self, context: RequestContext):
+        # Circular dependency, open to ideas to get rid of this
```
lferran, line 98: As Joan Antoni says, why not having all the middleware logic under `nucliadb_utils/audit/middleware.py` that will depend on the generic utility

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -54,6 +59,78 @@
...
+def get_request_context() -> Optional[RequestContext]:
+    return request_context_var.get()
+
+
+class AuditMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request: Request, call_next: RequestResponseEndpoint) -> Response:
+        context = RequestContext()
+        token = request_context_var.set(context)
+        context.audit_request.time.FromDatetime(datetime.now(tz=timezone.utc))
+        response = await call_next(request)
```
lferran, line 87: are unhandled error already handled once we hit this line of code? or should there be a try/catch here?

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -222,27 +300,46 @@ def report_resources(
             storage=Storage(resources=resources),
         )
 
-    async def visited(self, kbid: str, uuid: str, user: str, origin: str):
-        auditrequest = AuditRequest()
+    async def visited(
```
lferran, line 303: Seems like some of the audit methods are in fact not async. Can you do a pass on all of them and see if we can turn some of those into sync calls?

nucliadb_utils/src/nucliadb_utils/audit/basic.py
sunbit: done

nucliadb_utils/src/nucliadb_utils/audit/audit.py
sunbit: This made me realize that they were some audit functions already deprecated and some other that only used the inner kbusage utility, i decided to split the audit and report into completely independent utilities and use them explicitly where needed. This way the only "send=True" needed was in the `report`.

nucliadb_utils/src/nucliadb_utils/audit/stream.py
sunbit: already handled in the same way of any other unhander error anywhere else

nucliadb_utils/src/nucliadb_utils/audit/stream.py
sunbit: i refactored them all to be sync

nucliadb_utils/src/nucliadb_utils/audit/stream.py
sunbit: i like that idea to add the utility as a paramer to the middleware!, done

nucliadb/tests/ingest/fixtures.py
```diff
@@ -203,18 +203,20 @@ async def audit():
...
+        nats_servers=[natsd], nats_subject=usage_settings.usage_jetstream_subject
+    )
+    await report_util.initialize()
+    report_util.nats_connection_manager.js.publish = AsyncMock(
+        side_effect=report_util.nats_connection_manager.js.publish
     )
-    await audit.initialize()
-    yield audit
-    await audit.finalize()
+    set_utility(Utility.USAGE, report_util)
```
jotare, line 217: You must call `clean_utility` just after the yield to remove the utility from the register. It probably works because I think we have some hard cleanup, but can be problematic

nucliadb/tests/search/integration/api/v1/test_ask.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
jotare, line 1: Could you rename it to `test_ask_audit` or similar? So it's easier to find later?

nucliadb_utils/src/nucliadb_utils/tests/nats.py
```diff
@@ -191,7 +191,11 @@ def start_gnatsd(gnatsd: Gnatsd):  # pragma: no cover
 
 @pytest.fixture(scope="session")
 def natsd_server():  # pragma: no cover
-    if not os.path.isfile("nats-server"):
+    # Create a persistent temporary directory
+    tmpdir = tempfile.mkdtemp()
+    nats_server_path = os.path.join(tmpdir, "nats-server")
+
+    if not os.path.isfile(nats_server_path):
```
jotare, line 198: Not sure if we want to install nats-server every test run... @nuclia/nuclia-db?

lferran (approved):

PR #2295, opened 2024-07-09
Add PG pool metrics

javitonino:
Adds PG pool metrics

### DIFF
--- a/nucliadb/src/nucliadb/common/maindb/pg.py
+++ b/nucliadb/src/nucliadb/common/maindb/pg.py
@@ -44,13 +44,32 @@ CREATE TABLE IF NOT EXISTS resources (
 );
 """
 
+logger = logging.getLogger(__name__)
 
+# Request Metrics
 pg_observer = metrics.Observer(
     "pg_client",
     labels={"type": ""},
 )
 
-logger = logging.getLogger(__name__)
+# Pool metrics
+POOL_METRICS_COUNTERS = {
+    # Requests for a connection to the pool
+    "requests_num": metrics.Counter("pg_client_pool_requests_total"),
+    "requests_queued": metrics.Counter("pg_client_pool_requests_queued_total"),
+    "requests_errors": metrics.Counter("pg_client_pool_requests_errors_total"),
+    "requests_wait_ms": metrics.Counter("pg_client_pool_requests_queued_seconds_total"),
+    "usage_ms": metrics.Counter("pg_client_pool_requests_usage_seconds_total"),
+    # Pool opening a connection to PG
+    "connections_num": metrics.Counter("pg_client_pool_connections_total"),
+    "connections_ms": metrics.Counter("pg_client_pool_connections_seconds_total"),
+}
+POOL_METRICS_GAUGES = {
+    "pool_size": metrics.Gauge("pg_client_pool_connections_open"),
+    # The two below most likely change too rapidly to be useful in a metric
+    "pool_available": metrics.Gauge("pg_client_pool_connections_available"),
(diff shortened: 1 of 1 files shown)


nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -44,13 +44,32 @@
...
+    "usage_ms": metrics.Counter("pg_client_pool_requests_usage_seconds_total"),
+    # Pool opening a connection to PG
+    "connections_num": metrics.Counter("pg_client_pool_connections_total"),
+    "connections_ms": metrics.Counter("pg_client_pool_connections_seconds_total"),
+}
+POOL_METRICS_GAUGES = {
+    "pool_size": metrics.Gauge("pg_client_pool_connections_open"),
+    # The two below most likely change too rapidly to be useful in a metric
+    "pool_available": metrics.Gauge("pg_client_pool_connections_available"),
+    "requests_waiting": metrics.Gauge("pg_client_pool_requests_waiting"),
```
lferran, line 71: we could compute the rolling average in memory and report that, but not sure if it's worth the complexity

nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -275,11 +294,33 @@ async def initialize(self):
...
             await self.pool.close()
             self.initialized = False
+            self.metrics_task.cancel()
+
+    async def _report_metrics_task(self):
+        while True:
+            await self._report_metrics()
+            await asyncio.sleep(60)
+
+    async def _report_metrics(self):
```
lferran, line 310: can be a sync method

nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -44,13 +44,32 @@
...
-logger = logging.getLogger(__name__)
+# Pool metrics
+POOL_METRICS_COUNTERS = {
+    # Requests for a connection to the pool
+    "requests_num": metrics.Counter("pg_client_pool_requests_total"),
+    "requests_queued": metrics.Counter("pg_client_pool_requests_queued_total"),
+    "requests_errors": metrics.Counter("pg_client_pool_requests_errors_total"),
+    "requests_wait_ms": metrics.Counter("pg_client_pool_requests_queued_seconds_total"),
+    "usage_ms": metrics.Counter("pg_client_pool_requests_usage_seconds_total"),
+    # Pool opening a connection to PG
```
lferran, line 63: maybe you want to rename some of the metrics.Counter, as internally adds the trailing `_count` already

nucliadb/src/nucliadb/common/maindb/pg.py
javitonino: Yeah, but then you have to loop pretty frequently in order to get a value. And you can get an idea of average usage with other metrics (e.g: average queue time). I don't think it's worth the complexity.

I thought about removing them, but they might be useful if things go too badly and a queue starts to build up. Since it's only two metrics, I left them around in case they are useful some day in an emergency.

nucliadb/src/nucliadb/common/maindb/pg.py
javitonino: True, changed! Thanks!

nucliadb/src/nucliadb/common/maindb/pg.py
javitonino: Yeah, I know it adds `_total` if it's not there. However I find that having the same name here in the code as the metric in prometheus makes it easier to search for things. It would not be the first time I put the prometheus name into the code search and found nothing because something was being added to the name automatically.

PR #2297, opened 2024-07-09
Add encryption util

lferran:
### Description
This utility allows to encrypt/decrypt text and bytes via a shared secret key.
It uses the PyNacl library, which was already a dependency from nucliadb.
The first use-case will be to encrypt the pinecone api keys before storing them in pg.

### How was this PR tested?

### DIFF
--- a/nucliadb_utils/src/nucliadb_utils/encryption/__init__.py
+++ b/nucliadb_utils/src/nucliadb_utils/encryption/__init__.py
@@ -17,41 +17,57 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-import asyncio
-from concurrent.futures import ThreadPoolExecutor
+import base64
 
 import nacl.secret
 import nacl.utils
+
+from nucliadb_telemetry.metrics import Observer
+
+
+class DecryptionError(Exception): ...
+
+
+class DecryptionDataError(DecryptionError): ...
+
+
+endecryptor_observer = Observer(
+    "endecryptor",
+    error_mappings={
+        "DecryptionError": DecryptionError,
+        "DecryptionDataError": DecryptionDataError,
+    },
+    labels={"operation": ""},
+)
 
 
 class EndecryptorUtility:
     """
     Utility class for encryption and decryption of sensitive data using the nacl library.
-    It uses a thread pool to avoid blocking the event loop.
     """
 
-    def __init__(self, key: bytes, max_thread_pool_size: int = 1):
-        self.box = nacl.secret.SecretBox(key)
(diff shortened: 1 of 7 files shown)


nucliadb_utils/src/nucliadb_utils/encryption/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        self.box = nacl.secret.SecretBox(key)
+        self.executor = ThreadPoolExecutor(max_workers=max_thread_pool_size)
+
+    async def encrypt_text(self, message: str) -> str:
+        message_bytes = message.encode()
+        encrypted = await self.encrypt(message_bytes)
+        return encrypted.hex()
+
+    async def encrypt(self, message: bytes) -> nacl.utils.EncryptedMessage:
+        loop = asyncio.get_running_loop()
```
javitonino, line 43: Using a threadpool might be a bit overkill for encrypting short texts with a symmetric algorithm, but it's there and definitely won't hurt :)

nucliadb_utils/src/nucliadb_utils/encryption/settings.py
```diff
@@ -0,0 +1,38 @@
...
+
+from pydantic import Field
+from pydantic_settings import BaseSettings
+
+
+class EncryptionSettings(BaseSettings):
+    encryption_secret_key: Optional[str] = Field(
+        default=None,
+        title="Encryption Secret Key",
+        description="Secret key used for encryption and decryption of sensitive data in the database. The key must be a 32-byte string, base64 encoded.",
```
javitonino, line 29: Might be nice to have a mention here to a short command to generate a random key, e.g, the very unixy: `head -c 32 /dev/urandom | base64` or whatever you used to generate this.

nucliadb_utils/tests/unit/encryption/test_encryption.py
```diff
@@ -0,0 +1,54 @@
...
+    # Get a 32-bytes random string
+    secret_key = uuid4().hex[:32]
+    # Encode the secret key in base64
+    b64_encoded_secret_key = base64.b64encode(secret_key.encode()).decode()
+    settings.encryption_secret_key = b64_encoded_secret_key
+
+    yield settings
+
+
+async def test_endecryptor(encryption_settings):
```
javitonino, line 38: Would be nice to add another test that:
- Sets keyA
- Encrypts a text
- Sets keyB
- Decrypts the previous ciphertext

It should raise a PyNaCl error that it cannot decrypt the data (this is authenticated encryption, so it should be able to detect errors in decryption).

javitonino (approved):
Looks nice :sparkles: Just a couple of suggestions

lferran:
[sc-10590]

PR #2301, opened 2024-07-10
Add pinecone session util

lferran:
### Description
If we move forward with the integration, ideally the pinecone client should become its own package that we may eve publish to publish to pypi.

For now, this is the fastest /easiest way we have to share the client code across nucliadb components (will have to be used in the nucliadb_sidecar and nucliadb itself)

### How was this PR tested?
Unit tests


nucliadb_utils/src/nucliadb_utils/aiopynecone/client.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @pinecone_observer.wrap({"type": "delete_index"})
+    async def delete_index(self, name: str) -> None:
+        headers = {"Api-Key": self.api_key}
+        response = await self.session.delete(f"/indexes/{name}", headers=headers)
+        if response.status_code == 404:
+            logger.warning("Pinecone index not found.", extra={"index_name": name})
+            return
+        response.raise_for_status()
+        return
```
javitonino, line 63: Unneeded return?

javitonino (approved):

nucliadb_utils/src/nucliadb_utils/aiopynecone/client.py
lferran: yup, will change it on my next pr

lferran:
[sc-10590]

PR #2302, opened 2024-07-10
Pinecone KB creation

lferran:
### Description
Be able to create a KB that uses Pinecone as index provider
Pinecone api key is stored encrypted
The index is not populated  nor queried. These will come on future PRs.

### How was this PR tested?
Integration tests


nucliadb_protos/knowledgebox.proto
```diff
@@ -21,6 +21,40 @@ message KnowledgeBoxID {
     string uuid = 2;
 }
 
+// External Index node provider
+enum ExternalIndexProviderType {
```
jotare, line 25: I think we could think about nucliadb's index node as an index provider too, so we don't have to think about an abstraction only for Pinecone.
Therefore, the `UNSET` become `NUCLIA` or `NUCLIADB`.

nucliadb_protos/knowledgebox.proto
```diff
@@ -21,6 +21,40 @@ message KnowledgeBoxID {
...
+    // Index name to index host mapping.
+    map<string, string> index_hosts = 2;
+}
+
+// External Index node provider
+message CreateExternalIndexProviderMetadata {
+    ExternalIndexProviderType type = 1;
+    oneof config {
+        CreatePineconeConfig pinecone_config = 2;
+    }
```
jotare, line 47: No need for type if you're using `oneof`. If there's a `CreatePineconeConfig`, we know is a pinecone config, no need to store a type

nucliadb/src/nucliadb/ingest/service/writer.py
```diff
@@ -121,6 +121,7 @@ async def NewKnowledgeBox(  # type: ignore
                 description=request.config.description,
                 semantic_model=semantic_model,
                 release_channel=release_channel,
+                external_index_provider=request.external_index_provider,
```
jotare, line 124: As it is right now, we'll have the default with type `UNSET` but a default pinecone config in the oneof field, as backend is not populating anything

nucliadb/src/nucliadb/ingest/orm/knowledgebox.py
```diff
@@ -463,6 +486,54 @@ async def delete_vectorset(self, vectorset_id: str):
         shard_manager = get_shard_manager()
         await shard_manager.delete_vectorset(self.kbid, vectorset_id)
 
+    @classmethod
+    async def _maybe_create_external_index(
+        cls,
```
jotare, line 491: If you're in into having nucliadb index node as an index provider, we could maybe work on the `KBShardManager` abstraction to become the nucliadb index provider and have this pinecone logic there, outside the `KnowledgeBox` class

nucliadb_protos/knowledgebox.proto
lferran: type can be useful for checking what config to parse.
Otherwise we'll need to use proto's `HasField("")` for every possibility.

nucliadb/src/nucliadb/ingest/service/writer.py
lferran: this is always the case with protobuf. the concept of optional fields isn't really supported, they will always default to the default value of the primitive fields.

This is why having a type can help and make the code much cleaner:
- first check the type, then access the corresponding config.

jotare (approved):

lferran:
[sc-10590]

PR #2308, opened 2024-07-11
Build docker images using PDM dependency locks

javitonino:
- Make Dockerfile use the requirements locked by PDM. After this, the only thing using the old `requirements.lock.txt` is the install-vm script (and thus, e2e tests). Everything else is using PDM.
- Remove unused images
- Simplified Rust images (remove cargo chef)
- The old `Dockerfile` image used to download bindings from PyPi which was inconsistent with the rest of components in the image. There was a `Dockerfile.withbinding` that used the bindings. This PR removes the binding from the former, and changes tests using standalone mode to use the second image.

Future cleanup, pending decision:
- Remove requirements.lock.txt (either remove install-vm.sh script, or switch it to use PDM or even just rely on automatic dependency resolution)
- Remove requirements-test.txt and `make install-dev`
- Review build of binding for kubernets E2E tests, it's very slow (no cache). We could add cache or reuse the binding built for CI tests if we run inside the CI pipeline.


lferran (approved):
💘 thank you for the cleanup!

PR #2318, opened 2024-07-16
Pinecone writes

lferran:
### Description
Set and delete resources to/from a pinecone index

### How was this PR tested?
Unit and local tests

### DIFF
--- a/nucliadb/src/nucliadb/common/external_index_providers/base.py
+++ b/nucliadb/src/nucliadb/common/external_index_providers/base.py
@@ -1,0 +1,60 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+import abc
+import logging
+
+from nucliadb_protos.noderesources_pb2 import Resource
+
+logger = logging.getLogger(__name__)
+
+
+class ExternalIndexManager(abc.ABC, metaclass=abc.ABCMeta):
+    type: str
+
+    def __init__(self, kbid: str):
+        self.kbid = kbid
+
+    async def delete_resource(self, resource_uuid: str) -> None:
+        logger.info(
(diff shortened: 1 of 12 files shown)


nucliadb/src/nucliadb/common/external_index_providers/base.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __init__(self, kbid: str):
+        self.kbid = kbid
+
+    async def delete_resource(self, resource_uuid: str) -> None:
+        logger.info(
+            "Deleting resource to external index",
+            extra={
+                "kbid": self.kbid,
+                "resource_uuid": resource_uuid,
```
jotare, line 39: I think we should start to standarize logging keys. We have a mix between `rid`, `resource_id` and `resource_uuid`. Same for the variable name (although less important IMO)

nucliadb/src/nucliadb/common/external_index_providers/manager.py
```diff
@@ -0,0 +1,60 @@
...
+        delete_parallelism=settings.pinecone_delete_parallelism,
+        upsert_timeout=settings.pinecone_upsert_timeout,
+        delete_timeout=settings.pinecone_delete_timeout,
+    )
+
+
+@async_lru.alru_cache(maxsize=None)
+async def get_external_index_metadata(kbid: str) -> Optional[StoredExternalIndexProviderMetadata]:
+    async with datamanagers.with_ro_transaction() as txn:
+        return await datamanagers.kb.get_external_index_provider_metadata(txn, kbid=kbid)
```
jotare, line 60: You could use a decorated atomic instead of repeating the function here

nucliadb/src/nucliadb/common/external_index_providers/settings.py
```diff
@@ -0,0 +1,47 @@
...
+    pinecone_upsert_timeout: int = Field(
+        default=10,
+        title="Pinecone upsert timeout",
+        description="Timeout in seconds for each upsert operation to Pinecone",
+    )
+    pinecone_delete_timeout: int = Field(
+        default=10,
+        title="Pinecone delete timeout",
+        description="Timeout in seconds for each delete operation to Pinecone",
+    )
```
jotare, line 44: Timeouts are usually expressed as floats in the rest of the codebase

nucliadb/src/nucliadb/ingest/orm/processor/__init__.py
```diff
@@ -182,8 +184,8 @@ async def delete_resource(
                     shard = await kb.get_resource_shard(shard_id)
                     if shard is None:
                         raise AttributeError("Shard not available")
-
-                    await self.shard_manager.delete_resource(
+                    await self._maybe_external_index_delete_resource(message.kbid, uuid)
+                    await self.index_node_shard_manager.delete_resource(
```
jotare, line 188: Shouldn't this operations be exclusive? Why do we need to delete a resource on the nodes if we are using an external index?

lferran: It's not exclusive because we are still using the texts index (for catalog purposes).
Once we rewrite catalog in a pg index, we will be able to do it exclusive

jotare: Oh that's true...

jotare (approved):

PR #2321, opened 2024-07-17
Less locking when deleting resources

javitonino:
Remove some unnecessary locking when deleting a resource.

### DIFF
--- a/nucliadb/src/nucliadb/common/datamanagers/resources.py
+++ b/nucliadb/src/nucliadb/common/datamanagers/resources.py
@@ -63,7 +63,7 @@ async def resource_exists(txn: Transaction, *, kbid: str, rid: str) -> bool:
 
 
 async def get_resource_uuid_from_slug(txn: Transaction, *, kbid: str, slug: str) -> Optional[str]:
-    encoded_uuid = await txn.get(KB_RESOURCE_SLUG.format(kbid=kbid, slug=slug))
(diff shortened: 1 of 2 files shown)


nucliadb/src/nucliadb/common/datamanagers/resources.py
```diff
@@ -101,8 +101,10 @@ async def modify_slug(txn: Transaction, *, kbid: str, rid: str, new_slug: str) -
 
 
 @backoff.on_exception(backoff.expo, (Exception,), jitter=backoff.random_jitter, max_tries=3)
-async def get_resource_shard_id(txn: Transaction, *, kbid: str, rid: str) -> Optional[str]:
-    shard = await txn.get(KB_RESOURCE_SHARD.format(kbid=kbid, uuid=rid))
+async def get_resource_shard_id(
+    txn: Transaction, *, kbid: str, rid: str, for_update: bool = False
```
lferran, line 105: should default be to True, to be safer?

lferran (approved):

nucliadb/src/nucliadb/common/datamanagers/resources.py
lferran: although maybe it's ok as we're only really changing this key on rebalance and on resource creation, right?

javitonino: Yeah, I felt comfortable defaulting to false since this is only ever updated in rebalance, which I think it's obvious enough. In resource creation the key is created, so this is never called.

PR #2325, opened 2024-07-18
Pinecone index queries

lferran:
### Description
A bunch of search features are still missing, but this provides basic semantic search. Extra features will be implemented in upcoming PRs.

Implemented in this PR:
- Semantic retrieval
- Min score filtering

Missing features (not part of this PR):
- label filtering
- date range filtering
- paragraph second positions
- filter by resource id
- etc

The PR also adds a way to create onprem KBs using a pinecone index. This is very convenient for development

### How was this PR tested?
TODO


nucliadb/src/nucliadb/common/external_index_providers/base.py
```diff
@@ -19,42 +19,131 @@
...
+    id: str
+    resource_id: str
+    field: str
+    index: int
+    position_start: int
+    position_end: int
+    split: Optional[str] = None
+    text: Optional[str] = None
+    score: float
+    order: int
```
jotare, line 57: Could you use more verbose names? Something like
```suggestion
    id: str
    resource_id: str
    field_id: str
    index: int
    position_start: int
    position_end: int
    subfield_id: Optional[str] = None
    text: Optional[str] = None
    score: float
    order: int
```

Also, isn't `index` specific to vectors? Shouldn't be this an optional if we want to use this model for paragraphs?

nucliadb/src/nucliadb/search/search/results_hydrator/pinecone.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
```
jotare, line 1: Is this module needed?

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1246,6 +1246,7 @@ class SCORE_TYPE(str, Enum):
     VECTOR = "VECTOR"
     BM25 = "BM25"
     BOTH = "BOTH"
+    EXTERNAL = "EXTERNAL"
```
jotare, line 1249: Should we give more detail or "hiding" any implementation is fine?

nucliadb/src/nucliadb/search/search/results_hydrator/base.py
```diff
@@ -0,0 +1,215 @@
...
+    hydrate_ops = []
+    semaphore = asyncio.Semaphore(max_parallel_operations)
+    extracted_text_cache = paragraphs.ExtractedTextCache()
+    rcache = get_resource_cache(clear=True)
+    try:
+        resource_ids = set()
+        for text_block in query_results.iter_matching_text_blocks():
+            if text_block_min_score is not None and text_block.score < text_block_min_score:
+                # Ignore text blocks with a score lower than the minimum
+                continue
```
jotare, line 96: I guess this is from the original implementation, but isn't this hiding a bug in the index?

jotare (commented):
Looks really nice!

nucliadb/src/nucliadb/common/external_index_providers/base.py
```diff
@@ -19,42 +19,131 @@
...
+class QueryResults(BaseModel):
+    """
+    Model for the results of a query to an external index provider.
+    Must be subclassed by the specific external index provider.
+    """
+
+    type: ExternalIndexProviderType
+    results: Any
+
+    def iter_matching_text_blocks(self) -> Iterator[TextBlockMatch]:
```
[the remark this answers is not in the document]
lferran: yes! that is why I added the `raise NotImplementedError()`.  But I didn't want to extend the class with abstract abc metaclass since it is a model

nucliadb/src/nucliadb/search/search/results_hydrator/pinecone.py
lferran: no indeed

nucliadb/src/nucliadb/search/search/results_hydrator/base.py
lferran: Not really! PInecone doesn't have a way to filter by min_score, so we need to do it on our end

nucliadb_models/src/nucliadb_models/search.py
lferran: Actually, this should still be "vector" I think

jotare (approved):

PR #2343, opened 2024-07-25
Small fixes and improvements

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/tests/ndbfixtures/maindb.py
+++ b/nucliadb/tests/ndbfixtures/maindb.py
@@ -106,11 +106,21 @@ async def maindb_driver(request: FixtureRequest) -> AsyncIterator[Driver]:
 async def pg_maindb_settings(pg):
     url = f"postgresql://postgres:postgres@{pg[0]}:{pg[1]}/postgres"
 
+    # We want to be sure schema migrations are always run. As some tests use
+    # this fixture and create their own driver, we need to create one here and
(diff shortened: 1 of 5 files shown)


nucliadb/tests/ndbfixtures/maindb.py
```diff
@@ -106,13 +106,11 @@ async def maindb_driver(request: FixtureRequest) -> AsyncIterator[Driver]:
 async def pg_maindb_settings(pg):
     url = f"postgresql://postgres:postgres@{pg[0]}:{pg[1]}/postgres"
 
-    driver = PGDriver(url=url, connection_pool_min_size=2, connection_pool_max_size=2)
-    await driver.initialize()
-    await run_pg_schema_migrations(driver)
-
-    return DriverSettings(
+    yield DriverSettings(
```
javitonino, line 109: Just wanted to clarify this. I added this min/max_size = 2 and I only wanted to use that for the `run_pg_schema_migrations`. I don't think this should be in the returned settings, unless you got a reason why.

jotare: Then, why are you running schema migration twice? The `pg_maindb_driver` also runs them. Is there any reason?

jotare: Updated, `pg_maindb_settings` uses a driver to run migrations

javitonino (approved):

PR #2360, opened 2024-07-30
Migrate vectorset config to maindb vectorsets keys

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/migrations/0024_assign_models_to_kbs.py
+++ b/nucliadb/src/migrations/0024_assign_models_to_kbs.py
@@ -1,0 +1,112 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""Migration #24
+
+Vectorsets are coming and we need to be ready at nucliadb. Vector index config
+shouldn't be stored anymore in the `Shards` protobuffer, we need to migrate to
+the new vectorsets config.
+
+This migration asks learning_config for each KB configuration and saves the
+model name as the vectorset_id. Creates a vectorset configuration for each model
+and deprecates the vectors index config from the `Shards` protobuffer.
+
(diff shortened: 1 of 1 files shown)


nucliadb/src/migrations/0024_assign_models_to_kbs.py
```diff
@@ -0,0 +1,112 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def migrate(context: ExecutionContext) -> None: ...
+
+
+async def migrate_kb(context: ExecutionContext, kbid: str) -> None:
+    async with context.kv_driver.transaction(read_only=True) as txn:
+        vectorsets_count = len([vs async for vs in datamanagers.vectorsets.iter(txn, kbid=kbid)])
```
javitonino, line 54:
```suggestion
        vectorsets_count = len(list(datamanagers.vectorsets.iter(txn, kbid=kbid)))
```

nucliadb/src/migrations/0024_assign_models_to_kbs.py
```diff
@@ -0,0 +1,112 @@
...
+    if vectorsets_count > 0:
+        logger.info("Skipping KB with vectorsets already populated", extra={"kbid": kbid})
+        return
+
+    learning_config = await learning_proxy.get_configuration(kbid)
+    if learning_config is None:
+        logger.warning(f"KB has no learning config", extra={"kbid": kbid})
```
javitonino, line 61: This marks the migration as successful and we'll need to check the logs to see if something is needed. The alternative would be to raise an exception that will be more noisy.

Is this what we want? I guess it is, but just checking.

nucliadb/src/migrations/0024_assign_models_to_kbs.py
```diff
@@ -0,0 +1,112 @@
...
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""Migration #24
+
+Vectorsets are coming and we need to be ready at nucliadb. Vector index config
+shouldn't be stored anymore in the `Shards` protobuffer, we need to migrate to
```
javitonino, line 24: I don't see you cleaning the shards protobuffer. Is this intended? Do you plan to clean at a later date?

nucliadb/src/migrations/0024_assign_models_to_kbs.py
jotare: Does this work? `iter` is an async iterator

nucliadb/src/migrations/0024_assign_models_to_kbs.py
jotare: I think we can check manually. I've already checked in all our envs and it's rare that happens. If you want to be more noisy though, we can raise an exception

nucliadb/src/migrations/0024_assign_models_to_kbs.py
jotare: Yeah, just in case, we can clean that later

nucliadb/src/migrations/0024_assign_models_to_kbs.py
javitonino: Ah yeah, maybe not then :sweat:

javitonino (approved):

PR #2363, opened 2024-07-31
Fail fast in CI pytest

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/Makefile
+++ b/nucliadb/Makefile
@@ -47,6 +47,7 @@ test-%:
 
 test-cov-%:
 	$(PYTEST) \
+		-x \
 		--cov=nucliadb \
 		--cov-config=../.coveragerc \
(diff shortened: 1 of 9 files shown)


nucliadb/Makefile
```diff
@@ -47,6 +47,7 @@ test-%:
 
 test-cov-%:
 	$(PYTEST) \
+		-x \
```
javitonino, line 50: I don't like this here since in CI these tests run in parallel and are then retried serially. We expect failures in the parallel phase, so putting this here means that the first test to fail in parallel will make it to immediately go to serial mode and will make tests slower.

jotare: Oh true, I'll remove it

lferran (approved):

PR #2371, opened 2024-08-02
Pinecone improvements

lferran:
### Description
- Refactor index creation/deletion logic into manager
- Change how index names are generated: random ids now!
- Add integration test for ingestion
- Handle vectorsets logic better
- Add safe-guard for indexing on the event of external provider down
- Show external index provider type on GET kb
- Implement counts for external index kbs 

### How was this PR tested?
Integration and unit tests


Ciniesta (approved):
Looks good to me! 
(hopefully I've not missed anything :)  )

PR #2392, opened 2024-08-09
fix gcs tus writer

drf7:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/writer/tus/gcs.py
+++ b/nucliadb/src/nucliadb/writer/tus/gcs.py
@@ -126,9 +126,10 @@ class GCloudBlobStore(BlobStore):
 
         if self.json_credentials is not None and self.json_credentials.strip() != "":
             self.json_credentials_file = os.path.join(tempfile.mkdtemp(), "gcs_credentials.json")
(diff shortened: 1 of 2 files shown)


nucliadb/src/nucliadb/writer/tus/gcs.py
```diff
@@ -123,8 +125,12 @@ async def initialize(
         self._credentials = None
 
         if self.json_credentials is not None and self.json_credentials.strip() != "":
-            self._credentials = service_account.Credentials.from_service_account_info(
-                base64.b64decode(self.json_credentials).decode("utf-8"), scopes=SCOPES
+            self.json_credentials_file = os.path.join(tempfile.mkdtemp(), "gcs_credentials.json")
```
lferran, line 128: what about using context managers, to not leave anything open/created

nucliadb/src/nucliadb/writer/tus/gcs.py
```diff
@@ -123,8 +125,13 @@ async def initialize(
...
```
lferran, line 128: tmpfile can also be used in context manager mode I think

PR #2402, opened 2024-08-19
Validate content type

lferran:
### Description
- Validate content types on uploads
- Do not update resource's basic icon with content types that are not valid

### How was this PR tested?
Unit and integration tests


nucliadb/tests/standalone/integration/test_upload_download.py
```diff
@@ -220,3 +222,209 @@ async def test_tus_upload_handles_unknown_upload_ids(
...
+    )
+    assert resp.status_code == 415
+    error_detail = resp.json().get("detail")
+    assert error_detail == "Unsupported content type: invalid-content-type"
+
+
+@pytest.mark.parametrize(
+    "content_type",
+    [
+        "application/epub+zip",
```
lferran, line 277: this list has been obtained from prod

PR #2410, opened 2024-08-22
Neighbouring Paragraphs RAG Strategy

lferran:
### Description
Add a new rag strategy for the `/ask` endpoint:

```json
{
  "name": "neighbouring_paragraphs",
  "before": 3,
  "after": 0,
}
```
That allows to include sorrounding paragraphs to the ones matched on the retrieval step in the context sent to the LLM

Additionally, the PR:

- Refactors a bit the resource and extracted text cache implementation so that we only retrieve the resource and the texts once per request! (using context vars)
- Improves a bit the ids parsing logic

### How was this PR tested?
Unit and integration tests

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/find.py
+++ b/nucliadb/src/nucliadb/search/api/v1/find.py
@@ -30,6 +30,7 @@ from nucliadb.models.responses import HTTPClientError
 from nucliadb.search import predict
 from nucliadb.search.api.v1.router import KB_PREFIX, api
 from nucliadb.search.api.v1.utils import fastapi_query
+from nucliadb.search.search import cache
 from nucliadb.search.search.exceptions import InvalidQueryError
 from nucliadb.search.search.find import find
 from nucliadb.search.search.utils import min_score_from_query_params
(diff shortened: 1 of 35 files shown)


nucliadb/src/nucliadb/search/api/v1/find.py
```diff
@@ -192,9 +193,12 @@ async def _find_endpoint(
 ) -> Union[KnowledgeboxFindResults, HTTPClientError]:
     try:
         maybe_log_request_payload(kbid, "/find", item)
-        results, incomplete, _ = await find(kbid, item, x_ndb_client, x_nucliadb_user, x_forwarded_for)
-        response.status_code = 206 if incomplete else 200
-        return results
+        with cache.request_caches():
```
jotare, line 196: Is it really necessary to be explicit about caches on each search endpoint? It feels is repeating code and we always want to use caches.

jotare (approved):

nucliadb/src/nucliadb/search/api/v1/find.py
lferran: will change it to a middleware!

PR #2415, opened 2024-08-27
Consider vectorsets for rollover shards

jotare:
### Description
Avoid rollover-vectorset combo breaking everything :)

### How was this PR tested?


sunbit (approved):

lferran (commented):
Can you add an integration test for this?
it's pretty important to make sure we don't have any regressions

PR #2416, opened 2024-08-28
Refactor shards to be vectorset compatible

jotare:
### Description
- Remove fields related with vector index from `/kb/{kbid}/shards` endpoint. As it isn't exposed and used only internally, we don't need to maintain Bw/c
- Move KB shards model to `internal`, as it's for an internal endpoint, we don't need to expose it. Same Bw/c reasoning as above

### How was this PR tested?
Existing tests

### DIFF
--- a/nucliadb_models/src/nucliadb_models/search.py
+++ b/nucliadb_models/src/nucliadb_models/search.py
@@ -31,7 +31,6 @@ from nucliadb_models.metadata import RelationType, ResourceProcessingStatus
 from nucliadb_models.resource import ExtractedDataTypeName, Resource
 from nucliadb_models.security import RequestSecurity
 from nucliadb_models.utils import DateTime
-from nucliadb_models.vectors import SemanticModelMetadata, VectorSimilarity
 from nucliadb_protos.audit_pb2 import ClientType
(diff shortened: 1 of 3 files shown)


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -31,17 +30,15 @@
 from nucliadb_models.resource import ExtractedDataTypeName, Resource
 from nucliadb_models.security import RequestSecurity
 from nucliadb_models.utils import DateTime
-from nucliadb_models.vectors import SemanticModelMetadata, VectorSimilarity
 from nucliadb_protos.audit_pb2 import ClientType
 from nucliadb_protos.nodereader_pb2 import DocumentScored, OrderBy
 from nucliadb_protos.nodereader_pb2 import ParagraphResult as PBParagraphResult
 from nucliadb_protos.utils_pb2 import RelationNode
-from nucliadb_protos.writer_pb2 import ShardObject as PBShardObject
-from nucliadb_protos.writer_pb2 import Shards as PBShards
```
lferran, line 40: can you include the import to internal just to make sure we're bw/compat?

jotare: Sure, added!

lferran (approved):

PR #2422, opened 2024-09-02
Add mimetype / field and splits of qas

bloodbare:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -208,27 +208,50 @@ async def set_error(self, error: Error) -> None:
             error=error,
         )
 
-    async def get_question_answers(self) -> Optional[QuestionAnswers]:
-        if self.question_answers is None:
+    async def get_question_answers(self, force=False) -> Optional[FieldQuestionAnswers]:
+        if self.question_answers is None or force:
             sf = self.get_storage_field(FieldTypes.QUESTION_ANSWERS)
-            payload = await self.storage.download_pb(sf, QuestionAnswers)
+            payload = await self.storage.download_pb(sf, FieldQuestionAnswers)
```
jotare, line 214: Could you add a warning here if the operation fails. We can use it to find and delete any KB with the old Q&A implementation

nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -208,27 +208,50 @@ async def set_error(self, error: Error) -> None:
...
         sf = self.get_storage_field(FieldTypes.QUESTION_ANSWERS)
 
-        if payload.HasField("file"):
-            raw_payload = await self.storage.downloadbytescf(payload.file)
-            pb = QuestionAnswers()
-            pb.ParseFromString(raw_payload.read())
-            raw_payload.flush()
-            self.question_answers = pb
+        if actual_payload is None:
+            # Its first extracted text
```
jotare, line 232: Copy-pasted comment

nucliadb_models/src/nucliadb_models/common.py
```diff
@@ -271,3 +277,7 @@ class Answer(BaseModel):
 class QuestionAnswer(BaseModel):
     question: Question
     answers: List[Answer]
+
+
+class QuestionAnswers(BaseModel):
```
jotare, line 282: Do we need a new type only for a list of `QuestionAnswer`?

bloodbare: As a default pattern map the PB

jotare (approved):

PR #2424, opened 2024-09-03
Prequeries RAG strategy

lferran:
### Description

This PR extends the RAG strategies on the `/ask` endpoint with the `prequeries` one, which allows to specify up to 10 retrieval queries that will be used to craft the context for the LLM. The API allows to set weights to each of these queries to give more relevance to some queries than others.

Example payload:

```json
{
    "query": "Is Sebastian Iwasaki older than Julie Dubela?",
    "features": [
        "semantic"
    ],
    "debug": true,
    "rag_strategies": [
        {
            "name": "prequeries",
            "queries": [
                {
                    "request": {
                        "query": "Julie Dubela",
                        "min_score": 0.7
                    },
                    "weight": 1
                },
                {
                    "request": {
                        "query": "Sebastian Iwasaki",
                        "min_score": 0.7
                    },
                    "weight": 1
                }
            ]
        }
    ]
}
```

### How was this PR tested?
Integration, unit and local tests


nucliadb/src/nucliadb/search/search/find_merge.py
```diff
@@ -115,31 +115,13 @@ async def set_resource_metadata_value(
...
-        for key in self.boosted_items:
-            if key in returned:
-                continue
-            returned.add(key)
-            yield key
-
-        for key in self.items:
-            if key in returned:
-                continue
-            returned.add(key)
```
lferran, line 142: this wasn't being used in the codebase, so I removed it and simplified the class

nucliadb/src/nucliadb/search/search/find_merge.py
```diff
@@ -115,31 +115,13 @@ async def set_resource_metadata_value(
...
+        for key, _ in sorted(self.items.items(), key=lambda value: value[1], reverse=True):
```
jotare, line 124: Maybe we can remove the `Orderer` class, as there's only one ordering algorithm, a function may be enough

nucliadb/src/nucliadb/search/settings.py
```diff
@@ -36,6 +36,11 @@ class Settings(DriverSettings):
         title="Slow node query log threshold",
         description="The threshold in seconds for logging slow node queries",
     )
+    prequeries_max_parallel: int = Field(
```
jotare, line 39: Just wondering, should we implement a semaphore for search in general? I feel we have a lot of semaphores with different numbers per search and it's kind of difficult to know how many asyncio tasks a search is spawning

nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -721,7 +722,9 @@ class PromptContextBuilder:
     def __init__(
         self,
         kbid: str,
-        find_results: KnowledgeboxFindResults,
+        main_results: KnowledgeboxFindResults,
+        prequeries_results: Optional[list[PreQueryResult]] = None,
+        main_query_weight: float = 1.0,
```
jotare, line 727: I think this value should be more visible to the user, either with documentation or with a parameter.

Pre-query weights are a percentage, so 1 and 2 is the same as 10 and 20. But the result order will be different for the main query depending on the pre-query weights

lferran: there is a parameter in fact! I also add param documentation. let me know if its not clear enough.

and yes, the weight of the main query is just like any other prequery

nucliadb/src/nucliadb/search/settings.py
lferran: IMO it's fine to have different semaphores. Each task is different in nature: it's not the same querying to pinecone than fetching metadata from postgres, so I think it does make sense to have different settings.

nucliadb/src/nucliadb/search/search/find_merge.py
lferran: yup, agree!

lferran: removed

nucliadb/src/nucliadb/search/search/chat/prompt.py
jotare: Oh I missed it, sorry!

jotare (approved):

nucliadb/src/nucliadb/search/settings.py
jotare: I was thinking more about the overall number of tasks a search can create. Anyway, not related with the PR

PR #2433, opened 2024-09-05
Keyword Filtering

lferran:
### Description
Adds the ability to narrow down the text block retrieval step to only those documents that match the keyword filter expression.

Examples:

```json
{
    "query": "What is the average temperature?",
    "keyword_filters": ["Sevilla"]
}
```

This unlocks use-cases where users may want to filter by specific keywords that are not recognised as NERs by Nuclia because they are some custom ids or nomenclature (like product ids, or others).

More complex filtering use cases can also be achieved with the usual filtering logical expressions, for instance:

```json
{
    "query": "Which is the employee with the highest salary?",
    "keyword_filters": [
      {"any": ["Pepito", "Palotes", "Juanito"]}
    ]
}
```

### How was this PR tested?
Integration tests


nucliadb_texts2/tests/test_search.rs
```diff
@@ -97,13 +98,14 @@ fn test_prefilter_not_search() {
         paragraph_labels: HashSet::with_capacity(0),
     };
     let query = "{ \"not\": { \"literal\": \"/l/mylabel\" } }";
-    let expression = nucliadb_core::query_language::translate(query, &context).unwrap();
+    let expression = nucliadb_core::query_language::translate(Some(query), None, &context).unwrap();
```
jotare, line 101: Could be nice to add some test in the Rust side too, instead of setting the keyword filter always to `None`

nucliadb_core/src/query_language/mod.rs
```diff
@@ -186,41 +186,78 @@ fn split_mixed(expression: BooleanExpression, context: &QueryContext) -> NodeRes
...
-    let Json::Object(subexpressions) = serde_json::from_str(query)? else {
-        #[rustfmt::skip] return Err(node_error!(
-            "Unexpected query format {query}"
-        ));
-    };
-
-    let mut prefilter_queries = Vec::new();
+pub fn translate(
+    labels_query: Option<&str>,
+    keywords_query: Option<&str>,
```
jotare, line 196: Why are you passing this parameters directly instead of inside `QueryContext`?

nucliadb_core/src/query_language/mod.rs
```diff
@@ -186,41 +186,78 @@ fn split_mixed(expression: BooleanExpression, context: &QueryContext) -> NodeRes
...
-            prefilter_queries.push(splitted.fields_only);
-            search_queries.push(splitted.paragraphs_only);
-        } else if translation_report.has_field_labels {
-            prefilter_queries.push(translation_report.expression);
-        } else if translation_report.has_paragraph_labels {
-            search_queries.push(translation_report.expression);
+    // Translate labels query. Some of the labels are paragraph-type labels that need to be passed a search query.
+    if let Some(labels_query) = labels_query {
+        if !labels_query.is_empty() {
+            eprintln!("Labels query parsing: {labels_query}");
```
jotare, line 206: Use `tracing` macros instead of `eprintln`

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1079,6 +1079,19 @@ class ChatRequest(BaseModel):
         title="Filters",
         description="The list of filters to apply. Filtering examples can be found here: https://docs.nuclia.dev/docs/rag/advanced/search/#filters",  # noqa: E501
     )
+    keyword_filters: Union[list[str], list[Filter]] = Field(
+        default=[],
+        title="Keyword filters",
+        description=(
+            "List of keyword filter expressions to apply to the retrieval step. "
+            "The text block search will only be performed on the documents that contain the specified keywords. "
```
jotare, line 1087: Could you make more clear than keyword can be a sentence? Add a longer example or a comment on it. It feels that each term should be a single word

nucliadb_core/src/query_language/mod.rs
lferran: sorry, this is a leftover from when I was debugging. I will delete it

nucliadb_core/src/query_language/mod.rs
lferran: I didn't want to refactor what was already done

nucliadb_models/src/nucliadb_models/search.py
lferran: Ok! I will add an example with a name + surname

jotare (approved):

PR #2439, opened 2024-09-09
Keyword filters polishments

lferran:
### Description
- Stricter validation, to avoid confusion of users (special characters are not supported)
- Fixed tantivy query parsing, to make all supported use cases work
- A bit of renaming 🙈 

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/filters.py
+++ b/nucliadb/src/nucliadb/search/search/filters.py
@@ -118,7 +118,7 @@ def is_paragraph_labelset_kind(labelset_id: str, classification_labels: knowledg
         return False
 
 
-def flat_filter_literals(filters: Union[list[str], dict[str, Any]]) -> list[str]:
+def flatten_filter_literals(filters: Union[list[str], dict[str, Any]]) -> list[str]:
     if isinstance(filters, list):
         return filters
     else:
--- a/nucliadb/src/nucliadb/search/search/filters.py
+++ b/nucliadb/src/nucliadb/search/search/filters.py
@@ -118,32 +118,32 @@ def is_paragraph_labelset_kind(labelset_id: str, classification_labels: knowledg
         return False
 
 
-def flat_filter_labels(filters: Union[list[str], dict[str, Any]]) -> list[str]:
+def flat_filter_literals(filters: Union[list[str], dict[str, Any]]) -> list[str]:
     if isinstance(filters, list):
         return filters
     else:
-        return list(iter_filter_labels_expression(filters))
+        return list(iter_filter_expression_literals(filters))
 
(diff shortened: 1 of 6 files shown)


nucliadb/src/nucliadb/search/search/filters.py
```diff
@@ -118,32 +118,32 @@ def is_paragraph_labelset_kind(labelset_id: str, classification_labels: knowledg
         return False
 
 
-def flat_filter_labels(filters: Union[list[str], dict[str, Any]]) -> list[str]:
+def flat_filter_literals(filters: Union[list[str], dict[str, Any]]) -> list[str]:
```
jotare, line 121: What about `flatten` instead of `flat`

nucliadb_texts2/src/query_io.rs
```diff
@@ -19,46 +19,42 @@
...
-fn translate_literal_to_text_query(literal: &str, schema: &TextSchema) -> Box<dyn Query> {
-    // If the literal has spaces, convert them to multiple queries with an AND operator
-    if literal.contains(' ') {
-        let mut operands: Vec<(Occur, Box<dyn Query>)> = Vec::new();
-        for word in literal.split_whitespace() {
-            let term = Term::from_field_text(schema.text, &word.to_lowercase());
-            operands.push((Occur::Must, Box::new(TermQuery::new(term, IndexRecordOption::Basic))));
-        }
-        return Box::new(BooleanQuery::new(operands));
+fn translate_keyword_to_text_query(literal: &str, schema: &TextSchema) -> Box<dyn Query> {
```
javitonino, line 32: This inverted conditions is maybe a bit confusing? I'd also put the alternative branch on an else, since they are the same level of importance. And we can reuse the term construction. So something like this may be nicer. What do you think?

```rust
    let terms = literal.split_whitespace().map(|w| Term::from_field_text(schema.text, &w.to_lowercase())).collect_vec();
    if literal.len() == 1 {
        Box::new(TermQuery::new(terms[0], IndexRecordOption::Basic))
    } else {
        Box::new(PhraseQuery::new(terms.collect()))
    }
```

javitonino (approved):
Cool :sparkles: 

Just a small suggestion, take it or leave it.

nucliadb_texts2/src/query_io.rs
lferran: I like this code more indeed! but does not complie 😢 I'll fight a bit more with the compiler and see if I succeed.
Otherwise I'll leave it as is ok?

lferran: made it work! 🚀

PR #2441, opened 2024-09-10
Filter by field in semantic search aswell

lferran:
### Description
This was never implemented, and it must be as we're offering it in the API

### How was this PR tested?
Integration tests at the node


nucliadb_vectors/src/formula/mod.rs
```diff
@@ -74,6 +77,14 @@ impl AtomClause {
 
                 field_set.contains(std::str::from_utf8(&key[0..end_pos]).unwrap())
             }
+            Self::KeyField((field_type, field_name)) => {
+                let key = retriever.get_key(x);
+                let mut key_parts = key.split(|b| *b == b'/');
+                key_parts.next(); // Skip resource_id
+                let ftype = std::str::from_utf8(key_parts.next().unwrap()).unwrap();
+                let fname = std::str::from_utf8(key_parts.next().unwrap()).unwrap();
```
jotare, line 85: Not to tackle now, but those unwraps in the code are a bit worrying IMO. There's no validation of id but you're assuming a certain format. I thing we should be validating ids more explicitly so it's safer to make those assumptions

nucliadb_node/tests/test_search_fields_filtering.rs
```diff
@@ -0,0 +1,154 @@
...
+        ..Default::default()
+    };
+    let results = reader.search(Request::new(search_request)).await?.into_inner();
+
+    assert!(results.vector.is_some());
+    let vector_results = results.vector.unwrap();
+    assert_eq!(vector_results.documents.len(), 0);
+
+    // Search filtering with a real field, to check that the vector is returned
+    let query_vector2 = vec![17.0 / magnitude; VECTOR_DIMENSION];
```
jotare, line 72: Why aren't you using `query` again? You could `.clone()` it in the previous call if you needed to

nucliadb_node/tests/test_search_fields_filtering.rs
```diff
@@ -0,0 +1,154 @@
...
+            text: "Dummy text".to_string(),
+            labels: vec![],
+        },
+    );
+
+    let mut field_paragraphs = HashMap::new();
+    for i in 1..=20 {
+        let mut sentences = HashMap::new();
+        sentences.insert(
+            format!("{rid}/{field_id}/{i}/paragraph-{i}"),
```
jotare, line 115: Aren't vector ids more like `{rid}/{field_id}/{index}/{vstart}-{vend}`? I know it doesn't matter for the test but would be more correct

nucliadb_vectors/src/formula/mod.rs
javitonino: I think we should tackle this now, for the iteration: it's pretty easy.

```rust
                if let Some(ftype) = key_parts.next() {
                    if let Some(fname) = key_parts.next() {
                        let ftype_str = std::str::from_utf8(ftype).unwrap();
                        let fname_str = std::str::from_utf8(fname).unwrap();
                        return ftype == field_type && fname == field_name;
                    }
                }
                false
```

I think the unwraps in the UTF8 decoding are fine, if those fail, the segment is corrupted and a noisy failure is in order.

nucliadb_vectors/src/service/reader.rs
```diff
@@ -93,6 +93,17 @@ impl VectorReader for VectorReaderService {
...
         }
 
+        if !request.field_filters.is_empty() {
+            for field in request.field_filters.iter() {
+                // split "a/title" into "a" and "title"
+                let parts = field.split('/').collect::<Vec<&str>>();
+                let field_type = parts.first().unwrap_or(&"").to_string();
+                let field_name = parts.get(1).unwrap_or(&"").to_string();
+                let clause = AtomClause::key_field(field_type, field_name);
+                formula.extend(clause);
```
javitonino, line 103: Careful with this, this makes an `AND` for all field types. So if multiple fields are specified, it won't match anything at all. You'll need to create a `CompoundClause::new(BooleanOperator::Or, field_clauses)` and add that to the formula instead.

javitonino (changes requested):
This won't work if the request passes multiple fields.

nucliadb_vectors/src/service/reader.rs
javitonino: Maybe add a test as well?

lferran: oh good catch! totally, I will add a test for this

lferran: fixed!

javitonino (approved):
Looking perfect :man_cook:

PR #2446, opened 2024-09-11
Make dummy learning less dummy

jotare:
### Description
Dummy utilities are sometimes too dummy. Make them a bit more useful and aware of things

### How was this PR tested?


nucliadb/src/nucliadb/writer/api/v1/export_import.py
```diff
@@ -75,6 +83,52 @@ async def start_kb_export_endpoint(request: Request, kbid: str):
...
+    context = get_app_context(request.app)
+
+    # Read stream and parse learning configuration
+    stream_reader = ExportStreamReader(request.stream())
+    learning_config, leftover_bytes = await stream_reader.maybe_read_learning_config()
+    if learning_config is None:
+        raise Exception(
+            "Trying to import an export missing learning config. Try using import on an existing KB"
+        )
+    print("Import learning config:", learning_config)
```
lferran, line 105: print!

nucliadb/src/nucliadb/learning_proxy.py
```diff
@@ -418,3 +416,69 @@ def _handle_request(self, *args: Any, **kwargs: Any) -> httpx.Response:
...
+    async def get_configuration(self, kbid: str) -> Optional[LearningConfiguration]:
+        return self.in_memory_configs.get(kbid, None)
+
+    async def set_configuration(self, kbid: str, config: dict[str, Any]) -> LearningConfiguration:
+        parsed_config = LearningConfiguration.model_validate(config)  #
+        self.in_memory_configs[kbid] = parsed_config
+        return parsed_config
+
+    async def delete_configuration(self, kbid: str) -> None:
+        self.in_memory_configs.pop(kbid, None)
```
lferran, line 477: This is not stored as part of the global state anywhere right?
So from one request to the next, the config won't be "persisted" even if the process hasn't restarted

nucliadb/src/nucliadb/writer/api/v1/export_import.py
```diff
@@ -75,6 +83,52 @@ async def start_kb_export_endpoint(request: Request, kbid: str):
         return CreateExportResponse(export_id=export_id)
 
 
+@only_for_onprem
+@api.post(
+    f"/{KBS_PREFIX}/import",
+    summary="Create a KB from an export and import its content",
+    tags=["Knowledge Boxes"],
+    # response_model=CreateImportResponse,
```
lferran, line 91: can you add the param that excludes the endpoint def from the docs?

nucliadb/src/nucliadb/writer/api/v1/export_import.py
```diff
@@ -75,6 +83,52 @@ async def start_kb_export_endpoint(request: Request, kbid: str):
...
+            yield leftovers
+        async for chunk in stream:
+            yield chunk
+
+    import_id = uuid4().hex
+    await importer.import_kb(
+        context=context, kbid=kbid, stream=stream_with_leftovers(leftover_bytes, request.stream())
+    )
+
+    return (kbid, slug, import_id)
```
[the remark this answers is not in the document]
jotare: Yes, just pushed it

nucliadb/src/nucliadb/writer/api/v1/export_import.py
jotare: Sure

nucliadb/src/nucliadb/writer/api/v1/export_import.py
```diff
@@ -75,6 +84,53 @@ async def start_kb_export_endpoint(request: Request, kbid: str):
...
+@requires_one([NucliaDBRoles.MANAGER, NucliaDBRoles.WRITER])
+@version(1)
+async def kb_create_and_import_endpoint(request: Request) -> NewImportedKbResponse:
+    context = get_app_context(request.app)
+
+    # Read stream and parse learning configuration
+    stream_reader = ExportStreamReader(request.stream())
+    learning_config, leftover_bytes = await stream_reader.maybe_read_learning_config()
+    if learning_config is None:
+        raise Exception(
```
lferran, line 104: Return proper http error with a message in the response?

nucliadb/src/nucliadb/learning_proxy.py
jotare: Ups! Changed for a global LRU

nucliadb/src/nucliadb/writer/api/v1/export_import.py
```diff
@@ -75,6 +84,55 @@ async def start_kb_export_endpoint(request: Request, kbid: str):
...
+
+    # Import contents to the new KB
+
+    async def stream_with_leftovers(leftovers: bytes, stream: AsyncGenerator[bytes, None]):
+        if len(leftovers) > 0:
+            yield leftovers
+        async for chunk in stream:
+            yield chunk
+
+    await importer.import_kb(
```
lferran, line 126: I think we should not be importing synchronously here, and we should return the import_id for the client to check the status

jotare: Why not? I'm following the pattern of the regular import where standalone does this and hosted uploads to a bucket and then imports

lferran (approved):

PR #2451, opened 2024-09-12
Improve keyword filtering to use the same tokeinzer as when indexing

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_texts2/src/query_io.rs
+++ b/nucliadb_texts2/src/query_io.rs
@@ -21,7 +21,7 @@ use crate::schema::TextSchema;
 use nucliadb_core::query_language::{BooleanExpression, BooleanOperation, Operator};
 use tantivy::query::{AllQuery, BooleanQuery, Occur, PhraseQuery, Query, TermQuery};
 use tantivy::schema::{Facet, IndexRecordOption};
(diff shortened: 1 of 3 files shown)


nucliadb_texts2/src/query_io.rs
```diff
@@ -30,9 +31,14 @@ fn translate_label_to_facet_query(literal: &str, schema: &TextSchema) -> Box<dyn
 }
 
 fn translate_keyword_to_text_query(literal: &str, schema: &TextSchema) -> Box<dyn Query> {
-    let terms: Vec<Term> =
-        literal.split_whitespace().map(|w| Term::from_field_text(schema.text, &w.to_lowercase())).collect();
-
+    // Tokenize the literal in the same way we tokenize the text field at indexing time
+    let tokenizer = TextAnalyzer::from(SimpleTokenizer).filter(LowerCaser);
```
javitonino, line 35: I don't like this too much, I think it would be better to use the index default tokenizer so if its definition changes for some reason, this is still accurate.

Something like: `let tokenizer = TokenizerManager::default().get("default").unwrap();`

lferran: fixed!

PR #2466, opened 2024-09-18
Add citation threshold on ask

carlesonielfa:
### Description
Requires https://github.com/nuclia/learning/pull/1660 To be merged first

### How was this PR tested?


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -784,6 +784,12 @@ class ChatModel(BaseModel):
         default=None, description="Optional custom prompt input by the user"
     )
     citations: bool = Field(default=False, description="Whether to include the citations in the answer")
+    citation_threshold: Optional[float] = Field(
+        default=None,
+        description="If citations is set to True, this will be the similarity threshold. Value between 0 and 1, lower values will produce more citations. If not set, it will be set to the optimized threshold found by Nuclia.",  # noqa
```
lferran, line 789: The description is a bit unclear to me. Against which score is this threshold compared?

carlesonielfa: The score given by the reranker that we use to compute citations (our procedure is to use the reranker to compare parts of the answer to the retrieved paragraphs). 

I wanted to give some info but not that much since we don't specify how we compute citations and because of that mentioning the reranker could be confusing.

I also think that the wording could be improved, is something like this better?: 
`If citations is True, this sets the similarity threshold (0 to 1). Lower values generate more citations. If not specified, Nuclia's optimized threshold is used.`

lferran: Much better. Are we returning the reranked value to the user though? How will they know which value to use?

lferran (approved):

PR #2467, opened 2024-09-18
Text fastfields

javitonino:
### Description

### How was this PR tested?


nucliadb_texts3/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
```
javitonino, line 1: The best way to review nucliadb_texts3 is to download this branch and diff against nucliadb_texts2: `diff -ru nucliadb_texts2/ nucliadb_texts3`

nucliadb_node/src/settings.rs
```diff
@@ -258,6 +288,9 @@ pub struct EnvSettings {
     pub s3_indexing_bucket: String,
     pub s3_endpoint: Option<String>,
     pub azure_account_url: Option<String>,
+
+    // Mr.Flagly
+    pub flag_settings_url: Option<String>,
```
lferran, line 293: note: we need to make sure this env var is propagated on `nucliadb_shared` or the node chart, otherwise the default flag will be rollout:100 (from the DEFAULT_FLAGS json)

nucliadb_texts3/Cargo.toml
lferran: I think we need to be able to filter by field type aswell, so maybe you could split the nucliadb-field parts into two separate tantivy fields? one for the type (a, u, t, f...) and another one for the key part

wdyt?

javitonino: Oh, I already changed it back to facets. I'm pretty sure nobody uses that feature, but well.

nucliadb_node/src/settings.rs
javitonino: Yup, it's already done (via nucliadb-shared).

Something else I just realized I want to add, is to pass the environment to the FF check.

jotare (commented):
You forgot to update the Dockerfile.node to add nucliadb_texts3

javitonino:
> You forgot to update the Dockerfile.node to add nucliadb_texts3

Ah thanks! :heart:  I thought I got rid of all of that nonsense with the CI refactor, forgot to check the dockefile. Fixed!

lferran (approved):

PR #2471, opened 2024-09-18
Fix Pinecone /find response field ids

jotare:
### Description
Pinecone /find response was returning field ids as `{rid}/{type}/{key}` while index node results return without the resource id

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/prompt.py
+++ b/nucliadb/src/nucliadb/search/search/chat/prompt.py
@@ -524,10 +524,7 @@ async def get_paragraph_text_with_neighbours(
     ) -> tuple[ParagraphId, str]:
         return pid, await get_paragraph_text(
             kbid=kbid,
-            rid=pid.field_id.rid,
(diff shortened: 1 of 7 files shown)


nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -524,10 +524,7 @@ async def _get_paragraph_text(
     ) -> tuple[ParagraphId, str]:
         return pid, await get_paragraph_text(
             kbid=kbid,
-            rid=pid.field_id.rid,
-            field=pid.field_id.full(),
```
jotare, line 528: Here we pass the field as `full()` and https://github.com/nuclia/nucliadb/blob/349fa0c495a6c544b58c11b60ce1e5ef96be1a89/nucliadb/src/nucliadb/search/search/results_hydrator/base.py#L104 is using this id in the find response

lferran (approved):
Thanks for the ids logic cleanup

PR #2476, opened 2024-09-20
Faster delete resources

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/maindb/pg.py
+++ b/nucliadb/src/nucliadb/common/maindb/pg.py
@@ -107,6 +107,11 @@ class DataLayer:
             async with self.connection.cursor() as cur:
                 await cur.execute("DELETE FROM resources WHERE key = %s", (key,))
 
+    async def delete_by_prefix(self, prefix: str) -> None:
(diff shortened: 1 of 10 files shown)


nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -107,6 +107,11 @@ async def delete(self, key: str) -> None:
             async with self.connection.cursor() as cur:
                 await cur.execute("DELETE FROM resources WHERE key = %s", (key,))
 
+    async def delete_batch(self, keys: list[str]) -> None:
```
javitonino, line 110: This could be combined with `keys()` to do it in a single query like: `cur.execute('DELETE FROM resources WHERE key LIKE %s', (prefix + '%',))` (and passing the prefix into this, rather than the list of keys). In the case of the local storage, it could move the iteration over keys inside the method or do a `rmdir`

javitonino (approved):

nucliadb/src/nucliadb/common/maindb/pg.py
lferran: thanks! addressed

PR #2489, opened 2024-09-24
Fix vectorsets dimensions

javitonino:
Fixes some vectorsets that were created with incorrect dimensions. This must be done before the rollover, or the resulting shards will fail due to "inconsistent dimensions".

### DIFF
--- a/nucliadb/src/migrations/0027_rollover_texts3.py
+++ b/nucliadb/src/migrations/0027_rollover_texts3.py
@@ -59,8 +59,15 @@ async def maybe_fix_vector_dimensions(context: ExecutionContext, kbid: str) -> N
         if vectorset.vectorset_index_config.vector_dimension != 0:
             return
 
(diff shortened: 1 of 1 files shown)


nucliadb/src/migrations/0027_rollover_texts3.py
```diff
@@ -23,12 +23,44 @@
...
+            # If multiple vectorsets, they are new shards created correctly, we can safely skip it
+            logger.warning(f"KB has {len(vectorsets)} vectorsets, skipping...", extra={"kbid": kbid})
+            return
+        vectorset = vectorsets[0][1]
+
+        # Correct value, skip
+        if vectorset.vectorset_index_config.vector_dimension != 0:
+            return
+
+        logger.info(f"Fixing KB vectorset dimension", extra={"kbid": kbid})
```
jotare, line 62: You could add the difference, just to record it in the log

jotare (approved):

PR #2503, opened 2024-09-30
add user metadata ask field

drf7:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/search/api/v1/find.py
```diff
@@ -126,6 +126,16 @@ async def find_knowledgebox(
     security_groups: list[str] = fastapi_query(SearchParamDefaults.security_groups),
     show_hidden: bool = fastapi_query(SearchParamDefaults.show_hidden),
     reranker: Reranker = fastapi_query(SearchParamDefaults.reranker),
+    user_metadata: Optional[str] = Query(
```
lferran, line 129: I would rename this to be more specific , so it is clear that is just for audit
```suggestion
    audit_metadata: Optional[str] = Query(
```
wdyt?

nucliadb/src/nucliadb/search/api/v1/find.py
```diff
@@ -126,6 +126,16 @@ async def find_knowledgebox(
...
+        default=None,
+        title="User metadata",
+        description=(
```
lferran, line 132: I would also put some validation to make sure the metadata does not exceed certain size. Say... 100KB ?

nucliadb/src/nucliadb/search/api/v1/find.py
drf7: Yeah, it looks better, thanks

nucliadb/src/nucliadb/search/api/v1/find.py
drf7: We decided to remove this from GET endpoints, so now audit_metadata will only be available in POST endpoints

sunbit (approved):

nucliadb/src/nucliadb/search/api/v1/find.py
lferran: Even in a POST, we should still restrict it to a max size IMO!
As otherwise it's a way to store arbitrary data in the audit database

drf7: In the model, there are other fields, such as those of type str. Wouldn't we encounter the same issue with these fields, since a str can have an arbitrary size?
Why do we need to limit the size for this field and not for the others?

lferran (approved):

nucliadb/src/nucliadb/search/api/v1/find.py
lferran: IMO all params in an API should have limits, even if they are very big. Limits help you to scale the service and also the user to know how to deal properly with the api.

There are some params that don't have limits, and they should, but it's out of the scope of this PR to add them -- don't worry.

IMO `audit_metadata` should be limited in size (100kB or 1MB if you want), as it is arbitrary metadata that will be stored in the audit storage. This can help you reason about the right sizing of the persistant layers the audit service may need.

drf7: Limits of 10KB added, I think 10KB is enough, we can always raise them later

lferran (approved):

PR #2506, opened 2024-10-01
Cleanup /query endpoint and use predict default vectorset

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/query.py
+++ b/nucliadb/src/nucliadb/search/search/query.py
@@ -25,7 +25,6 @@ from typing import Any, Awaitable, Optional, Union
 from async_lru import alru_cache
 
 from nucliadb.common import datamanagers
-from nucliadb.common.datamanagers.vectorsets import BrokenInvariant
 from nucliadb.common.maindb.utils import get_driver
 from nucliadb.search import logger
 from nucliadb.search.predict import SendToPredictError, convert_relations
@@ -88,6 +87,7 @@ class QueryParser:
     """
 
     _query_information_task: Optional[asyncio.Task] = None
+    _get_vectorset_task: Optional[asyncio.Task] = None
     _detected_entities_task: Optional[asyncio.Task] = None
     _entities_meta_cache_task: Optional[asyncio.Task] = None
     _deleted_entities_groups_task: Optional[asyncio.Task] = None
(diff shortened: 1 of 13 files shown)


nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -463,7 +483,8 @@ async def parse_vector_search(self, request: nodereader_pb2.SearchRequest) -> bo
                         else:
                             incomplete = True
                     else:
-                        query_vector = query_info.sentence.data
+                        for query_vector in query_info.sentence.vectors.values():
```
lferran, line 486: maybe it's better to use popitem? I know it shouldn't but what if for some reason vectors is empty?

nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -403,38 +437,24 @@ def parse_paragraph_search(self, request: nodereader_pb2.SearchRequest) -> None:
...
+
+        # validate vectorset
+        async with datamanagers.with_ro_transaction() as txn:
+            if not await datamanagers.vectorsets.exists(
+                txn, kbid=self.kbid, vectorset_id=self.vectorset
+            ):
+                raise InvalidQueryError(
+                    "vectorset",
+                    f"Vectorset {self.vectorset} doesn't exist in you Knowledge Box",
+                )
```
lferran, line 456: this could be hard cached no?

nucliadb/src/nucliadb/search/search/query.py
jotare: I'd prefer not to modify the `QueryInfo` payload, but I can check for empty vector

nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -382,12 +409,19 @@ async def parse_min_score(self, request: nodereader_pb2.SearchRequest, incomplet
...
+                else:
+                    logger.warning(
+                        "Semantic threshold not found in query information, using default",
+                        extra={"kbid": self.kbid},
+                    )
             else:
                 logger.warning(
-                    "Semantic threshold not found in query information, using default",
-                    extra={"kbid": self.kbid},
+                    "Vectorset unset, using default semantic threshold", extra={"kbid": self.kbid}
```
lferran, line 424: will this log be very chatty? I assume lots of requests won't have the vectorset param provided and will just use the default no?

nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -191,18 +191,45 @@ def _get_query_information(self) -> Awaitable[QueryInfo]:
...
+
+        query_information = await self._get_query_information()
+
+        if query_information.sentence is None:
+            logger.error(
+                "Asking for a vectorset but /query didn't return one", extra={"kbid": self.kbid}
+            )
+            return self.vectorset
+
+        for vectorset in query_information.sentence.vectors.keys():
```
lferran, line 218: nit: is popitem better/safer?

nucliadb/src/nucliadb/search/search/query.py
lferran: (I mean checking the lookup for if the vectorset exists)

nucliadb/src/nucliadb/search/search/query.py
jotare: No, this is a log for either predict or the user setting the vectorset. Shouldn't happen. I'll clarify the log

nucliadb/src/nucliadb/search/search/query.py
jotare: I prefer not to modify the `query_information` object

javitonino (approved):

PR #2507, opened 2024-10-02
Resource extracted metadata endpoint

lferran:
### Description
Endpoint to get specific extracted metadata of a field.
Useful for downloading extracted text, vectors, or other metadata when they are too big to fit in a resource GET response.

### How was this PR tested?
Integration tests


lferran:
no need for new endpoints, the regular GET resource can be used

PR #2519, opened 2024-10-07
Fix rag image strategies

lferran:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -319,11 +320,11 @@ async def extend_prompt_context_with_metadata(
     for text_block_id in context.text_block_ids():
         try:
             text_block_ids.append(parse_text_block_id(text_block_id))
-        except ValueError:
+        except ValueError:  # pragma: no cover
```
jotare, line 323: If that's common, maybe you could write a test for it instead of adding a `no cover

jotare (approved):

PR #2521, opened 2024-10-07
Ask endpoint reranking

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/ask.py
+++ b/nucliadb/src/nucliadb/search/search/chat/ask.py
@@ -176,7 +176,7 @@ class AskResult:
 
     def _ndjson_encode(self, item: AskResponseItemType) -> str:
         result_item = AskResponseItem(item=item)
-        return result_item.model_dump_json(exclude_unset=False, exclude_none=True) + "\n"
+        return result_item.model_dump_json(exclude_none=True) + "\n"
 
     async def _stream(self) -> AsyncGenerator[AskResponseItemType, None]:
         # First, stream out the predict answer
@@ -333,7 +333,7 @@ class AskResult:
                 self.prompt_context, self.prompt_context_order
             )
             response.prompt_context = sorted_prompt_context
-        return response.model_dump_json(exclude_unset=True)
+        return response.model_dump_json()
 
     async def get_relations_results(self) -> Relations:
         if self._relations is None:
@@ -402,7 +402,7 @@ class NotEnoughContextAskResult(AskResult):
             answer=NOT_ENOUGH_CONTEXT_ANSWER,
             retrieval_results=self.main_results,
             status=AnswerStatusCode.NO_CONTEXT,
-        ).model_dump_json(exclude_unset=True)
+        ).model_dump_json()
 
 
 async def ask(
--- a/nucliadb/src/nucliadb/search/search/chat/ask.py
+++ b/nucliadb/src/nucliadb/search/search/chat/ask.py
@@ -34,6 +34,7 @@ from nucliadb.search.predict import (
     JSONGenerativeResponse,
     MetaGenerativeResponse,
     RephraseMissingContextError,
+    RerankGenerativeResponse,
     StatusGenerativeResponse,
     TextGenerativeResponse,
 )
@@ -54,8 +55,10 @@ from nucliadb.search.search.exceptions import (
 )
 from nucliadb.search.search.metrics import RAGMetrics
 from nucliadb.search.search.query import QueryParser
+from nucliadb.search.search.rerankers import RankedItem, apply_reranking
 from nucliadb.search.utilities import get_predict
 from nucliadb_models.search import (
+    SCORE_TYPE,
     AnswerAskResponseItem,
     AskRequest,
     AskResponseItem,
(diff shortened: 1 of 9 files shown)


nucliadb/src/nucliadb/search/search/chat/ask.py
```diff
@@ -201,6 +196,23 @@ async def _stream(self) -> AsyncGenerator[AskResponseItemType, None]:
...
+                RankedItem(id=paragraph_id, score=score, score_type=SCORE_TYPE.RERANKER)
+                for paragraph_id, score in self._reranking.context_scores.items()
+            ]
+            apply_reranking(self.main_results, reranked)
+
+        yield RetrievalAskResponseItem(results=self.main_results)
+
+        if len(self.prequeries_results) > 0:
+            item = PrequeriesAskResponseItem()
+            for index, (prequery, result) in enumerate(self.prequeries_results):
```
lferran, line 211: just a thought: should we rerank prequeries aswell?

nucliadb/src/nucliadb/search/search/rerankers.py
```diff
@@ -187,3 +145,58 @@ def get_reranker(kind: search_models.Reranker) -> Reranker:
...
+        for field_id, field in resource.fields.items():
+            for paragraph_id, paragraph in field.paragraphs.items():
+                inverted_results[paragraph_id] = (
+                    paragraph,
+                    (field_id, field),
+                    (rid, resource),
+                )
+
+    # order results by new score
+    best_matches = []
```
lferran, line 162: should this be a set to handle duplicated paragraphs (same paragraph matching different indexes)?

nucliadb/src/nucliadb/search/search/chat/ask.py
jotare: The prequeries are `FindRequest`s, so you can add a reranker option and will be reranked. However, they involve a round trip to predict, as we do them like a regular find

nucliadb/src/nucliadb/search/search/chat/ask.py
```diff
@@ -174,17 +179,7 @@ def _ndjson_encode(self, item: AskResponseItemType) -> str:
...
-
-        if len(self.prequeries_results) > 0:
-            item = PrequeriesAskResponseItem()
-            for index, (prequery, result) in enumerate(self.prequeries_results):
-                prequery_id = prequery.id or f"prequery_{index}"
-                item.results[prequery_id] = result
-            yield item
-
-        # Then stream out the predict answer
+        # First, stream out the predict answer
```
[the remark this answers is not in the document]
jotare: According to @lferran and @ebrehault, the consumers of the stream *shouldn't* be relying on stream item ordering.

As we are not assuming any order on the predict stream, we'd have to wait until the reranking chunk comes. If that's after the text, we'll loose the "real time" feeling of the response. Otherwise, we could assume an order on the predict chunks creating a dependency between repos. I don't think any of those two options are worth if we can reorder chunks.

nucliadb/src/nucliadb/search/search/rerankers.py
jotare: Sure, but if you don't mind, I'll do it in another PR and fix the one in find_merge too

nucliadb/src/nucliadb/search/search/rerankers.py
```diff
@@ -187,3 +145,58 @@ def get_reranker(kind: search_models.Reranker) -> Reranker:
...
+def _get_items_to_rerank(results: KnowledgeboxFindResults) -> list[RerankableItem]:
+    return [
+        RerankableItem(
+            id=paragraph_id,
+            score=paragraph.score,
+            score_type=paragraph.score_type,
+            content=paragraph.text,
+        )
+        for rid, resource in results.resources.items()
+        for field_id, field in resource.fields.items()
```
lferran, line 200: nit: these two could be values

```suggestion
        for resource in results.resources.values()
        for field in resource.fields.values()
```

nucliadb/src/nucliadb/search/search/chat/query.py
```diff
@@ -181,11 +182,17 @@ async def run_main_query(
...
+    if item.reranker != Reranker.PREDICT_RERANKER:
+        find_request.reranker = item.reranker
     # We don't support pagination, we always get the top_k results.
-    find_request.page_size = item.top_k
-    find_request.page_number = 0
+    find_request.top_k = item.top_k
     find_request.show_hidden = item.show_hidden
 
+    # this executes the model validators, that can tweak some fields
+    FindRequest.model_validate(find_request)
```
lferran, line 194: mh, interesting. can you elaborate a bit more?

shouln't it be:
`find_request = FindRequest.model_validate(find_request)`
?

lferran (approved):

nucliadb/src/nucliadb/search/search/chat/query.py
jotare: No, the validation modifies the object in-place.

```python
>>> from nucliadb_models.search import FindRequest
>>> req = FindRequest()
>>> req
FindRequest(..., page_size=20, top_k=None, ...)
>>> FindRequest.model_validate(req)
FindRequest(..., page_size=10, top_k=10, ...)
>>> req
FindRequest(..., page_size=10, top_k=10, ...)
```

PR #2534, opened 2024-10-11
Leave prev logic on rag image strategies

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_models/src/nucliadb_models/search.py
+++ b/nucliadb_models/src/nucliadb_models/search.py
@@ -1351,9 +1351,15 @@ If empty, the default strategy is used, which simply adds the text of the matchi
         ],
     )
     rag_images_strategies: list[RagImagesStrategies] = Field(
-        default=[],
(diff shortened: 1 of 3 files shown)


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1351,9 +1351,15 @@ class AskRequest(AuditMetadataBase):
         ],
     )
     rag_images_strategies: list[RagImagesStrategies] = Field(
-        default=[],
+        default=[ParagraphImageStrategy()],
```
jotare, line 1354: Do we want to always have this strategy as default? All `/ask` requests will use it right?

jotare (approved):

nucliadb_models/src/nucliadb_models/search.py
lferran: this was already the case before I started the changes (see the description)
just that the implementation was too implicit and, since there were no tests, I did break it

PR #2542, opened 2024-10-15
add fields needed for eml

Ciniesta:
Add fields needed for eml to FileExtractedData


nucliadb_protos/resources.proto
```diff
@@ -170,6 +170,9 @@ message FileExtractedData {
     string icon = 11;
     map<string, NestedPosition> nested_position = 12 [deprecated = true];
     map<string, NestedListPosition> nested_list_position = 13;
+    google.protobuf.Timestamp date = 14;
+    string title = 15;
+    repeated string colaborators =16;
```
lferran, line 175: what do you think about doing it more explicit?

```
message EmailMetadata {
   string subject = 0;
   Timestamp sent_date = 1;
   repeated string participants = 2;
}
```

and then

```
EmailMetadata email_metadata = 14;
```

Ciniesta: Much better approach! updated

lferran (approved):

PR #2550, opened 2024-10-18
Upgrade all deps

jotare:
### Description
I used the wrong command in https://github.com/nuclia/nucliadb/pull/2548, now it really upgrades `starlette` and all upgradable deps

### How was this PR tested?


javitonino (changes requested):
This broke a telemetry test for some reason, so better to take a second look at what's happening.

javitonino (approved):

PR #2566, opened 2024-10-22
Add origin field to relation metadata

carlesonielfa:
### Description
Add origin field to relation metadata to track where a relation is computed for Data Augmentation purposes

### How was this PR tested?


jotare (commented):
Could you use another name or a more explanatory type? I feel we're overloading the word `origin` with too many things. We already have an origin in: `Audit`, `AuditRequest`, `BrokerMessage`, `ResourceGet` and `Label` and I'm not sure all of them stand for the same concept.

jotare (approved):

PR #2568, opened 2024-10-22
Add index id to merge jobs

jotare:
### Description
Add `index_id` to `merge_jobs` table and use it to validate jobs (instead of the index from the first segment

### How was this PR tested?


nidx/nidx_types/src/lib.rs
```diff
@@ -30,6 +30,16 @@ impl From<u64> for Seq {
         Self(value as i64)
     }
 }
+impl From<i32> for Seq {
```
javitonino, line 33: This is only used for tests using seq literals. I think it's better if we remove it and change the tests instead, for example to `100i64.into()` or `Seq::from(100i64)`.

nidx/nidx_types/src/lib.rs
```diff
@@ -30,6 +30,16 @@ impl From<u64> for Seq {
...
+    fn from(value: i32) -> Self {
+        Self(value as i64)
+    }
+}
+impl From<Seq> for i64 {
```
javitonino, line 38: This is fine, although it can be easily be avoided by doing `i64::from(&seq)` instead of `i64::from(seq)`.

nidx/src/maintenance/scheduler.rs
```diff
@@ -158,23 +169,112 @@ pub async fn purge_deletions(meta: &NidxMetadata, consumer: &mut PullConsumer) -
...
+        r#"SELECT index_id, array_agg(id) AS "segment_ids!: Vec<SegmentId>" FROM segments WHERE delete_at IS NULL AND merge_job_id IS NULL AND seq < $1 GROUP BY index_id"#,
+        i64::from(before_seq),
     )
     .fetch_all(&meta.pool)
     .await?;
     for m in merges {
-        if m.len() > 3 {
-            MergeJob::create(meta, &m, oldest_confirmed_seq as i64).await?;
+        if m.segment_ids.len() > 3 {
+            MergeJob::create(meta, m.index_id.into(), &m.segment_ids, before_seq).await?;
```
javitonino, line 184: You introduced a consistency bug here. Before, this was `oldest_confirmed_seq` and now it is `before_seq = oldest_pending_seq`.

An option maybe to change the call to this function to pass `oldest_confirmed_seq` and change the query to be `<=` instead of `<`.

**Explanation**
The key concept here is that we are merging segments while also indexing new ones. Merging also involves applying all deletions and since new segments may affect older segments (they include updates for the same resource in a previous segment), it is crucial that we merge "in-order". The situation we want to avoid is the following:
- We merge segments with seqs 100 and 102, applying all deletions. The seq number of the new segment is 102.
- We index segment seq 101 that includes deletions which affect 100. However, since 100 is merged into 102, it will never see the deletions for 100.

In order to ensure this, we only merge segments older than the oldest pending seq (nats message still processing). In this case, since message 101 is still pending (not indexed = not ack), NATS will report `oldest_confirmed_seq = consumer_info.ack_floor.stream_sequence = 100`. We can merge all segments up to and including 100 (< 101 in the previous code, it could be <= 100), and the resulting seq for the merged segment will be 100.

**Note**: Since this is super-important, it would be nice to have a explicit test for it.

nidx/src/maintenance/scheduler.rs
```diff
@@ -158,23 +169,112 @@ pub async fn purge_deletions(meta: &NidxMetadata, consumer: &mut PullConsumer) -
...
+        use nidx_types::Seq;
+        use uuid::Uuid;
+
+        use super::*;
+
+        use crate::metadata::{Index, IndexId, IndexKind, NidxMetadata, Shard};
+
+        #[sqlx::test]
+        async fn test_schedule_merges_for_shard_with_single_index(pool: sqlx::PgPool) {
+            let meta = NidxMetadata::new_with_pool(pool).await.unwrap();
```
javitonino, line 207: You may want to make the tests return `Result<()>` so you can use the `?` operator instead of using so many `.unwrap()`s.

nidx/src/maintenance/scheduler.rs
```diff
@@ -158,23 +169,112 @@ pub async fn purge_deletions(meta: &NidxMetadata, consumer: &mut PullConsumer) -
...
+
+            schedule_merges(&meta, Seq::from(100)).await.unwrap();
+
+            // one job has been scheduled for the index
+            let mut jobs = vec![];
+            while let Some(job) = MergeJob::take(&meta.pool).await.unwrap() {
+                jobs.push(job)
+            }
+            assert_eq!(jobs.len(), 1);
+            assert_eq!(IndexId::from(jobs[0].index_id), index.id);
```
javitonino, line 230: Maybe add a couple of asserts:
- The merge job links to all 10 segments
- The seq from the merge job is the same as in the scheduler.

nidx/src/maintenance/scheduler.rs
```diff
@@ -158,23 +169,112 @@ pub async fn purge_deletions(meta: &NidxMetadata, consumer: &mut PullConsumer) -
...
+                Index::create(&meta.pool, shard.id, IndexKind::Vector, Some("multilingual")).await.unwrap(),
+                Index::create(&meta.pool, shard.id, IndexKind::Vector, Some("english")).await.unwrap(),
+                Index::create(&meta.pool, shard.id, IndexKind::Text, Some("fulltext")).await.unwrap(),
+                Index::create(&meta.pool, shard.id, IndexKind::Paragraph, Some("keyword")).await.unwrap(),
+                Index::create(&meta.pool, shard.id, IndexKind::Relation, Some("relation")).await.unwrap(),
+            ];
+            let mut seq: i64 = 0;
+
+            for index in indexes.iter() {
+                for _ in 0..10 {
```
javitonino, line 249: nitpick: I would flip the order of these two loops to be a bit closer to reality, i.e: loop over resources first and loop over indexes inside.

nidx/src/maintenance/scheduler.rs
```diff
@@ -158,23 +169,112 @@ pub async fn purge_deletions(meta: &NidxMetadata, consumer: &mut PullConsumer) -
...
+            }
+
+            // scheduled a job per index
+
+            assert_eq!(jobs.len(), indexes.len());
+
+            let mut indexes_ids = HashSet::new();
+            for index in indexes.iter() {
+                indexes_ids.insert(index.id);
+            }
```
javitonino, line 270: I cannot contain my love for one-liners, feel free to ignore xD

Maybe something like this?

```rust
let indexes_ids = HashSet::from_iter(indexes.iter().map(|i| i.id));
```

or

```rust
let indexes_ids: HashSet<_> = indexes.iter().map(|i| i.id).collect();
```

javitonino (changes requested):

javitonino:
Also, you may want to retarget this to `dev` to get it deployed.

nidx/nidx_types/src/lib.rs
jotare: I added the implementation because I thought it might be useful but I'll remove it

nidx/nidx_types/src/lib.rs
jotare: I think I prefer to keep both in this case if you don't mind

nidx/src/maintenance/scheduler.rs
jotare: Fixed! Let me know what you think. I also added some docs and tests

javitonino (approved):

PR #2575, opened 2024-10-28
Entity traceability to task in proto

carlesonielfa:
### Description
~Add a field to track where each NER was computed, so they can be tracked and removed/updated with the corresponding task id~

~Also added documentation of some fields that needed it~

~The current manner in which the NER fields are structured is not ideal, this PR aims to add the id without breaking anything~

My proposal is to add a list of entities (with their text, label and locations in the field) indexed by the data augmentation task id.

Ideally, in the near future this would also deprecate the field ner and positions which is where we send the entities extracted in the processor now. We could send them in the new field with an empty key "" or with key "processor" .

If you have any feedback or suggestions they are welcome! Also if you want to have a chat about it, we can meet

### How was this PR tested?


jotare (approved):

bloodbare (commented):
Do you think that will be an overlap of entities on multiple DA ? Setting the DA id on the entity may be easier to clean up.

PR #2578, opened 2024-10-28
Handle error to avoid ask from failing with a 500

lferran:
### Description

### How was this PR tested?


jotare (approved):
Why does this happens though? Should we return some kind of error to the user? How will they know they don't have the context from the neighbors?

PR #2585, opened 2024-10-30
Nidx shards API

jotare:
### Description
Partial implementation of NodeWriter gRPC API for Nidx

### How was this PR tested?


nidx/src/api/shards.rs
```diff
@@ -0,0 +1,82 @@
...
+        Ok(shard)
+    }
+
+    /// Mark a shard, its indexes and segments for eventual deletion
+    pub async fn delete_shard(&self, shard_id: Uuid) -> anyhow::Result<()> {
+        let mut tx = self.meta.transaction().await?;
+        let Some(shard) = Shard::try_get(&mut *tx, shard_id).await? else {
+            return Ok(());
+        };
+        for index in shard.indexes(&mut *tx).await?.into_iter() {
```
javitonino, line 72: You can do this in a single query instead of looping. It's fine either way, but since you are already writing a query specifically for this use case, might be better to have the query do all the things.

nidx/src/metadata/shard.rs
```diff
@@ -18,27 +18,43 @@
...
+    pub async fn create(meta: impl Executor<'_, Database = Postgres>, kbid: Uuid) -> sqlx::Result<Shard> {
         sqlx::query_as!(Shard, "INSERT INTO shards (kbid) VALUES ($1) RETURNING *", kbid).fetch_one(meta).await
     }
 
-    pub async fn get(meta: impl Executor<'_, Database = Postgres>, id: Uuid) -> Result<Shard, sqlx::Error> {
+    pub async fn get(meta: impl Executor<'_, Database = Postgres>, id: Uuid) -> sqlx::Result<Shard> {
         sqlx::query_as!(Shard, "SELECT * FROM shards WHERE id = $1", id).fetch_one(meta).await
     }
 
+    pub async fn try_get(meta: impl Executor<'_, Database = Postgres>, id: Uuid) -> sqlx::Result<Option<Shard>> {
```
javitonino, line 39: Consider removing this and instead, call `get` and match the result against `Err(RowNotFound)`

nidx/src/metadata/shard.rs
```diff
@@ -18,27 +18,43 @@
...
+        sqlx::query_as!(Shard, "SELECT * FROM shards WHERE id = $1", id).fetch_optional(meta).await
+    }
+
+    pub async fn mark_delete(&self, meta: impl Executor<'_, Database = Postgres>) -> sqlx::Result<()> {
+        sqlx::query!("UPDATE shards SET deleted_at = NOW() WHERE id = $1", self.id).execute(meta).await?;
+        Ok(())
+    }
+
+    pub async fn delete(&self, meta: impl Executor<'_, Database = Postgres>) -> sqlx::Result<()> {
```
javitonino, line 48: Is this called?

nidx/src/searcher/sync.rs
```diff
@@ -39,11 +39,18 @@ pub async fn run_sync(
 ) -> anyhow::Result<()> {
     let mut last_updated_at = PrimitiveDateTime::MIN.replace_year(2000)?;
     loop {
+        let deleted = Index::marked_to_delete(&meta.pool).await?;
         let indexes = Index::recently_updated(&meta.pool, last_updated_at).await?;
         for index in indexes {
             // TODO: Handle errors
+
+            if deleted.contains(&index.id) {
```
javitonino, line 47: This should be outside the loop. If an index is deleted that has not been recently updated, it won't be deleted from sync here.

nidx/src/searcher/sync.rs
```diff
@@ -83,6 +90,16 @@ async fn sync_index(
     Ok(())
 }
 
+async fn delete_index(
+    index_id: IndexId,
+    sync_metadata: Arc<SyncMetadata>,
+    notifier: &Sender<IndexId>,
+) -> anyhow::Result<()> {
+    sync_metadata.delete(&index_id).await;
```
javitonino, line 98: We also need to remove all downloaded segments for that index here.

charts/nidx/templates/shards-api-deploy.yaml
```diff
@@ -0,0 +1,63 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: nidx-shards-api
```
javitonino, line 4: I would rename the charts to `nidx-api`. Even if we have multiple API's we can deploy all of them in a single pod/process.

nidx/src/metadata/shard.rs
jotare: Ups, leftover. Removing it

charts/nidx/templates/shards-api-deploy.yaml
```diff
@@ -0,0 +1,63 @@
...
+      {{- with .Values.searcher.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.searcher.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      dnsPolicy: ClusterFirst
+      serviceAccount: {{ .Values.searcher.serviceAccount | default "default" }}
```
javitonino, line 52: We need to apply the changes from https://github.com/nuclia/nucliadb/pull/2590 here as well

PR #2588, opened 2024-10-30
Use new entities field in FieldMetadata

carlesonielfa:
### Description
* Adds ingestion of entities in FieldMetadata through the new field with fallback to the legacy field
* Sets the stage to deprecate the use of the fields `ner` and `positions`
* Keeps resource API backwards compatibility

The changes are compatible with
- current state (processor entities in legacy field + data augmentation entities in new)
- processor entities in both new and legacy fields 
- using only the new fields

### How was this PR tested?
Modified existing tests and fixtures to use the new field. Extended ingestion tests to test that data augmentation entities were being ingested

### DIFF
--- a/nucliadb/src/nucliadb/ingest/orm/brain.py
+++ b/nucliadb/src/nucliadb/ingest/orm/brain.py
@@ -499,7 +499,13 @@ class ResourceBrain:
                     )
                 )
         # Data Augmentation + Processor entities
+        use_legacy_entities = True
         for data_augmentation_task_id, entities in metadata.entities.items():
+            # If we recieved the entities from the processor here, we don't want to use the legacy entities
+            # TODO: Remove this when processor doesn't use this anymore
+            if data_augmentation_task_id == "processor":
+                use_legacy_entities = False
+
             for ent in entities.entities:
                 entity_text = ent.text
                 entity_label = ent.label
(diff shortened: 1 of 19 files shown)


nucliadb/src/nucliadb/ingest/orm/brain.py
```diff
@@ -498,24 +498,54 @@ def process_field_metadata(
...
+            for ent in entities.entities:
+                entity_text = ent.text
+                entity_label = ent.label
+                # Seems like we don't care about where the entity is in the text
+                # entity_positions = entity.positions
+                labels["e"].add(
+                    f"{entity_label}/{entity_text}"
+                )  # Add data_augmentation_task_id as a prefix?
+                relation_node_entity = RelationNode(
+                    value=entity_text,
```
jotare, line 518: Is this change on purpose? In the legacy version, `value` was the `label/text`, while now is only `text`.

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -1254,3 +1256,25 @@ def extract_field_metadata_languages(
...
+    target_entites_dict: Union[dict[str, str], ScalarMap[str, str]], field_metadata: FieldMetadata
+):
+    """
+    Update the entities dict with the entities from the field metadata.
+    Method created to ease the transition from legacy ner field to new entities field.
+    """
+    # Data Augmentation + Processor entities
+    # This will overwrite entities detected from more than one data augmentation task
+    # TODO: Change TrainMetadata proto to accept multiple entities with the same text
+    ents = {
```
jotare, line 1271: nit: prefer the full name (like `entities`) rather than a partial word like `ents`

nucliadb/src/nucliadb/ingest/orm/brain.py
carlesonielfa: Maybe I am mistaken, but in the legacy version I see that the value=`entity` and `entity` is only the text
```python
entity_array = klass_entity.split("/")
klass = entity_array[0] # This is the label
entity = "/".join(entity_array[1:]) # This takes everything except the label (so the text)
```

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -1254,3 +1256,25 @@ def extract_field_metadata_languages(
     for _, splitted_metadata in field_metadata.metadata.split_metadata.items():
         languages.add(splitted_metadata.language)
     return list(languages)
+
+
+def _update_entities_dict(
+    target_entites_dict: Union[dict[str, str], ScalarMap[str, str]], field_metadata: FieldMetadata
```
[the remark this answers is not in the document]
carlesonielfa: fixed! Good catch! I did not know the protobuf type implemented Mapping

nucliadb/src/nucliadb/ingest/orm/resource.py
carlesonielfa: Fixed!

jotare (approved):

PR #2589, opened 2024-10-30
cleanup of collector feedback

drf7:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/reader/api/v1/learning_collector.py
```diff
@@ -1,61 +0,0 @@
...
-from fastapi_versioning import version
-
-from nucliadb.learning_proxy import learning_collector_proxy
-from nucliadb.reader.api.v1.router import KB_PREFIX, api
-from nucliadb_models.resource import NucliaDBRoles
-from nucliadb_utils.authentication import requires
-
-
-@api.get(
-    path=f"/{KB_PREFIX}/{{kbid}}/feedback/{{month}}",
```
jotare, line 30: You'll have to remove the feedback endpoint from the VirtualService here: https://github.com/nuclia/nucliadb/blob/45fedfc20c9329d7b808ee7fd37f1082838cfada/charts/nucliadb_reader/templates/reader.vs.yaml#L45

jotare (changes requested):

nucliadb/src/nucliadb/reader/api/v1/learning_collector.py
drf7: Thanks! Done

jotare (approved):

PR #2599, opened 2024-11-06
Use strategy pattern for rank fusion algorithms + Implement reciprocal rank fusion (RRF)

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_models/src/nucliadb_models/search.py
+++ b/nucliadb_models/src/nucliadb_models/search.py
@@ -1352,7 +1352,9 @@ class AskRequest(AuditMetadataBase):
         title="Prompts",
         description="Use to customize the prompts given to the generative model. Both system and user prompts can be customized. If a string is provided, it is interpreted as the user prompt.",  # noqa
     )
-    rank_fusion: Union[RankFusionName, RankFusion] = SearchParamDefaults.rank_fusion.to_pydantic_field()
+    rank_fusion: SkipJsonSchema[Union[RankFusionName, RankFusion]] = (
+        SearchParamDefaults.rank_fusion.to_pydantic_field()
+    )
     reranker: Reranker = SearchParamDefaults.reranker.to_pydantic_field()
     citations: bool = Field(
         default=False,
@@ -1580,7 +1582,9 @@ class FindRequest(BaseSearchRequest):
             SearchOptions.SEMANTIC,
         ]
     )
-    rank_fusion: Union[RankFusionName, RankFusion] = SearchParamDefaults.rank_fusion.to_pydantic_field()
+    rank_fusion: SkipJsonSchema[Union[RankFusionName, RankFusion]] = (
+        SearchParamDefaults.rank_fusion.to_pydantic_field()
+    )
     reranker: Reranker = SearchParamDefaults.reranker.to_pydantic_field()
 
     keyword_filters: Union[list[str], list[Filter]] = Field(
--- a/nucliadb/src/nucliadb/search/api/v1/find.py
+++ b/nucliadb/src/nucliadb/search/api/v1/find.py
@@ -126,7 +126,9 @@ async def find_knowledgebox(
     autofilter: bool = fastapi_query(SearchParamDefaults.autofilter),
     security_groups: list[str] = fastapi_query(SearchParamDefaults.security_groups),
     show_hidden: bool = fastapi_query(SearchParamDefaults.show_hidden),
-    rank_fusion: RankFusionName = fastapi_query(SearchParamDefaults.rank_fusion),
+    rank_fusion: RankFusionName = fastapi_query(
+        SearchParamDefaults.rank_fusion, include_in_schema=False
+    ),
     reranker: Reranker = fastapi_query(SearchParamDefaults.reranker),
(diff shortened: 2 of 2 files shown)


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -389,6 +389,26 @@ class SortOptions(BaseModel):
...
+class LegacyRankFusion(_BaseRankFusion):
+    name: Literal[RankFusionName.LEGACY] = RankFusionName.LEGACY
+
+
+class ReciprocalRankFusion(_BaseRankFusion):
+    name: Literal[RankFusionName.RECIPROCAL_RANK_FUSION] = RankFusionName.RECIPROCAL_RANK_FUSION
+
+
+RankFusion = Union[LegacyRankFusion, ReciprocalRankFusion]
+
```
lferran, line 410: I think you need to do something similar to what we do with rag strategies so that pydantic can parse/validate the correct model at hand

```python
RagStrategies = Annotated[
    Union[
        FieldExtensionStrategy,
        FullResourceStrategy,
        HierarchyResourceStrategy,
        NeighbouringParagraphsStrategy,
        MetadataExtensionStrategy,
        PreQueriesStrategy,
    ],
    Field(discriminator="name"),
```

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1530,16 +1556,9 @@ class FindRequest(BaseSearchRequest):
             SearchOptions.SEMANTIC,
         ]
     )
+    rank_fusion: Union[RankFusionName, RankFusion] = SearchParamDefaults.rank_fusion.to_pydantic_field()
```
lferran, line 1559: why not just `RankFusion`?

nucliadb_models/src/nucliadb_models/search.py
jotare: Done :)

nucliadb/src/nucliadb/search/search/rank_fusion.py
```diff
@@ -0,0 +1,148 @@
...
+        merged.sort(key=lambda x: x.score, reverse=True)
+        return merged
+
+
+def get_default_rank_fusion() -> RankFusionAlgorithm:
+    return LegacyRankFusion()
+
+
+def get_rank_fusion(
+    rf: Union[search_models.RankFusionName, search_models.RankFusion],
```
lferran, line 129: should be `Optional[Union[...]]`

nucliadb_models/src/nucliadb_models/search.py
jotare: To make the API cleaner. You can call by name and expect default parameters or specify them (if they have).
You can be shorter with
```
 {"rank_fusion": "legacy"}
```
instead of
```
{"rank_fusion": {"name": "legacy"}}
```
And you can be explicit and use parameters if you need to:
```
{"rank_fusion": {"name": "rrf", "k": 50}}
```

nucliadb/src/nucliadb/search/search/rank_fusion.py
jotare: Oh, it was a leftover. I won't compare with `None` and keep the `Union` as it is now

nucliadb/src/nucliadb/search/search/rank_fusion.py
```diff
@@ -0,0 +1,148 @@
...
+                match_positions.setdefault(result.paragraph_id, []).append(position)
+
+        merged = []
+        for paragraph_id, positions in match_positions.items():
+            for r, i in positions:
+                score = scores[paragraph_id]
+                result = rankings[r][i]
+                result.score = score
+                result.score_type = SCORE_TYPE.RANK_FUSION
+                # NOTE we are appending multi-matches. Should we merge them?
```
lferran, line 117: I think we should somehow boost them no?

jotare: Nevermind, I'll add it

lferran (approved):

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -389,6 +390,40 @@ class SortOptions(BaseModel):
...
+        default=60.0,
+        title="RRF k parameter",
+        description="k parameter changes the influence top-ranked and lower-ranked elements have. Research has shown that 60 is a performant value across datasets",  # noqa: E501
+    )
+    boosting: dict[Union[Literal["keyword"], Literal["semantic"]], float] = Field(
+        default_factory=partial(dict, keyword=1.0, semantic=1.0),
+        title="Retriever boosting",
+        description="Define weights for each retriever. This allows to give different priorities to different retrieval methods. Scores will be multiplied by this value, so a boost of 1.0 is the same as no boost. As an example, multilingual search may benefit from semantic boosting",  # noqa: E501
+        min_length=2,
+    )
```
jotare, line 418: Is this too hacky?

lferran: why not definig a new model with validation? 
I would put defaults to 1 to both keyword and semantic, and then compute the relative weights between both.

This way if the user specifies only one of them, the other is taken as 1.

PR #2601, opened 2024-11-07
Show mime type on extracted metadata

bloodbare:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/ingest/orm/brain.py
```diff
@@ -483,6 +515,8 @@ def process_field_metadata(
         relation_node_document: RelationNode,
         user_canceled_labels: set[str],
     ):
+        if metadata.mime_type != "":
+            labels["mt"].add(metadata.mime_type)
```
lferran, line 519: @bloodbare we already have a label for the icon `/n/i`. Shouldn't we reuse that one?

bloodbare: /n/i is the icon of the resource where it belongs, not the mime tyupe of the field

PR #2606, opened 2024-11-07
Fix synonyms advanced query

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/query.py
+++ b/nucliadb/src/nucliadb/search/search/query.py
@@ -547,7 +547,8 @@ class QueryParser:
         - Query: "What is Foo?"
         - Advanced Query: "What is (Foo OR Bar OR Baz)?"
         """
-        if not self.with_synonyms:
+        if not self.with_synonyms or not self.query:
+            # Nothing to do
             return
 
         if self.has_vector_search or self.has_relations_search:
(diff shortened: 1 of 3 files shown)


nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -548,27 +557,32 @@ async def parse_synonyms(self, request: nodereader_pb2.SearchRequest) -> None:
...
+        variants: dict[str, str] = {}
+        for term, term_synonyms in synonyms.terms.items():
+            if len(term_synonyms.synonyms) > 0:
+                variants[term] = f"({" OR ".join([term] + list(term_synonyms.synonyms))})"
+
+        # Split the query into terms
+        query_terms = self.query.split()
+
+        # Remove punctuation from the query terms
+        clean_query_terms = [term.strip(".,;:!?¿") for term in query_terms]
```
javitonino, line 575: You might want to use something like `string.punctuation` in order to get more symbols?. Or if we need international support we can use `unicodedata.category(chr)` or a regex.

javitonino (approved):
A small suggestion in case we want to catch more punctuation.

nucliadb/src/nucliadb/search/search/query.py
lferran: cool, I'll go for string punctuation for now

PR #2609, opened 2024-11-08
Nidx shards API & tests

javitonino:
Complete shards API and migrate tests.

There is a tricky thing in this PR, we want the NidxApi grpc service to containe `NewShard` from nodewriter and `GetShard` from nodereader. At some point, we will completely split the nidx protos but, until then:
- Create new grpc services NidxApi and NidxSearcher structured how nidx likes it. Still uses all `Message`s from nucliadb_protos
- Since the grpc service name is part of the grpc protocol (in the URL path), made a middleware to remap requests to the correct path.

Aside from that, another notable change is to add a `NidxError` type to make it somewhat easier to map application errors to tonic errors. The idea is to only add whatever needs to be mapped differently.


nidx/nidx_protos/build.rs
```diff
@@ -22,13 +22,15 @@ use std::io::Result;
 
 fn main() -> Result<()> {
     println!("cargo:rerun-if-changed=../../nucliadb_protos");
+    println!("cargo:rerun-if-changed=src");
     tonic_build::configure().emit_rerun_if_changed(false).compile_protos(
         &[
+            "src//nidx.proto",
```
jotare, line 28: Double slash for any reason?
```suggestion
            "src/nidx.proto",
```

nidx/src/errors.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+}
+
+impl From<NidxError> for tonic::Status {
+    fn from(value: NidxError) -> Self {
+        match value {
+            NidxError::NotFound => tonic::Status::not_found("Not found"),
+            NidxError::InvalidRequest(_) => tonic::Status::invalid_argument(value.to_string()),
+            NidxError::InvalidUuid(_) => tonic::Status::invalid_argument(value.to_string()),
```
jotare, line 63: I'm up for changing this, but I remember we used internal errors everywhere to not mix application errors with the protocol

nidx/src/api/grpc.rs
```diff
@@ -44,26 +45,46 @@ impl ApiServer {
...
+            fields: *index_stats.get(&IndexKind::Text).unwrap_or(&0) as u64,
+            paragraphs: *index_stats.get(&IndexKind::Paragraph).unwrap_or(&0) as u64,
+            sentences: *index_stats.get(&IndexKind::Vector).unwrap_or(&0) as u64,
+        }))
+    }
+    async fn new_shard(&self, request: Request<NewShardRequest>) -> Result<Response<ShardCreated>> {
         // TODO? analytics event
         let request = request.into_inner();
-        let kbid = Uuid::from_str(&request.kbid).map_err(|e| Status::internal(e.to_string()))?;
+        let kbid = Uuid::from_str(&request.kbid).map_err(NidxError::from)?;
```
jotare, line 80: I love this new error type <3

jotare (approved):

nidx/nidx_protos/build.rs
javitonino: Oops :upside_down_face:

nidx/src/errors.rs
javitonino: Yeah, I tried to keep them separate. This is (or should be) the single place where they mix, to do the generic mapping from one to another. I think it's better than repeating ourselves all over the codebase. The idea is to keep all conversion here and just do `map_err` whenever we need to.

PR #2619, opened 2024-11-11
Nidx search test

javitonino:
Migrate remaining nucliadb_node tests related to search capabilities.

TODO: implement suggest()

### DIFF
--- a/nidx/nidx_relation/src/lib.rs
+++ b/nidx/nidx_relation/src/lib.rs
@@ -69,7 +69,7 @@ impl RelationIndexer {
         }
 
         index_relations(&mut indexer, resource, field_schema)?;
-        Ok(Some(indexer.finalize()?))
+        indexer.finalize()
     }
 
     pub fn deletions_for_resource(&self, resource: &nidx_protos::Resource) -> Vec<String> {
--- a/nidx/nidx_relation/src/lib.rs
+++ b/nidx/nidx_relation/src/lib.rs
@@ -32,7 +32,6 @@ use nidx_tantivy::{
     TantivyIndexer, TantivyMeta, TantivySegmentMetadata,
 };
 use nidx_types::OpenIndexMetadata;
-use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
 use reader::{HashedRelationNode, RelationsReaderService};
 use resource_indexer::index_relations;
 pub use schema::Schema as RelationSchema;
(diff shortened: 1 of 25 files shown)


nidx/nidx_relation/src/lib.rs
```diff
@@ -120,4 +129,36 @@ impl RelationSearcher {
     pub fn search(&self, request: &RelationSearchRequest) -> anyhow::Result<RelationSearchResponse> {
         self.reader.search(request)
     }
+
+    pub fn suggest(&self, prefixes: Vec<String>) -> Vec<RelationNode> {
+        let requests =
+            prefixes.par_iter().filter(|prefix| prefix.len() >= MIN_SUGGEST_PREFIX_LENGTH).cloned().map(|prefix| {
```
jotare, line 135: Adding rayon for this. Do we really want it?

nidx/src/searcher/shard_suggest.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    let paragraph_task = suggest_paragraphs.then_some(move || paragraph_searcher.suggest(&request));
+
+    let relation_task = suggest_entities.then_some(move || relation_searcher.suggest(prefixes));
+
+    let mut rparagraph = None;
+    let mut rrelation = None;
+
+    std::thread::scope(|scope| {
```
jotare, line 142: Using std::thread instead of crossbeam to avoid another dependency

nidx/nidx_relation/src/lib.rs
javitonino: I think we can live without this and just use a plain iterator. If we find the need for speed here, we can add it later.

I don't think we will have many prefixes anyway (one per word on the query) and I'd rather remove the complexity until we are sure we need it. It's easy to add back again if we want to.

nidx/src/api/shards.rs
```diff
@@ -72,3 +72,23 @@ pub async fn delete_shard(meta: &NidxMetadata, shard_id: Uuid) -> NidxResult<()>
...
 }
+
+pub async fn delete_vectorset(meta: &NidxMetadata, shard_id: Uuid, vectorset: &str) -> NidxResult<()> {
+    let mut tx = meta.transaction().await?;
+
+    // TODO: query to check how many vector indexes we have left
+    let count =
+        Index::for_shard(&mut *tx, shard_id).await?.iter().filter(|index| index.kind == IndexKind::Vector).count();
+    if count <= 1 {
+        return Err(NidxError::InvalidRequest("Can't delete the your vectorset".to_string()));
```
javitonino, line 83:
```suggestion
        return Err(NidxError::InvalidRequest("Can't delete the last vectorset".to_string()));
```

nidx/src/api/shards.rs
```diff
@@ -72,3 +72,23 @@ pub async fn delete_shard(meta: &NidxMetadata, shard_id: Uuid) -> NidxResult<()>
...
+    }
+
+    let index = Index::find(&mut *tx, shard_id, IndexKind::Vector, vectorset).await?;
+    MergeJob::delete_many_by_index(&mut *tx, index.id).await?;
+    Segment::mark_delete_by_index(&mut *tx, index.id).await?;
+    index.mark_delete(&mut *tx).await?;
```
javitonino, line 89: Don't we have to mark deletions as well? Same in the case of shard deletion

PR #2625, opened 2024-11-12
Nidx binding for tests

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/find.py
+++ b/nucliadb/src/nucliadb/search/search/find.py
@@ -64,7 +64,6 @@ async def find(
     x_forwarded_for: str,
     generative_model: Optional[str] = None,
     metrics: RAGMetrics = RAGMetrics(),
-    nidx: bool = False,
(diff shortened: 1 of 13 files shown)


nucliadb/src/nucliadb/search/search/find.py
```diff
@@ -64,7 +64,6 @@ async def find(
     x_forwarded_for: str,
     generative_model: Optional[str] = None,
     metrics: RAGMetrics = RAGMetrics(),
-    nidx: bool = False,
```
jotare, line 67: Wouldn't be useful to do queries to nidx and nodes to validate?

jotare (approved):

nucliadb/src/nucliadb/search/search/find.py
javitonino: Maybe, but it's very annoying to have this parameter (not only here, but in all search endpoints: search, ask, suggest, etc.). We can try adding a feature flag later if we need to, maybe? Otherwise, we will have the tests to validate.

PR #2626, opened 2024-11-12
Remove *Search gRPC methods and cleanup search

jotare:
### Description

### How was this PR tested?


nidx/nidx_protos/src/nidx.proto
```diff
@@ -33,7 +33,6 @@ service NidxSearcher {
   rpc DocumentSearch(nodereader.DocumentSearchRequest) returns (nodereader.DocumentSearchResponse) {}
   rpc ParagraphSearch(nodereader.ParagraphSearchRequest) returns (nodereader.ParagraphSearchResponse) {}
   rpc VectorSearch(nodereader.VectorSearchRequest) returns (nodereader.VectorSearchResponse) {}
-  rpc RelationSearch(nodereader.RelationSearchRequest) returns (nodereader.RelationSearchResponse) {}
```
javitonino, line 36: Maybe we can remove DocumentSearch, ParagraphSearch, etc. as well?

nucliadb/src/nucliadb/search/search/shards.py
```diff
@@ -61,11 +61,41 @@ async def get_shard(node: AbstractIndexNode, shard_id: str) -> Shard:
 async def query_paragraph_shard(
     node: AbstractIndexNode, shard: str, query: ParagraphSearchRequest
 ) -> ParagraphSearchResponse:
-    req = ParagraphSearchRequest()
-    req.CopyFrom(query)
-    req.id = shard
+    req = SearchRequest()
```
javitonino, line 64: This is called in a single place `resource_search`. Maybe it makes more sense to put this logic (or a subset of it) in `resource_search` and remove this method altogether?

javitonino (approved):
Nice. Left a comment for a little further cleanup if you feel like doing it now (we can always do a second pass at some point)

nidx/nidx_protos/src/nidx.proto
jotare: Yes, I was working on it :)

nucliadb/src/nucliadb/search/search/shards.py
jotare: Yes, still working on it. The first step was to remove the calls in the innermost point and go bottom to top

nidx/nidx_protos/src/nidx.proto
jotare: Done

nucliadb/src/nucliadb/search/search/shards.py
jotare: Done

javitonino (approved):

PR #2629, opened 2024-11-13
Standalone tests (except rollovers)

javitonino:
We got some problem with standalone tests with nidx:

- Some queries don't send the `vectorset` field in the search. In this example, the `test_ask_rag_strategy_prequeries_` will fail because of this. Not sure where the field is missing.
- Search and indexer in nidx is not synced, we have to wait for searcher to download the newly indexed segments before querying. In this case, I do it by explicitly waiting for a sync to occur (this will fail if not running with NIDX_ENABLED=1, but that's an easy fix). Maybe we could make this implicit (put it inside any call to create shard or index a resource). 

To test this, run with `NIDX_ENABLED=1` as an env variable.

### DIFF
--- a/nucliadb/tests/nucliadb/integration/common/cluster/test_rebalance.py
+++ b/nucliadb/tests/nucliadb/integration/common/cluster/test_rebalance.py
@@ -62,8 +62,6 @@ async def test_rebalance_kb_shards(
         )
         assert resp.status_code == 201
 
-    breakpoint()
-
     counters1_resp = await nucliadb_manager.get(f"/kb/{knowledgebox}/counters")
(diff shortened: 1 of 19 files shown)


nucliadb/tests/nucliadb/integration/common/cluster/test_rebalance.py
```diff
@@ -62,6 +62,8 @@ async def test_rebalance_kb_shards(
         )
         assert resp.status_code == 201
 
+    breakpoint()
```
jotare, line 65: Forgot a breakpoint?

PR #2633, opened 2024-11-14
Nidx rollover

javitonino:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/common/cluster/manager.py
```diff
@@ -317,6 +318,17 @@ async def rollback_shard(self, shard: writer_pb2.ShardObject):
                         exc_info=True,
                     )
 
+        nidx_api = get_nidx_api_client()
+        if nidx_api:
+            try:
+                await nidx_api.DeleteShard(noderesources_pb2.ShardId(id=shard.nidx_shard_id))
```
lferran, line 324: could it be that you're trying to rollback a nidx shard id that hasnt been set?

PR #2636, opened 2024-11-15
Remove local maindb driver

lferran:
### Description
We remove the local implementation of the maindb driver for the following reasons:

- We have opted in for pg as maindb since long time ago.
- It has never been production-ready, it has some known bugs and it has been only used for testing so far.
- Right now it's very easy to spin up a local pg instance and make nucliadb depend on it.

This will allow us to:
- Remove the catalog implementation based on the index node. 
- Then, we can have shard assignation for resources when ingesting messages coming from processing.
- The above two points allow us to increase/remove the back-pressure mechanism -- we had it in place because we were assigning resources to shards on resource creation, and that, when the ingestion pipeline was too loaded, could make us place way more resources into a shard than we were capable to handle.

### How was this PR tested?
Existing tests should cover us. I'm just removing code


nucliadb_sdk/src/nucliadb_sdk/tests/fixtures.py
```diff
@@ -27,18 +27,22 @@
...
 
+import docker  # type: ignore
 import pytest
 import requests
 from pytest_docker_fixtures import images  # type: ignore
 from pytest_docker_fixtures.containers._base import BaseImage  # type: ignore
 
 import nucliadb_sdk
 
+images.settings["postgresql"]["version"] = "11"
```
jotare, line 38: I think nidx needs v13, maybe we can already use that

jotare (approved):

PR #2638, opened 2024-11-15
Remove non pg catalog

lferran:
### Description
Local files maindb driver is no longer supported, so we'll always have the catalog enabled!

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/search.py
+++ b/nucliadb/src/nucliadb/search/api/v1/search.py
@@ -35,7 +35,7 @@ from nucliadb.search.requesters.utils import Method, debug_nodes_info, node_quer
 from nucliadb.search.search import cache
 from nucliadb.search.search.exceptions import InvalidQueryError
 from nucliadb.search.search.merge import fetch_resources, merge_results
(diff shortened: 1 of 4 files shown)


nucliadb/src/nucliadb/search/api/v1/search.py
```diff
@@ -322,51 +322,15 @@ async def catalog(
...
-                    resources=[r.rid for r in catalog_results.fulltext.results],
-                    kbid=kbid,
-                    show=[ResourceProperties.BASIC, ResourceProperties.ERRORS],
-                    field_type_filter=list(FieldTypeName),
-                    extracted=[],
-                )
-                queried_nodes = []
-            if len(queried_nodes) > 0:
-                queried_shards = [shard_id for _, shard_id in queried_nodes]
-                catalog_results.shards = queried_shards
```
javitonino, line 369: You are no longer returning shards. Make sure this doesn't break anything in frontend! Not sure if they removed depending on this yet.

javitonino (approved):
We can also remove the initial index message when uploading a field, that just sets some stuff in the text index.

PR #2641, opened 2024-11-15
Index only upon processing message received

lferran:
### Description
We no longer need to index anything when ingesting writer messages, as the catalog is responsible for showing the title and the labels to the user after an upload/resource creation request is finished.

Also, when a kb is configured with an external index provider, we no longer need to index on our index either.

### How was this PR tested?


lferran:
not ready yet for this. will come back to it later

PR #2660, opened 2024-11-27
Handle highlight errors

lferran:
### Description

### How was this PR tested?


lferran:
fixes SEARCH-WV

jotare (approved):
Which error are we trying to catch here?

lferran:
if you look at sentry, there is some problems with highlighting the greek alphabet.

We can fix that separately, but at least we won't be failing the whole request now

PR #2664, opened 2024-11-28
Remove legacy rank fusion algorithm from the API

jotare:
### Description
Legacy rank fusion was an heuristic built long time ago. We consider RRF better, so this PR changes the default and removes the legacy algorithm from the public API. The algorithm implementation hasn't been removed yet.

### How was this PR tested?
Existing tests

### DIFF
--- a/nucliadb/src/nucliadb/search/search/rank_fusion.py
+++ b/nucliadb/src/nucliadb/search/search/rank_fusion.py
@@ -107,6 +107,16 @@ class LegacyRankFusion(RankFusionAlgorithm):
         return merged
 
 
+class CombSimple(RankFusionAlgorithm):
+    @rank_fusion_observer.wrap({"type": "legacy"})
+    def _fuse(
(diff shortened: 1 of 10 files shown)


nucliadb/src/nucliadb/search/search/rank_fusion.py
```diff
@@ -107,6 +107,16 @@ def _fuse(
...
 
 
+class CombSimple(RankFusionAlgorithm):
+    @rank_fusion_observer.wrap({"type": "legacy"})
+    def _fuse(
+        self, keyword: Iterable[TextBlockMatch], semantic: Iterable[TextBlockMatch]
+    ) -> list[TextBlockMatch]:
+        merged = [*keyword, *semantic]
+        merged.sort(key=lambda r: r.score, reverse=True)
+        return merged
```
lferran, line 117: is this used anywhere? if not, let's remove it no?

jotare: Ups, leftover! I'll remove it

lferran (approved):

PR #2665, opened 2024-11-28
Ingest data augmentation support

jotare:
### Description
Support sending back to process fields autogenerated by data augmentation tasks

### How was this PR tested?


nucliadb/src/nucliadb/ingest/orm/processor/processor.py
```diff
@@ -0,0 +1,741 @@
...
+
+                        if await has_generated_fields(message, resource):
+                            await send_generated_fields_to_process(kbid, resource, message)
+
+                    else:
+                        # TODO: use a more concrete exception
+                        raise Exception(f"Unknown broker message source {message.source}")
+
+                    # apply changes from the broker message to the resource
+                    await self.apply_resource(message, resource, update=(not created))
```
javitonino, line 305: nitpick: Maybe we can rename the variable `created` to `updated` or something and avoid this inversion? It's not bad this way anyway, so up to you.

Edit: Ah! It was this way before and used below. Let's not change it then.

nucliadb/src/nucliadb/ingest/orm/processor/processor.py
```diff
@@ -0,0 +1,741 @@
...
+                for message in messages:
+                    if resource is not None:
+                        assert resource.uuid == message.uuid
+
+                    if message.source == writer_pb2.BrokerMessage.MessageSource.WRITER:
+                        resource = await kb.get(uuid)
+
+                        if resource is None:
+                            # It's a new resource
+                            resource = await kb.add_resource(uuid, message.slug, message.basic)
```
javitonino, line 280: This seems to be a new call, not sure yet why this is here :thinking: 

Edit: Ah, this was inside apply_resource before. :ok_hand:

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
```diff
@@ -0,0 +1,159 @@
...
+    ingest the processed thing later).
+
+    Given a broker message and a resource, this function returns the list of
+    generated fields, that can be empty.
+
+    """
+    generated_fields = GeneratedFields()
+
+    # only messages from processor can add generated fields
+    if bm.source != writer_pb2.BrokerMessage.MessageSource.PROCESSOR:
```
javitonino, line 64: I think this check is not strictly necessary. Not saying we should remove it, double checks are good.

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+@dataclass
+class GeneratedFields:
+    texts: list[str] = field(default_factory=list)
+    links: list[str] = field(default_factory=list)
+    files: list[str] = field(default_factory=list)
+    conversations: list[str] = field(default_factory=list)
+
+    def __len__(self):
```
javitonino, line 42: This is cool, although it made me do a double take (I saw the `len(generated_fields)` first and I was surprised classes in Python had length :sweat_smile:).

Anyway, since we only care if there are generated fields or no, what about replacing this function for something like `is_empty()`. It would express our intent more explicitly.

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
```diff
@@ -0,0 +1,159 @@
...
+
+    return generated_fields
+
+
+async def send_generated_fields_to_process(
+    kbid: str,
+    resource: Resource,
+    bm: writer_pb2.BrokerMessage,
+):
+    generated_fields = await _get_generated_fields(bm, resource)
```
javitonino, line 93: We are calling the `_get_generated_fields` function twice (first to check if there are any, second here).

This would be fine, except that `_get_generated_fields` calls `has_field` multiple times. `has_field` is not cached, each call is a call to the database. So we end up with 2*number_of_fields redundant calls to the database.

So, two things to see here:
1. In `_get_generated_fields` avoid calling `has_field` multiple times. Instead, call `get_all_field_ids` and compare if the incoming field is in that list.
2. Avoid calling `_get_generated_fields`. The usual approaches are:
  a) Moving the check inside this function
  b) Making this function take the list of generated fields by parameter.

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
jotare: Yeah, it's not but I think it's good to have the explicit check there, so the function is complete and not depending on a previous check

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
jotare: Solved with option 2

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
```diff
@@ -39,13 +39,18 @@ class GeneratedFields:
     files: list[str] = field(default_factory=list)
     conversations: list[str] = field(default_factory=list)
 
-    def __len__(self):
-        return len(self.texts) + len(self.links) + len(self.files) + len(self.conversations)
+    def is_empty(self) -> bool:
```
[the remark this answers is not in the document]
jotare: Hahaha, sure

nucliadb/src/nucliadb/ingest/orm/processor/data_augmentation.py
javitonino: Still missing the problem where the function calls `has_field` repeatedly, and each call is a query.

javitonino (approved):

PR #2678, opened 2024-12-02
Fix broken src-layout references

emmanuel-ferdman:
### Description
Small PR - fixes the broken references made due to the d7a3531973890f059abfd43f3cb557d3960c4e30 change.


jotare (approved):
Thanks for the contribution!!

PR #2680, opened 2024-12-02
Don't serialize unset values on SDK requests

jotare:
### Description
To minimize friction between API changes and SDK versions, this PR removes unset parameters from API calls. This way, only explicitly set values are sent to the server. That way, the server can evolve more flexibly without breaking the SDK.

This could lead to a weird situation where the user can see a default in a model, the server has a differnet one and leaving it unset will use the server one instead of the one in the user's local model. However, as the service is a live API without strong versioning, live documentation should always be the reference.


lferran (commented):
WDYT about doing a major version bump for this? so we're covered in case there are some behaviour changes. We may need to communicate this aswell?

bloodbare (approved):

PR #2688, opened 2024-12-03
Full resource rag strategy exclude

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_models/src/nucliadb_models/search.py
+++ b/nucliadb_models/src/nucliadb_models/search.py
@@ -1069,7 +1069,8 @@ class FieldExtensionStrategy(RagStrategy):
 class FullResourceApplyTo(BaseModel):
     exclude: list[str] = Field(
         default_factory=list,
-        description="Use only the paragraph for the resources containing any of these labels",
+        title="Labels to exclude from full resource expansion",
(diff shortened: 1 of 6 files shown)


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1066,6 +1066,13 @@ def fields_validator(cls, fields) -> Self:
         return fields
 
 
+class FullResourceApplyTo(BaseModel):
```
lferran, line 1069: What about naming this more generically, just in case? `ApplyTo`

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1066,6 +1066,13 @@ def fields_validator(cls, fields) -> Self:
...
+    exclude: list[str] = Field(
```
lferran, line 1070: exclude_with_labels ?

nucliadb_models/src/nucliadb_models/search.py
jotare: I prefixed it with FullResource on purpose so if we need a different ApplyTo for a different RAG strategy they don't collide. But I can change it if you don't like it

nucliadb_models/src/nucliadb_models/search.py
jotare: It was defined like this in the story, let me ask to our API master :)

javitonino (approved):

PR #2692, opened 2024-12-04
Ingest nats pull consumers

lferran:
### Description
This PR changes ingest nats consumers from push-based to pull-based.

### How was this PR tested?
Integration tests, and live tests in a dev cluster.


nucliadb/src/nucliadb/ingest/consumer/consumer.py
```diff
@@ -272,24 +284,77 @@ class IngestProcessedConsumer(IngestConsumer):
...
+        """
+        XXX NOTE: Getting the last sequence id for the processed pull consumer is only needed when the new pull consumer is created.
+
+        For environments where we had the previous push consumer, we need to get the last sequence id from the old consumer info so that
+        we can start from there.
+
+        This code must be deleted once the new pull consumers are deployed on all environments and all the old push consumers have been deleted.
+        From then on, we will rely on the nats server to keep track of the last sequence id.
+        """
+        if not has_feature(const.Features.PULL_PROCESSED_CONSUMERS_DEPLOYED):
```
javitonino, line 297: Shouldn't this be inverted?.

What I think is the desired behaviour:
If the feature flag is enabled, it is deployed and we don't check nats anymore.
If the feature flags is disabled, when creating the pull consumer we check nats of the old stream.

It is currently inverted.

javitonino (changes requested):

nucliadb/src/nucliadb/ingest/consumer/consumer.py
lferran: huh, good catch. I inverted the "meaning" of the feature flag and forgot to change the `not`.
I will add a test for this just in case.
Thanks mate!

javitonino (approved):

PR #2707, opened 2024-12-09
Do not require querying metadata during search

javitonino:
Avoid querying metadata during search. Instead, store needed metadata during sync and use that instead. This avoids hitting the database during searches.


nidx/src/searcher/shard_search.rs
```diff
@@ -83,34 +101,34 @@ pub async fn search(
...
 
 fn blocking_search(
-    search_request: SearchRequest,
-    paragraph_searcher: &ParagraphSearcher,
-    relation_searcher: &RelationSearcher,
-    text_searcher: &TextSearcher,
+    query_plan: QueryPlan,
+    paragraph_searcher: Option<&ParagraphSearcher>,
+    relation_searcher: Option<&RelationSearcher>,
+    text_searcher: Option<&TextSearcher>,
```
jotare, line 107: Not really convinced with these `Option`. You're trying to express that each index will only be there if it's needed in the request, but the function signature don't express this semantics properly.
Probably you can't do better with the current query plan though

jotare (approved):

nidx/src/searcher/shard_search.rs
javitonino: Yeah, not a fan either. I'm open to suggestions. I thought about wrapping all this in a `QueryPlanExecutor` struct, but in the end, it ends up being pretty much the same logic as this function, just hidden away in an struct constructor. Since it's just used in this place, I didn't think it was worth the extra abstraction.

PR #2715, opened 2024-12-10
Fix conversational Text

bloodbare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/prompt.py
+++ b/nucliadb/src/nucliadb/search/search/chat/prompt.py
@@ -680,6 +680,7 @@ async def conversation_prompt_context(
             if field_type == "c" and paragraph.score_type in (
                 SCORE_TYPE.VECTOR,
                 SCORE_TYPE.BOTH,
+                SCORE_TYPE.BM25,
             ):
                 field_unique_id = "-".join([rid, field_type, field_id])
                 if field_unique_id in analyzed_fields:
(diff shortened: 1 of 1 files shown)


nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -680,6 +680,7 @@ async def conversation_prompt_context(
             if field_type == "c" and paragraph.score_type in (
                 SCORE_TYPE.VECTOR,
                 SCORE_TYPE.BOTH,
+                SCORE_TYPE.BM25,
```
lferran, line 683: If we're comparing against all score types, should we remove this check entirely?

nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -695,20 +696,31 @@ async def conversation_prompt_context(
...
+
                         for message in conv.messages:
-                            text = message.content.text.strip()
-                            pid = f"{rid}/{field_type}/{field_id}/{message.ident}/0-{len(message.content.text) + 1}"
+                            ident = message.ident
+                            if extracted_text is not None:
+                                text = extracted_text.split_text.get(ident, message.content.text.strip())
+                            else:
+                                text = message.content.text.strip()
+                            pid = f"{rid}/{field_type}/{field_id}/{ident}/0-{len(message.content.text) + 1}"
```
lferran, line 709: Shoudn't this be:
```suggestion
                            pid = f"{rid}/{field_type}/{field_id}/{ident}/0-{len(text) - 1}"
```
?

nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -695,20 +696,31 @@ async def conversation_prompt_context(
...
                     if len(first_page.messages) > 0:
                         message = first_page.messages[0]
-                        text = message.content.text.strip()
-                        pid = f"{rid}/{field_type}/{field_id}/{message.ident}/0-{len(message.content.text) + 1}"
+                        ident = message.ident
+                        if extracted_text is not None:
+                            text = extracted_text.split_text.get(ident, message.content.text.strip())
+                        else:
+                            text = message.content.text.strip()
+                        pid = f"{rid}/{field_type}/{field_id}/{ident}/0-{len(message.content.text) + 1}"
```
lferran, line 723: same as above. Shouldn't we be using the length of the computed variable `text` instead?

lferran (approved):

nucliadb/src/nucliadb/search/search/chat/prompt.py
jotare: I think we are fine being explicit, otherwise adding a new score type will, for sure, miss an implicit condition here

PR #2716, opened 2024-12-11
Storage inserts instead of uploads

lferran:
### Description
This PR aims to speed up the way we store extracted metadata (protobuff serialized as binaries) to the object store.
Right now we are doing so as regular uploads, which include 3 requests to the storage (start, append, finish). The PR replaces these with a simple insert/put request.

### How was this PR tested?
Improved existing integration tests, and local tests.

### DIFF
--- a/nucliadb_utils/src/nucliadb_utils/storages/gcs.py
+++ b/nucliadb_utils/src/nucliadb_utils/storages/gcs.py
@@ -425,6 +425,7 @@ class GCSStorageField(StorageField):
             else:
                 return None
 
+    @storage_ops_observer.wrap({"type": "upload"})
     async def upload(self, iterator: AsyncIterator, origin: CloudFile) -> CloudFile:
         self.field = await self.start(origin)
         if self.field is None:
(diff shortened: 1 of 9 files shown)


nucliadb_utils/src/nucliadb_utils/storages/gcs.py
```diff
@@ -740,19 +740,19 @@ async def insert_object(self, bucket_name: str, key: str, data: bytes) -> None:
...
         if self.session is None:  # pragma: no cover
             raise AttributeError()
-        url = "{base_url}/{bucket_name}/o?name={object_name}&uploadType=media".format(
-            base_url=self.object_base_url,
-            bucket_name=bucket_name,
-            object_name=quote_plus(key),
-        )
+        url = self._upload_url.format(bucket=bucket_name)
+        params = {
+            "uploadType": "media",
```
jotare: Is this overwriting `uploadType=resumable` or sending both?

nucliadb_utils/src/nucliadb_utils/storages/storage.py
```diff
@@ -532,6 +525,19 @@ async def get_stream_message(self, key: str) -> bytes:
...
         await self.delete_upload(key, cast(str, self.indexing_bucket))
 
+    @abc.abstractmethod
+    async def insert_object(self, bucket: str, key: str, data: bytes) -> None:
+        """
+        Put some binary data into the object storage without any object metadata.
+        """
+        ...
+
+    async def upload(self, bucket: str, key: str, data: bytes) -> None:
```
jotare, line 535: I feel that you're introducing a new confusing concept here. The difference between inserting or uploading is unknown except if you read the implementation.
Could you go for something more explicit, knowing that `uploadbytes` is a multipart upload and `insert_object` a upload in one go?

jotare (approved):

PR #2723, opened 2024-12-12
Provide computed metadata from splits also

bloodbare:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -1164,6 +1164,9 @@ def add_field_classifications(basic: PBBasic, fcmw: FieldComputedMetadataWrapper
     fcfs = FieldClassifications()
     fcfs.field.CopyFrom(fcmw.field)
     fcfs.classifications.extend(fcmw.metadata.metadata.classifications)
+    for split_id, split in fcmw.metadata.split_metadata.items():
```
lferran, line 1167: I think you will need to change the condition of above (line 1159) aswell to something slightly more complex

PR #2727, opened 2024-12-12
Faster merging

javitonino:
### Description
* `madvise` calls were taking too long and offer very little value when merging/indexing a segment, since it will already be loaded into memory from all the copying of nodes. So we disable it in that case, while keeping it enabled when searching.
* After this, I noticed we were taking a significant chunk of time accessing the HashSet of visited nodes, so I switched over to a faster implementation.

Together, this means ~2x speedup of merges.

### DIFF
--- a/nidx/src/settings.rs
+++ b/nidx/src/settings.rs
@@ -174,7 +174,7 @@ pub struct MergeSettings {
 impl Default for MergeSettings {
     fn default() -> Self {
         Self {
-            max_deletions: 100,
+            max_deletions: 500,
             log_merge: Default::default(),
             vector_merge: Default::default(),
         }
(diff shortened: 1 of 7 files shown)


nidx/src/settings.rs
```diff
@@ -174,7 +174,7 @@ pub struct MergeSettings {
 impl Default for MergeSettings {
     fn default() -> Self {
         Self {
-            max_deletions: 100,
+            max_deletions: 500,
```
javitonino, line 177: This was supposed to be 1000, I set it to 100 during testing and forgot. I've left a lower default for now (500) and let's test it out.

jotare (approved):

PR #2740, opened 2024-12-18
NucliaDB tests unification: writer

jotare:
### Description
- Port writer tests to ndbfixtures
- Add a new magic fixture for component `nucliadb_writer`
- Add more typing to the tests

### How was this PR tested?
Tunned tests


nucliadb/tests/conftest.py
```diff
@@ -18,6 +18,7 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 pytest_plugins = [
+    "pytest_mock",
```
javitonino, line 21: Can we remove the `conftest.py` in the `writer` directory now?

nucliadb/tests/ndbfixtures/__init__.py
```diff
@@ -18,19 +18,28 @@
...
 
 # components and deployment modes
 
 from .reader import *  # noqa
+from .writer import *  # noqa
 
 # subcomponents
 
-from .common import *  # noqa
+from .common import *  # type: ignore # noqa
```
javitonino, line 39: Why? :thinking: I guess some things are getting overridden? In that case, could we not just stop doing `import *` of all  the things?

nucliadb/tests/ndbfixtures/writer.py
```diff
@@ -0,0 +1,197 @@
...
+    async with client_factory(roles=[NucliaDBRoles.WRITER]) as client:
+        yield client
+
+
+@pytest.fixture(scope="function")
+async def nucliadb_writer_manager(
+    nucliadb_writer: AsyncClient,
+) -> AsyncIterator[AsyncClient]:
+    roles = [NucliaDBRoles.MANAGER, NucliaDBRoles.WRITER]
+    nucliadb_writer.headers["X-NUCLIADB-ROLES"] = ";".join([role.value for role in roles])
```
javitonino, line 64: Could we use the `client_factory` for this fixture as well? It feels a bit cleaner than manipulating headers directly.

nucliadb/tests/ndbfixtures/writer.py
```diff
@@ -0,0 +1,197 @@
...
+    storage_driver = request.param
+    set_utility(Utility.STORAGE, storage_driver)
+
+    yield storage_driver
+
+    clean_utility(Utility.STORAGE)
+
+
+# FIXME: this is a weird situation, we can use a hosted-like nucliadb while this
+# creates a KB as it was onprem. The end result should change much but still, is
```
javitonino, line 145:
```suggestion
# creates a KB as it was onprem. The end result should not change much but still, is
```

javitonino (approved):
Some questions/comments, but looks good!

nucliadb/tests/ndbfixtures/writer.py
jotare: I'd prefer not to. Maybe it's not explicit enough but this is using a magic fixture `nucliadb_writer`, that is an AsyncClient to the component, standalone or whatever. It doesn't need to know

nucliadb/tests/conftest.py
jotare: Yeah, I'll change to a more explicit approach in conftest and avoid this hack

nucliadb/tests/ndbfixtures/__init__.py
jotare: Some hack I did in the past. I'll change to a more explicit approach in conftest and avoid this hack

nucliadb/tests/conftest.py
jotare: It is deleted. This is a general conftest for all migrated packages (currently reader and writer)

javitonino: Uhm, I don't see it deleted in this PR. :confused:

PR #2759, opened 2024-12-26
fix apparent inconsistencies between activity log and accounting

drf7:
### Description
https://app.shortcut.com/flaps/story/11343/fix-apparent-inconsistencies-between-activity-log-and-accounting

### How was this PR tested?

### DIFF
--- a/nucliadb_utils/src/nucliadb_utils/audit/stream.py
+++ b/nucliadb_utils/src/nucliadb_utils/audit/stream.py
@@ -58,7 +58,7 @@ class RequestContext:
     def __init__(self):
         self.audit_request: AuditRequest = AuditRequest()
         self.start_time: float = time.monotonic()
-        self.path = None
+        self.path: str = ""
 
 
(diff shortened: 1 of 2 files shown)


lferran (approved):
LGTM! 
Does it need some test to make sure we don't have any regression later?

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -58,6 +58,7 @@ class RequestContext:
     def __init__(self):
         self.audit_request: AuditRequest = AuditRequest()
         self.start_time: float = time.monotonic()
+        self.path: str = ""
```
jotare, line 61: Why do you need `self.path` for? I'm not finding where is this used

drf7: It is used below
Here:
`context.path = request.url.path`

And here:
```
if "/ask" in context.path:
    auditrequest.type = AuditRequest.CHAT
else:
    auditrequest.type = AuditRequest.SEARCH
```

PR #2764, opened 2025-01-02
Make sure broker messages from vectorset migrations are properly ingested

lferran:


lferran:
The problem is that `sentences_to_delete` and `paragraphs_to_delete` is populated and vectors and paragraphs are removed from all indexes.

nucliadb/tests/nucliadb/integration/test_vectorsets_api.py
```diff
@@ -127,3 +139,154 @@ async def test_learning_config_errors_are_proxied_correctly(
...
+            txn,
+            kbid=kbid,
+            config=VectorSetConfig(
+                vectorset_id="en-2024-05-06",
+                vectorset_index_config=VectorIndexConfig(
+                    similarity=utils_pb2.VectorSimilarity.COSINE,
+                    vector_dimension=1024,
+                ),
+            ),
+        )
```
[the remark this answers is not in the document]
lferran: maybe both are not needed indeed. feel free to change, I think it was me trying things out to reproduce as similar scenario

nucliadb/src/nucliadb/ingest/orm/brain.py
```diff
@@ -301,7 +301,11 @@ def apply_field_vectors(
 
         if replace_field:
             full_field_id = ids.FieldId(rid=self.rid, type=fid.type, key=fid.key).full()
-            self.brain.sentences_to_delete.append(full_field_id)
+            if vectorset is None:
+                # DEPRECATED
```
lferran, line 305: should we add a warning here? or will it be too noisy because we still have KBs on prod without vectorsets?

nucliadb/tests/nucliadb/integration/test_vectorsets_api.py
```diff
@@ -127,3 +138,140 @@ async def test_learning_config_errors_are_proxied_correctly(
...
+    await inject_message(nucliadb_grpc, bm2)
+
+    # Make a search with the new vectorset and check that the document is found
+    await _check_semantic_search(nucliadb_reader, kbid, vectorset="en-2024-05-06")
+
+    # With the default vectorset the document should also be found
+    await _check_semantic_search(nucliadb_reader, kbid)
+
+
+async def _check_semantic_search(
```
lferran, line 264: should we also check keyword search, just in case?
I wonder if the paragraphs/texts index is being deleted or not

nucliadb/src/nucliadb/ingest/orm/brain.py
jotare: Not sure, but I'd add a warning the day we want to finally remove the default vectorset concept

nucliadb/tests/nucliadb/integration/test_vectorsets_api.py
jotare: Sure, I'll fix the broker message and extend the test

PR #2771, opened 2025-01-10
Remove node

javitonino:
### Description

### How was this PR tested?


nucliadb_models/src/nucliadb_models/internal/shards.py
```diff
@@ -63,6 +63,7 @@ class ShardReplica(BaseModel):
 
 class ShardObject(BaseModel):
     shard: str
+    nidx_shard_id: Optional[str]
```
lferran, line 66: shouldn't this be a mandatory attr?

nucliadb_models/src/nucliadb_models/internal/shards.py
```diff
@@ -63,6 +63,7 @@ class ShardReplica(BaseModel):
...
     replicas: List[ShardReplica]
```
lferran, line 67: can be removed?

lferran (approved):

nucliadb_models/src/nucliadb_models/internal/shards.py
javitonino: looks like it can be removed but it's annoying and breaks some tests. Maybe for the second pass?

PR #2777, opened 2025-01-15
Add storage key part to vectorsets + migration

jotare:
### Description

### How was this PR tested?


nucliadb_protos/knowledgebox.proto
```diff
@@ -212,6 +212,13 @@ message VectorSetConfig {
     // list of possible subdivisions of the matryoshka embeddings (if the model
     // supports it)
     repeated uint32 matryoshka_dimensions = 3;
+
+    enum StorageKeyKind {
+        UNSET = 0;              // for Bw/c during the migration
+        LEGACY = 1;             // simply "extracted_vectors", without vectorset
+        VECTORSET_PREFIX = 2;   // "{vectorset}/extracted_vectors"
+    }
+    StorageKeyKind storage_field_key = 4;
```
lferran, line 221: do we want to store only the kind, or the key itself?

lferran: if it's just the kind, i would change the attribute name to `storage_key_kind` or `vectorset_storage_key_type`

lferran (approved):

nucliadb_protos/knowledgebox.proto
jotare: I think provides us with enough information to use a key or another depending on it and we'll save storage (an enum variant instead of a string per vectorset per KB). Also, the code using this will be less obscure to trace (you'll read a hardcoded string in the code, not a value in the DB)

PR #2779, opened 2025-01-15
Broker message generator

javitonino:
### Description

### How was this PR tested?


nucliadb/src/migrations/0028_backfill_field_status.py
```diff
@@ -0,0 +1,93 @@
...
+                # A resource here is in error state, set the field status with the error
+                status = writer_pb2.FieldStatus(
+                    status=writer_pb2.FieldStatus.Status.ERROR,
+                )
+                field_error = writer_pb2.FieldError(
+                    source_error=error,
+                )
+                field_error.created.GetCurrentTime()
+                status.errors.append(field_error)
+                status_key = key[:-5] + "status"
```
lferran, line 89: replace, to be more explicit? 😛

lferran (approved):

nucliadb/src/migrations/0028_backfill_field_status.py
javitonino: Well, replace has the problem that the field may be named "error" and we break it completely. Maybe I can do a split by /, replace last element and join again

nucliadb/src/migrations/0029_backfill_field_status.py
```diff
@@ -0,0 +1,149 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""Migration #28
```
jotare, line 21:
```suggestion
"""Migration #29
```

jotare (approved):

PR #2781, opened 2025-01-16
Vectorset storage field kind migration

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/migrations/0028_extracted_vectors_reference.py
+++ b/nucliadb/src/migrations/0028_extracted_vectors_reference.py
@@ -1,0 +1,60 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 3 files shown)


nucliadb/src/migrations/0028_extracted_vectors_reference.py
```diff
@@ -0,0 +1,60 @@
...
+async def migrate(context: ExecutionContext) -> None: ...
+
+
+async def migrate_kb(context: ExecutionContext, kbid: str) -> None:
+    async with datamanagers.with_rw_transaction() as txn:
+        vectorsets = [vs async for (_vid, vs) in datamanagers.vectorsets.iter(txn, kbid=kbid)]
+
+        if len(vectorsets) == 0:
+            # should never happen, everyone should have at least one
+            logger.warning(f"KB has no vectorsets!", extra={"kbid": kbid})
```
lferran, line 44: nit: `continue` here, otherwise it commits at the end.

Should we fail hard if this happens though? This is what you already checked in prod right?

lferran (approved):

nucliadb/src/migrations/0028_extracted_vectors_reference.py
jotare: Ups, adding it!
No, I think is fine with a log. I'll check it when it runs just in case but yeah, is what I validated so nobody should have a KB like this

PR #2790, opened 2025-01-20
Rewrite writer virtual service rules to be more strict and avoid collision

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/charts/nucliadb_writer/templates/writer.vs.yaml
+++ b/charts/nucliadb_writer/templates/writer.vs.yaml
@@ -10,12 +10,6 @@ spec:
   http:
     - name: nucliadb_writer
       match:
-        # KBs (only for onprem)
-        - uri:
-            regex: '^/api/v\d+/kb/[^/]+'
-          method:
-            regex: "POST|PATCH|PUT|DELETE|OPTIONS"
-
         # KB config
         - uri:
             regex: '^/api/v\d+/kb/[^/]+/configuration'
--- a/charts/nucliadb_writer/templates/writer.vs.yaml
+++ b/charts/nucliadb_writer/templates/writer.vs.yaml
@@ -10,22 +10,68 @@ spec:
   http:
     - name: nucliadb_writer
       match:
+        # KBs (only for onprem)
         - uri:
             regex: '^/api/v\d+/kb/[^/]+'
           method:
             regex: "POST|PATCH|PUT|DELETE|OPTIONS"
-        - uri:
-            regex: '^/api/v\d+/kb/[^/]+/(resource|slug|resources|entitiesgroup|labelset|upload|custom-synonyms|export|import).*'
-          method:
-            regex: "POST|PATCH|PUT|DELETE|OPTIONS"
+
+        # KB config
         - uri:
             regex: '^/api/v\d+/kb/[^/]+/configuration'
           method:
             regex: "POST|PATCH|OPTIONS"
         - uri:
-            regex: '^/api/v\d+/kb/[^/]+/vectorsets/.*'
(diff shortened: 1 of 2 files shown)


charts/nucliadb_writer/templates/writer.vs.yaml
```diff
@@ -10,22 +10,68 @@ spec:
   http:
     - name: nucliadb_writer
       match:
+        # KBs (only for onprem)
```
javitonino, line 13: I'm not sure if this is needed here.
Also, the POST `/kbs` and `/kbs/import` endpoint are missing (and has always been).

No action needed, just thoroughly going through all routes.

charts/nucliadb_writer/templates/writer.vs.yaml
```diff
@@ -10,22 +10,68 @@ spec:
...
+          method:
+            regex: "POST|OPTIONS"
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/(resource|slug)/[^/]+/file/[^/]+/reprocess'
+          method:
+            regex: "POST|OPTIONS"
+
+        # misc
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/(entitiesgroup|labelset|custom-synonyms|export|import).*'
```
javitonino, line 71: I'd split:
- `PATCH/DELETE/OPTIONS entitiesgroup/[^/]+`.
- `POST/DELETE/OPTIONS labelset/[^/]+`.

That leaves the following which can be made into one or two routes:
`PUT/DELETE/OPTIONS custom-synonyms`
`POST/OPTIONS export/import`

charts/nucliadb_writer/templates/writer.vs.yaml
```diff
@@ -10,22 +10,68 @@ spec:
...
+        # resources
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/resources'
+          method:
+            regex: "POST|OPTIONS"
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/(resource|slug)/[^/]+$'
+          method:
+            regex: "PATCH|DELETE|OPTIONS"
+
```
javitonino, line 38: Missing route?

`PATCH/DELETE/OPTIONS /(resource|slug)/{rid}`.

for
modify_resource_rslug_prefix
modify_resource_rid_prefix
delete_resource_rslug_prefix
delete_resource_rid_prefix

charts/nucliadb_writer/templates/writer.vs.yaml
```diff
@@ -10,22 +10,68 @@ spec:
...
+            regex: '^/api/v\d+/kb/[^/]+/(resource|slug)/[^/]+/file/[^/]+/reprocess'
+          method:
+            regex: "POST|OPTIONS"
+
+        # misc
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/(entitiesgroup|labelset|custom-synonyms|export|import).*'
+          method:
+            regex: "POST|PATCH|PUT|DELETE|OPTIONS"
+
```
javitonino, line 74: Would also be good to remove `.*` at the end of the tusupload routes. But that's way less likely to collide, so not urgent.

javitonino (changes requested):

javitonino (approved):

charts/nucliadb_writer/templates/writer.vs.yaml
javitonino: me being very blind

charts/nucliadb_writer/templates/writer.vs.yaml
jotare: Done

PR #2809, opened 2025-01-24
Start publishing nidx_binding (test)

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/.github/workflows/nidx_binding_release.yml
+++ b/.github/workflows/nidx_binding_release.yml
@@ -49,6 +49,10 @@ jobs:
       - name: Check system deps
         run: make -C nucliadb check-system
 
+      - name: Install protoc (mac)
+        if: ${{ matrix.os == 'macos' }}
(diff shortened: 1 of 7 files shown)


.github/workflows/nidx_binding_release.yml
```diff
@@ -1,73 +1,34 @@
...
-            target: aarch64
-            container: messense/manylinux_2_24-cross:aarch64
-          # musllinux
-          - os: ubuntu
-            platform: linux
             target: x86_64
-            manylinux: musllinux_1_1
-          - os: ubuntu
-            platform: linux
+          - os: macos
```
javitonino, line 30: Removed a lot of platforms that we were not using and that were going to be a pain to make work with newer rust compilers. Left the two main platforms: linux on x86_64 and mac on arm64 and the source distribution so others can install from source.

lferran (approved):

PR #2824, opened 2025-01-29
Test implement sdk

bloodbare:
### Description

### How was this PR tested?


nucliadb_sdk/src/nucliadb_sdk/v2/sdk.py
```diff
@@ -95,31 +99,315 @@
...
+    "get_custom_synonyms": SDKDefinition(
+        path_template="/v1/kb/{kbid}/custom-synonyms",
+        method="GET",
+        path_params=("kbid",),
+    ),
+}
 
 
-def ask_response_parser(response: httpx.Response) -> SyncAskResponse:
+def ask_response_parser(response_type: Type[BaseModel], response: httpx.Response) -> BaseModel:
```
lferran, line 403: why do you need to pass `response_type` as a parameter here? isn't it always `SyncAskResponse`?
I think the return type of this function should be `SyncAskResponse`.

And also, for the async version, I think we shuld be iterating the response asynchronously no? (that wasnt implemented before either)

lferran (commented):
Great improvement! finally this will fix the problems we had with typings.
Left a few comments

nucliadb_sdk/src/nucliadb_sdk/v2/sdk.py
bloodbare: I'm not changing the API SDK behavior on this PR, just fixing the typing. Can not be SyncAskResponse because the Generic is bound to BaseModel.

nucliadb_sdk/src/nucliadb_sdk/v2/sdk.py
```diff
@@ -521,7 +525,7 @@ def prepare_request(
             elif not isinstance(content, request_type):
                 raise TypeError(f"Expected {request_type}, got {type(content)}")
             elif isinstance(content, BaseModel):
-                data = content.model_dump_json(by_alias=True, exclude_unset=True)
+                data = content.model_dump_json(by_alias=True, exclude_unset=False)
```
jotare, line 528: Why are you changing `exclude_unset` here? This affects the ability to change public models and makes bw/c harder in nucliadb

bloodbare: Because otherwise RagStrategy objects does not work properly unless I set the name of the Pydantic object. But I can force it if its a problem.

javitonino (approved):

PR #2834, opened 2025-02-03
GraphRAG: Better search for entities

javitonino:
Moves the logic to search for entities from graph strategy to nidx, so it can be done faster and in a single index request.


nucliadb_protos/nodereader.proto
```diff
@@ -254,7 +254,11 @@ message RelationEdgeFilter {
 }
 
 message RelationPrefixSearchRequest {
+    // Search for a single prefix
     string prefix = 1;
+    // Search for entities appearing on a query
+    optional string query = 3;
```
jotare, line 260: What do you think of instead of adding the `query` field as an implicit oneof with `prefix`, rename `prefix` to query and add an enum or number to define how many words do we want to use (last 3, all of them...)

jotare (approved):

PR #2841, opened 2025-02-04
Raise errors from apply_for_all_shards

jotare:
### Description
`apply_for_all_shards` was returning a list of exceptions that could be ignored. Be more explicit and raise an error if there is one.

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/cluster/manager.py
+++ b/nucliadb/src/nucliadb/common/cluster/manager.py
@@ -104,6 +104,14 @@ class KBShardManager:
             errors.capture_exception(exc)
             raise NodeError("Node unavailable for operation") from exc
 
+        for result in results:
+            if isinstance(result, Exception):
+                errors.capture_exception(result)
+                raise NodeError(
+                    f"Error while applying {aw.__name__} for all shards. Other similar errors may have been shadowed.\n"
+                    f"{type(result).__name__}: {result}"
+                ) from result
+
         return results
 
     # TODO: move to data manager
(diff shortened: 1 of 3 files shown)


nucliadb/src/nucliadb/ingest/orm/entities.py
```diff
@@ -316,21 +311,18 @@ async def query_indexed_entities_group_names(node: AbstractIndexNode, shard_id:
             response: SearchResponse = await query_shard(node, shard_id, request)
             try:
                 facetresults = response.document.facets["/e"].facetresults
-                return {facet.tag.split("/")[-1] for facet in facetresults}
             except KeyError:
                 # No entities found
                 return set()
+            else:
+                return {facet.tag.split("/")[-1] for facet in facetresults}
```
javitonino, line 318: Is this intended? This changes means if the KeyError occurs in this line, it will be raised instead of ignored. Probably OK, just double-checking.

nucliadb/src/nucliadb/common/cluster/manager.py
```diff
@@ -104,6 +104,14 @@ async def apply_for_all_shards(
             errors.capture_exception(exc)
             raise NodeError("Node unavailable for operation") from exc
 
+        for result in results:
+            if isinstance(result, Exception):
+                errors.capture_exception(result)
+                raise NodeError(
```
javitonino, line 110: I think this is a good idea. Too easy to ignore error otherwise.

javitonino (approved):
Cool cleanup. This may make other errors more visible as well which is pretty nice.

nucliadb/src/nucliadb/ingest/orm/entities.py
jotare: Yes, if we don't find `/e` means there's no entities, but a `KeyError` here would mean indexing changed or is wrong, we should notice it

PR #2843, opened 2025-02-04
Purge segments from deleted indexes

jotare:
### Description
In a really specific case where a shard is deleted while the merger has a new job for it, it can happen that a new segment without `delete_at` is created for an index to delete. This will cause the segment, index and shard to never be deleted.

### How was this PR tested?


nidx/tests/test_shards_api.rs
```diff
@@ -105,6 +107,11 @@ async fn test_shards_create_and_delete(pool: sqlx::PgPool) -> anyhow::Result<()>
         sqlx::query!("UPDATE segments SET delete_at = NOW() WHERE index_id = $1", index_id as IndexId,)
```
javitonino, line 107: I think this can be deleted now, might make the test more clear.

nidx/tests/test_shards_api.rs
```diff
@@ -105,6 +107,11 @@ async fn test_shards_create_and_delete(pool: sqlx::PgPool) -> anyhow::Result<()>
...
             .execute(&meta.pool)
             .await?;
+
+        // Add a new segment as if it was coming from a merge job started before
+        // deletion and finished just before it. The segment has no deletion
+        // mark but shouldn't retain himself, its index or shard to be purged
+        Segment::create(&meta.pool, index_id, Seq::from(1000i64), 100, JsonValue::Null).await?;
```
javitonino, line 114: I'd mark the newly created segment as ready. Newly created segments are marked for deletion in the future so that might make this test less robust.

jotare: Done

PR #2856, opened 2025-02-07
Add security group header

sunbit:
### Description
When using the new Temporal service accounts feature, that can have embedded security groups, Authorizer will forward this to nucliad, for now only on the ask endpoint. This ensures that if present, this value is set and maybe overrides the value of the payload

### How was this PR tested?
the existing security tests were done against standalone, that doesn't use the auth backend used in the cloud setup, so a mocked test to ensure the injection is done has ben set at a search component level

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/ask.py
+++ b/nucliadb/src/nucliadb/search/api/v1/ask.py
@@ -66,6 +66,7 @@ async def ask_knowledgebox_endpoint(
     current_user: NucliaUser = request.user
     # If present, security groups from AuthorizationBackend overrides any
     # security group of the payload
+    print(current_user)
     if current_user.security_groups:
         if item.security is None:
             print(f"item security={item.security}, header security={current_user.security_groups}")
--- a/nucliadb/src/nucliadb/search/api/v1/ask.py
+++ b/nucliadb/src/nucliadb/search/api/v1/ask.py
@@ -68,9 +68,12 @@ async def ask_knowledgebox_endpoint(
     # security group of the payload
     if current_user.security_groups:
         if item.security is None:
+            print(f"item security={item.security}, header security={current_user.security_groups}")
             item.security = RequestSecurity(groups=current_user.security_groups)
         else:
+            print(f"item security={item.security}, header security={current_user.security_groups}")
             item.security.groups = current_user.security_groups
+    print(f"POST item security={item.security}")
     return await create_ask_response(
         kbid, item, x_nucliadb_user, x_ndb_client, x_forwarded_for, x_synchronous
     )
--- a/nucliadb/src/nucliadb/search/api/v1/ask.py
+++ b/nucliadb/src/nucliadb/search/api/v1/ask.py
@@ -36,7 +36,8 @@ from nucliadb_models.search import (
     SyncAskResponse,
     parse_max_tokens,
 )
-from nucliadb_utils.authentication import requires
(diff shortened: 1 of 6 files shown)


nucliadb_utils/src/nucliadb_utils/authentication.py
```diff
@@ -66,7 +69,17 @@ async def authenticate(self, request: HTTPConnection) -> Optional[Tuple[AuthCred
...
         if self.user_header in request.headers:
             user = request.headers[self.user_header]
-            nuclia_user: BaseUser = NucliaUser(username=user)
+            nuclia_user: NucliaUser = NucliaUser(username=user)
+
+            raw_security_groups: Optional[str] = request.headers.get(self.security_groups_header)
+
+            security_groups: Optional[list[str]] = None
+            if raw_security_groups is not None:
+                security_groups = raw_security_groups.split(";")
```
jotare, line 78: This parsing implies `;` is not a valid character for security groups, but we're not enforcing it in the API models

nucliadb_utils/src/nucliadb_utils/authentication.py
```diff
@@ -32,8 +32,9 @@
 
 
 class NucliaUser(BaseUser):
-    def __init__(self, username: str) -> None:
+    def __init__(self, username: str, security_groups: Optional[list[str]] = None) -> None:
         self.username = username
+        self.security_groups = security_groups
```
jotare, line 37: As endpoints will directly access this field, maybe we should use a read only property instead

nucliadb/src/nucliadb/search/api/v1/ask.py
```diff
@@ -62,6 +63,15 @@ async def ask_knowledgebox_endpoint(
         "This is slower and requires waiting for entire answer to be ready.",
     ),
 ) -> Union[StreamingResponse, HTTPClientError, Response]:
+    current_user: NucliaUser = request.user
+    # If present, security groups from AuthorizationBackend overrides any
+    # security group of the payload
```
jotare, line 68: Should we document this behavior in the /ask endpoint?

nucliadb_utils/src/nucliadb_utils/authentication.py
jotare: Security models are defined here: https://github.com/nuclia/nucliadb/blob/main/nucliadb_models/src/nucliadb_models/security.py

sunbit: OR i can send the value in JSON with a proper list, what do you think?

nucliadb/src/nucliadb/search/api/v1/ask.py
sunbit: i dont think so, but i'll explain on the document about security groups where i'll introduce the use of these new tokens. from a "normal" use of the ask endpoint that cannot happen.

if you're using any kind of authorization proxy, (this tokens is exactly this in some sense) providing security groups directly from the end-user requests doesn't make much sense. In fact, probably this parameter shouldn't be exposed at all if we had a "cloud only" docs. but thats another story.

nucliadb_utils/src/nucliadb_utils/authentication.py
sunbit: sure

jotare (approved):

PR #2867, opened 2025-02-12
Add automatic pydantic conversions for resources and utils proto

carlesonielfa:
### Description
Adds automatic pydantic conversions for resources and utils proto by using the library `protobuf-to-pydantic`.

The protos for multiple messages in resources.proto are needed for the develompent of predict data augmentation endpoint as the protos need to be sent through the REST API

The conversion had a couple of issues that I was able to fix by reordering a message and creating a dummy one to trigger a missing import. (https://github.com/so1n/protobuf_to_pydantic/issues/83)

### How was this PR tested?


carlesonielfa:
No longer required

PR #2869, opened 2025-02-12
Query entities: a parameter to find your entities

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/query.py
+++ b/nucliadb/src/nucliadb/search/search/query.py
@@ -49,6 +49,7 @@ from nucliadb_models.labels import LABEL_HIDDEN, translate_system_to_alias_label
 from nucliadb_models.metadata import ResourceProcessingStatus
 from nucliadb_models.search import (
     Filter,
+    KnowledgeGraphEntity,
     MaxTokens,
     MinScore,
     SearchOptions,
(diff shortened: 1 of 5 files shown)


nucliadb/src/nucliadb/search/search/query.py
```diff
@@ -372,8 +376,22 @@ async def parse_vector_search(self, request: nodereader_pb2.SearchRequest) -> bo
...
+        # BUG: autofiler should autofilter, not enable relation search
         if self.has_relations_search or self.autofilter:
-            detected_entities = await self.fetcher.get_detected_entities()
+            if self.query_entities:
+                detected_entities = []
+                for entity in self.query_entities:
+                    relation_node = utils_pb2.RelationNode()
+                    relation_node.value = entity.name
+                    if entity.type is not None:
+                        relation_node.ntype = utils_pb2.RelationNode.NodeType.Value(
```
javitonino, line 387: Can we use the models we have for this?

Change the KnowledgeGraphEntity.type to be nucliadb_models.metadata.RelationNodeType

and here use the conversion map `RelationNodeTypeMap` to convert to the protobuffer type.

javitonino (approved):

nucliadb/src/nucliadb/search/search/query.py
jotare: Sure

jotare:
[sc-11677]

PR #2870, opened 2025-02-12
Return rephrased query in results

javitonino:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/search/search/find.py
```diff
@@ -120,6 +120,7 @@ async def _index_node_retrieval(
             results,
             kbid=kbid,
             query=pb_query.body,
+            rephrased_query=await query_parser.fetcher.get_rephrased_query(),
```
jotare, line 123: The fetcher inside the `QueryParser` class was more a middle step while moving thing to `query_parser` module, I don't love exposing it here.

jotare (approved):

nucliadb/src/nucliadb/search/search/find.py
javitonino: I can proxy the get_rephrased_query function

PR #2881, opened 2025-02-14
Nucliadb indexed notification

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/tests/nucliadb/integration/test_notifications.py
+++ b/nucliadb/tests/nucliadb/integration/test_notifications.py
@@ -19,30 +19,46 @@
 #
 import asyncio
 from typing import Union
+from unittest.mock import patch
 
 import httpx
 import pytest
 import uvicorn
 from httpx import AsyncClient
-from tests.ingest.fixtures import broker_resource
-from tests.utils import inject_message
 
 from nucliadb.reader.api.v1.router import KB_PREFIX
+from nucliadb.reader.app import create_application
 from nucliadb_models.notifications import (
     Notification,
     ResourceIndexedNotification,
     ResourceProcessedNotification,
     ResourceWrittenNotification,
 )
-from nucliadb_protos import writer_pb2
 from nucliadb_protos.writer_pb2 import BrokerMessage
+from nucliadb_utils.settings import transaction_settings
+from tests.ingest.fixtures import broker_resource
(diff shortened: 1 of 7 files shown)


nucliadb/tests/nucliadb/integration/test_notifications.py
```diff
@@ -0,0 +1,172 @@
...
+    storage,
+    maindb_driver,
+    local_files,
+    indexing_utility,
+    nidx,
+):
+    application = create_application()
+    with patch.object(transaction_settings, "transaction_local", True):
+        async with application.router.lifespan_context(application):
+            yield application
```
jotare, line 54: Maybe you could, instead of redefining a "private" fixture, call the original and then pass a nidx_utility fixture which overwrites the utility.

nucliadb/tests/nucliadb/integration/test_notifications.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@pytest.fixture(scope="function")
+async def nucliadb_stream_reader(nidx_reader_api_server):
```
jotare, line 58: Do you think it is worth to do this for everyone or the cost is not worth it?

nucliadb/tests/nucliadb/integration/test_notifications.py
javitonino: I tried it at some point and it didn't work, due to the order of initialization of several fixtures. Since then,. I removed a bunch of them (e.g: nucliadb_writer), so let's try it again.

nucliadb/tests/nucliadb/integration/test_notifications.py
javitonino: It's quite costly/slow, it open an HTTP server and closes it. I'd rather not do it when not necessary.

nucliadb/tests/nucliadb/integration/test_notifications.py
javitonino: Okay, it worked fine, thanks for the tip!

I also moved this to the search directory since it uses the nidx image and it's only built in CI in that environment. Not the cleanest, but I don't think that's horrible, is it?

jotare (approved):

PR #2885, opened 2025-02-18
Knowledge graph query

jotare:
### Description

### How was this PR tested?


nidx/nidx_relation/src/graph_query_parser.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    fn parse_path_query(&self, query: PathQuery) -> Box<dyn Query> {
+        match query {
+            PathQuery::DirectedPath((source, relation, destination)) => {
+                match (source, relation, destination) {
+                    // AND expressions like `(A & B)-[R]->C` must be splitted and queried as an
+                    // intersection, as there is no document that satisfies a source `A & B` at the
+                    // same time (as they'll be different triplets). However, `(A & B)-[R]->C` is
+                    // equivalent to `A-[R]->C & B-[R]->C`, so we can query the intersection of
+                    // both.
```
javitonino, line 139: I think this only makes sense in the case where the And'ed conditions do not overlap, ie:
`AND(PERSON & Anastasia)`
but it does not makes sense when they overlap
`AND(PERSON & NUMBER)`

And for the case they don't overlap, you can just set all the information in the Node. That is, your test example (Anastasia & PERSON) is just `Node { subtype: "PERSON", value: "Anastasia" }`

I can't find any real usecase for this And splitting and it adds a ton of complexity. I'd consider removing it.

I'm also not sure about the explanation about the split. For the example case, I think it generates the same:
`Must(Must(node_subtype=PERSON), Must(node_value=Anastasia))`?

In more complex cases, as you mention, this changes the query from `A&B -> C` to `A -> C & B -> C`. But still, it tries to match a single relation that matches both conditions (there is no single relation that satisfies A -> C and B -> C at the same time if A & B are exclusive). i.e: the intersection of A->C & B->C is empty if the intersection of A&B is empty.

Anyway, I think this is a lot of complexity and does not change anything at all in the results.

It feels like just removing all the And-specific code will result in the same queries.

nidx/nidx_relation/src/graph_query_parser.rs
```diff
@@ -0,0 +1,326 @@
...
+                                self.has_node_as_destination(&query)
+                            }
+                            Expression::Or(queries) => {
+                                let subqueries = queries.into_iter().map(|query| self.has_node_as_destination(&query));
+                                Box::new(BooleanQuery::union(subqueries.collect()))
+                            }
+                            Expression::And(_) => unreachable!("Pattern already matched in the outer match"),
+                        };
+
+                        Box::new(BooleanQuery::new(vec![
```
javitonino, line 221: You could also build this vec! dynamically to avoid having to add `Must(AllQuery)` terms to the query. 

This changes would also have to be made in  `has_node_as_destination`-style methods:
1. They also have a fixed number of queries (with AllQuery), we could make them build a vec of queries dynamically like I suggest here. 
2.  They could return an Option::None if the subquery vector is empty

And here, if the query is None, just don't add it to the vector of subqueries.

It probably makes the code a bit uglier, but it makes query parsing/evaluation a bit faster, saves from iterating a few fastfields.

Additionally, a quirck of tantivy is that if all of the `occurs` are `MustNot`, the query won't run (it just returns empty set). In this case adding a `Must(AllQuery)` is actually needed.

nidx/nidx_relation/src/reader.rs
```diff
@@ -59,6 +60,24 @@ impl RelationsReaderService {
             prefix: self.prefix_search(request)?,
         })
     }
+
+    pub fn advanced_graph_query(&self, query: GraphQuery) -> anyhow::Result<Vec<nidx_protos::Relation>> {
+        let query_parser = GraphQueryParser::new();
+        let index_query: Box<dyn Query> = query_parser.parse_graph_query(query);
+        // println!("Index query: {index_query:#?}");
```
javitonino, line 67: At some point :fire:

nidx/nidx_relation/src/graph_query_parser.rs
jotare: Yeah, maybe I went a bit too far with the `AND` trying to allow a hidden multi-statement query. The interesting things you really want to query are for example: does Anna and Peter live in the same place? 
You want the triplets for Anna and Peter living in New York iff both live there. So I want the triplet `(Peter, LIVE_IN, New York)` and `(Anna, LIVE_IN, New York)`. But that in fact, this is a hidden `AND` between two queries.

So I agree it adds lots of complexity and if we need the use case, we could implement multi-statement boolean expressions instead. And as you say, for non-overlapping fields, you can write them in the same node expression. 

And yeah, I think the implementation is off, it was the last one I did and didn't add proper tests yet

nidx/nidx_relation/src/graph_query_parser.rs
jotare: I've refactored a bunch of stuff and reduced the amount of `Must(AllQuery)` I spread in the query. Tell me if there are more improvements you think of

nidx/nidx_relation/src/graph_query_parser.rs
jotare: And removed as needs multi-hop query and multi-shard aggregation. I've added multi-statement with OR and NOT and maintained the OR and NOT logic for node/relation so almost any 1-hop query should be representable (AND involving postprocessing for the user)

jotare:
[sc-11736]

nidx/nidx_relation/src/graph_query_parser.rs
```diff
@@ -0,0 +1,508 @@
...
+pub struct GraphSearcher {
+    schema: Schema,
+    parser: GraphQueryParser,
+    searcher: Searcher,
+}
+
+#[derive(Clone)]
+pub struct FuzzyTerm {
+    pub value: String,
+    pub fuzzy_distance: Option<u8>,
```
javitonino, line 38: Why is this an Option? Feels like it could always be specified (you have decided to do fuzzy search anyway).

javitonino (approved):

nidx/nidx_relation/src/graph_query_parser.rs
jotare: Yup, I left it from a previous design. Removing it

PR #2891, opened 2025-02-19
Cleanup feature flags

lferran:
### Description
- rebalance flag: already implemented long time go
- nats sync ack: not needed -- was just part of a debugging session
- wait for index: not used since > a year
- nidx_reads: already implemented
- field statsu: already implemented 
- versioned private predict: not needed anymore. learning stayed unversioned in the end
- read replica searches, no longer needed after nidx

### How was this PR tested?
existing tests

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/catalog.py
+++ b/nucliadb/src/nucliadb/search/api/v1/catalog.py
@@ -79,7 +79,6 @@ async def catalog_get(
     sort_order: SortOrder = fastapi_query(SearchParamDefaults.sort_order),
     page_number: int = fastapi_query(SearchParamDefaults.catalog_page_number),
     page_size: int = fastapi_query(SearchParamDefaults.catalog_page_size),
-    shards: list[str] = fastapi_query(SearchParamDefaults.shards, deprecated=True),
     with_status: Optional[ResourceProcessingStatus] = fastapi_query(
         SearchParamDefaults.with_status, deprecated="Use filters instead"
     ),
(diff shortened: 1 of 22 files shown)


nucliadb/src/nucliadb/search/api/v1/catalog.py
```diff
@@ -79,7 +79,6 @@ async def catalog_get(
     sort_order: SortOrder = fastapi_query(SearchParamDefaults.sort_order),
     page_number: int = fastapi_query(SearchParamDefaults.catalog_page_number),
     page_size: int = fastapi_query(SearchParamDefaults.catalog_page_size),
-    shards: list[str] = fastapi_query(SearchParamDefaults.shards, deprecated=True),
```
jotare, line 82: Even if it does nothing, have we communicated the removal of this parameter?

nucliadb/src/nucliadb/search/api/v1/resource/search.py
```diff
@@ -76,7 +76,6 @@ async def resource_search(
     highlight: bool = fastapi_query(SearchParamDefaults.highlight),
     x_ndb_client: NucliaDBClientType = Header(NucliaDBClientType.API),
     debug: bool = fastapi_query(SearchParamDefaults.debug),
-    shards: list[str] = fastapi_query(SearchParamDefaults.shards),
```
jotare, line 79: Same as above ^

javitonino: I don't think we even need to communicate removal from Find/Ask Request. If they send it, they will be ignored and behave exactly as they do now. That is, nothing changes at all for the user.

lferran: the "problem" is that those same model defs are also being used on the python sdks, so if someone blindly upgrades will start failing. maybe I am being too paranoid now

javitonino (approved):

PR #2892, opened 2025-02-19
Expose arbitrary filter expression in nidx api

javitonino:


nidx/nidx_text/tests/test_search.rs
```diff
@@ -307,4 +353,101 @@ fn test_int_order_pagination() {
...
+    );
+    assert_eq!(
+        search(ResourceFilter {
+            resource_id: "fake".into(),
+        }),
+        0
+    );
+}
+
+// TODO: order, only_faceted, with_status
```
jotare, line 453: with_status only apply to catalog right?

jotare (approved):
As we commented, Some expressions feel too verbose, but with a struct instead of pb we'll solve it. Looks really nice!

nidx/nidx_text/tests/test_search.rs
javitonino: Yeah, in the nucliadb PR I'm removing it.

PR #2914, opened 2025-02-25
Knowledge graph query: proto layer graph search

jotare:
### Description

### How was this PR tested?


nucliadb_protos/nodereader.proto
```diff
@@ -186,6 +186,79 @@ message EntitiesSubgraphResponse {
...
+    GraphQuery query = 2;
+
+    // decides what do we want in return: paths (triplets), nodes or relations.
+    // Returning paths implies returning nodes and relations.
+    enum QueryKind {
+        PATH = 0;
+        NODES = 1;
+        RELATIONS = 2;
+    }
+    QueryKind kind = 3;
```
javitonino, line 236: As far as I can see, this is unused and we always return everything.

It also means that setting `kind = NODES` and `top_k = 10` is kind of a lie, and it may return the same entity 10 times (since it always returns triplets).

Is this expected?

javitonino (approved):

nucliadb_protos/nodereader.proto
jotare: Kind is not used yet. The repeated entities should be implemented with a custom collector

PR #2917, opened 2025-02-25
Catalog expression

javitonino:
Allow using filter_expressions at catalog


nucliadb/src/nucliadb/search/api/v1/catalog.py
```diff
@@ -93,23 +99,32 @@ async def catalog_get(
...
+            range_creation_start=range_creation_start,
+            range_creation_end=range_creation_end,
+            range_modification_start=range_modification_start,
+            range_modification_end=range_modification_end,
+            hidden=hidden,
+        )
+        if sort_field:
+            item.sort = SortOptions(field=sort_field, limit=sort_limit, order=sort_order)
+        return await catalog(kbid, item)
+    except ValidationError as exc:
```
lferran, line 125: Is the `return await catalog()` line ok to be inside this try catch? I wonder if it would be masking errors on the serialization of the response as if they were client errors.

Also, should the InvalidQueryError be catched aswell?

lferran (approved):

nucliadb/src/nucliadb/search/api/v1/catalog.py
javitonino: `InvalidQueryError` is handled inside the `catalog` function. I reduced the try scope so it only catches the expr parsing.

PR #2928, opened 2025-02-28
Remove get index nodes

lferran:
### Description
Removing some of the old index node logic

### How was this PR tested?
Existing tests


nucliadb/tests/ndbfixtures/search.py
```diff
@@ -138,37 +136,15 @@ async def wait_for_shard(knowledgebox_ingest: str, count: int) -> str:
...
-                break
-            await asyncio.sleep(1)
-
+    checks[""] = False
+    req = GetShardRequest()
+    req.shard_id.id = shard.nidx_shard_id
+    for i in range(30):
+        count_shard: Shard = await nidx_api.GetShard(req)  # type: ignore
+        if count_shard.fields >= count:
+            checks[""] = True
```
javitonino, line 145: You can also remove the `checks` dictionary here now :fire:

nucliadb/tests/search/integration/api/v1/test_suggest.py
```diff
@@ -47,15 +47,14 @@ async def test_suggest_resource_all(
...
         async for key in txn.keys(key):
             value = await txn.get(key)
             assert value is not None
             shards = PBShards()
             shards.ParseFromString(value)
             for replica in shards.shards[0].replicas:
-                node_obj = INDEX_NODES.get(replica.node)
-
-                if node_obj is not None:
+                if True:
```
javitonino, line 57: I think you can remove this loop (`for replica in shards.shards[0].replicas:`) entirely. Shard replicas will always be empty now. We probably want to keep the asserts inside, so maybe replace the loop by a single query to nidx fake node.

javitonino (approved):
Nice cleanup! Suggested a couple of extra changes.

PR #2934, opened 2025-03-04
Improve tasks and export import

lferran:
### Description
- Mostly typing improvements
- Removed unused task registry stuff
- Separate the stream subjects from the consumer subject: this makes it easier to integrate the tasks for use-cases where the stream may have more than one subject

### How was this PR tested?


nucliadb/src/nucliadb/tasks/__init__.py
```diff
@@ -17,6 +17,5 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
-from .consumer import NatsTaskConsumer, create_consumer, start_consumer  # noqa
-from .producer import NatsTaskProducer, create_producer, get_producer  # noqa
-from .registry import register_task  # noqa
+from .consumer import NatsTaskConsumer, create_consumer  # noqa
+from .producer import NatsTaskProducer, create_producer  # noqa
```
jotare, line 21: You can use F401 to only skip module import but unused (if you want to be more picky)  

```suggestion
from .consumer import NatsTaskConsumer, create_consumer  # noqa: F401
from .producer import NatsTaskProducer, create_producer  # noqa: F401
```

jotare (approved):

nucliadb/src/nucliadb/tasks/__init__.py
lferran: I'll add it on my next pr! thx

PR #2955, opened 2025-03-10
Able to return user relations in graph strategy

javitonino:
### Description

When using user-relations (without paragraph id), we can still make graph strategy work by faking paragraph ids if using relation as text.


nucliadb/src/nucliadb/search/search/graph_strategy.py
```diff
@@ -369,10 +369,12 @@ async def get_graph_results(
                     kbid=kbid,
                     entities=entities_to_explore,
                     timeout=5.0,
-                    only_with_metadata=True,
+                    only_with_metadata=not graph_strategy.relation_text_as_paragraphs,
                     only_agentic_relations=graph_strategy.agentic_graph_only,
+                    only_entity_to_entity=True,
                     deleted_entities=explored_entities,
                 )
+                # We only want entity to entity relations
```
jotare, line 377: Is this comment intended to be after the call?

jotare (approved):

PR #2964, opened 2025-03-11
Custom collector for unique nodes and relations

jotare:
### Description

### How was this PR tested?


nidx/nidx_relation/src/graph_collector.rs
```diff
@@ -0,0 +1,229 @@
...
+        Ok(unique)
+    }
+}
+
+impl Collector for TopUniqueRelationCollector {
+    type Fruit = Vec<TantivyDocument>;
+    type Child = TopUniqueRelationSegmentCollector;
+
+    fn requires_scoring(&self) -> bool {
+        true
```
javitonino, line 129: Given that you are not using the score at all, I'd suggest setting this to false (same for the other collector). It avoids calculating scores just to be discarded and it's a bit faster.

nidx/nidx_relation/src/graph_collector.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    fn merge_fruits(
+        &self,
+        segment_fruits: Vec<<Self::Child as SegmentCollector>::Fruit>,
+    ) -> tantivy::Result<Self::Fruit> {
+        // 1. Merge de hash maps
+        // 2. Build a vec of documents for unique relations
+
+        let mut unique = HashSet::new();
+        let mut docs = Vec::new();
```
javitonino, line 149: Why is this needed? You could just return the `unique` HashSet.

Also typo in the comments.

nidx/nidx_relation/src/reader.rs
```diff
@@ -92,6 +110,63 @@ impl RelationsReaderService {
...
+    fn nodes_graph_search(
+        &self,
+        query: &dyn Query,
+        top_k: usize,
+        selector: NodeSelector,
+    ) -> anyhow::Result<GraphSearchResponse> {
+        // Node search returns unique nodes using a custom collector
+        let collector = TopUniqueNodeCollector::new(selector, top_k);
+        let searcher = self.reader.searcher();
+        let results = searcher.search(query, &collector)?;
```
javitonino, line 134: I don't think this affair with the nodes search and the NodeSelector is quite what we want.

What we want, as I understand it, is to find all nodes that match a condition. What this is implementing is matching all nodes that are part of a triplet that matches a condition.

I think what we need to do is to run two queries:
- Match on the source node and return source nodes
- Match on the destination node and return destination nodes

And then just merge the results from both queries.

Example query: Find me nodes that match "James". We want to return "James"
With the current system, how would you do that?
- If using SourceNodes/DestinationNodes, you need to know beforhand in which part of the triplet it appears.
- If using Nodes, it will also include any related nodes in the response (e.g: "Ian Fleming")

javitonino (changes requested):

nidx/nidx_relation/src/graph_collector.rs
jotare: Ups, the comments were for myself. Deleting them.
And we need the docs because we want to serialize the relation metadata in the response. The other option would be doing it in the collector and returning the whole `Relation` instead of parts.

nidx/nidx_relation/src/graph_collector.rs
jotare: In this matter, I think to be correct we should consider scores and only return the top K from best to worse, as fuzzy could cause weird behavior.

nidx/nidx_relation/src/reader.rs
jotare: Sure, I started for something more generic and differed a bit from the original. I'll change it

javitonino (approved):

PR #2970, opened 2025-03-12
Search configs

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/charts/nucliadb_reader/templates/reader.vs.yaml
+++ b/charts/nucliadb_reader/templates/reader.vs.yaml
@@ -52,6 +52,14 @@ spec:
             regex: '^/api/v\d+/kb/[^/]+/schema'
           method:
             regex: "GET|OPTIONS"
+        - uri:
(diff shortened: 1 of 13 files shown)


charts/nucliadb_reader/templates/reader.vs.yaml
```diff
@@ -52,6 +52,14 @@ spec:
             regex: '^/api/v\d+/kb/[^/]+/schema'
           method:
             regex: "GET|OPTIONS"
+        - uri:
+            regex: '^/api/v\d+/kb/[^/]+/search_configurations/[^/]+$'
+          method:
+            regex: "GET"
```
lferran, line 58: options is needed too here I think

javitonino: It's being routed to the writer. I could add it here and it would be a conflict, although we don't care where it goes. In both cases, the response to options should be the same.

lferran (approved):
Looks great! Do we want to integrate also resource/find and resource/ask endpoints?

PR #2973, opened 2025-03-13
Send classification labels to processing

lferran:
### Description
This PR makes sure that when:
- Creating a new resource with labels
- Updating a resource
- Adding a new field to a resource that has labels
- Reprocessing a field
- Reprocessing a resource

the classification labels (the ones coming on the request and the already stored ones) are sent to the processing engine request, so that only certain DA are applied depending on label criteria.

### How was this PR tested?
Integration tests


lferran:
sc-11666

PR #2981, opened 2025-03-18
Remove old relation fields

javitonino:
### Description

### How was this PR tested?


nucliadb_protos/writer.proto
```diff
@@ -83,7 +83,6 @@ message BrokerMessage {
 
     resources.Origin origin = 9;
 
-    repeated utils.Relation relations = 10;
```
lferran, line 86: can you add the 10 to the reserved field numbers avobe?

lferran (approved):

nucliadb_protos/writer.proto
javitonino: Done. I also added a migration to remove the maindb keys that are no longer needed.

lferran (approved):

PR #2988, opened 2025-03-19
Graph node encoding and leverage fast field in collector

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nidx/nidx_relation/src/graph_collector.rs
+++ b/nidx/nidx_relation/src/graph_collector.rs
@@ -18,7 +18,7 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 //
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{hash_map::Entry, HashMap, HashSet};
 
 use tantivy::{
     DocId, Score, SegmentOrdinal, SegmentReader, TantivyDocument,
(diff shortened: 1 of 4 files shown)


nidx/nidx_relation/src/graph_collector.rs
```diff
@@ -194,6 +171,108 @@ impl SegmentCollector for TopUniqueNodeSegmentCollector {
...
+    }
+}
+
+impl TopUniqueRelationCollector {
+    pub fn new(schema: crate::schema::Schema, limit: usize) -> Self {
+        Self { limit, schema }
+    }
+}
+
+impl Collector for TopUniqueRelationCollector {
```
javitonino, line 239: Should we do a fast field for relations to improve this collector as well? The collector would be pretty much the same (maybe even exactly the same), we would just have to do a the encoding to u64 (just the relation label, so much simpler).

nidx/nidx_relation/src/graph_collector.rs
```diff
@@ -194,6 +171,108 @@ impl SegmentCollector for TopUniqueNodeSegmentCollector {
...
+        while fruit.is_some() && unique.len() < self.limit {
+            unique.insert(fruit.unwrap());
+            fruit = fruits.next();
+        }
+        Ok(unique)
+    }
+}
+
+impl SegmentCollector for TopUniqueNodeSegmentCollector2 {
+    type Fruit = HashSet<Vec<u64>>;
```
javitonino, line 217: Still not scoring, I guess for next PR (not as critical as this one, since that can be added without reindexing).

javitonino (approved):

javitonino (approved):

PR #3001, opened 2025-03-25
Proper node/relation top unique N with scoring

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nidx/nidx_relation/src/top_unique_n.rs
+++ b/nidx/nidx_relation/src/top_unique_n.rs
@@ -1,0 +1,116 @@
+// Copyright (C) 2021 Bosutech XXI S.L.
+//
+// nucliadb is offered under the AGPL v3.0 and as commercial software.
+// For commercial licensing, contact us at info@nuclia.com.
+//
+// AGPL:
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as
+// published by the Free Software Foundation, either version 3 of the
+// License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


nidx/nidx_relation/src/top_unique_n.rs
```diff
@@ -0,0 +1,116 @@
...
+            .or_insert(score);
+    }
+
+    // Truncate the current set of element to N leaving only the top-scoring
+    // elements. Return the smallest score across the top.
+    fn truncate_top_n(&mut self) -> f32 {
+        let mut vec = Vec::from_iter(self.elements.drain());
+        vec.sort_unstable_by(|a, b| a.1.total_cmp(&b.1).reverse());
+        vec.truncate(self.top_n);
+        let lowest_score = vec.last().expect("truncation must never be done without any element").1;
```
javitonino, line 70: You could just return `-infinity` in the case where it's empty and avoid this panic. This can happen if `top_k=0` which we use in some cases withouth graph (e.g: we only want faceted results in catalog). This is safe now , but I can see a future in which we add graph searching to something and suddenly panic because of this.

nidx/nidx_relation/src/top_unique_n.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        self.elements.extend(vec.into_iter());
+
+        lowest_score
+    }
+
+    pub fn into_sorted_vec(mut self) -> Vec<(K, f32)> {
+        if self.elements.len() > self.top_n {
+            self.truncate_top_n();
+        }
```
javitonino, line 80: I think this whole if is not needed, you are already ordering and cutting below.

javitonino (approved):

nidx/nidx_relation/src/top_unique_n.rs
jotare: Yes, changed

PR #3004, opened 2025-03-26
Unleash node query power

jotare:
### Description

### How was this PR tested?


nidx/nidx_relation/src/graph_query_parser.rs
```diff
@@ -482,42 +485,52 @@ impl<'a> GraphQueryParser<'a> {
...
+                    Box::new(FuzzyTermQuery::new_prefix(exact_term, fuzzy.fuzzy_distance, true))
+                } else {
+                    Box::new(FuzzyTermQuery::new(exact_term, fuzzy.fuzzy_distance, true))
+                }
+            }
+
+            Term::FuzzyWord(fuzzy) =>
+            {
+                #[allow(clippy::collapsible_else_if)]
+                if fuzzy.is_prefix {
```
javitonino, line 504: You can remove a bunch of duplication here by extracting the function to a variable. Something like:

```
let query_builder = if fuzzy { FuzzyTermQuery::new_prefix } else { FuzzyTermQuery::new };
```

and then use it in a single codepath like:
```
  if tokenized_terms.len() == 1 {
      let tokenized_term = tokenized_terms.into_iter().next().unwrap();
      Box::new(query_builder(tokenized_term, fuzzy.fuzzy_distance, true))
  } else {
```

javitonino (approved):
Just a suggestion

nidx/nidx_relation/src/graph_query_parser.rs
jotare: Done :)

PR #3033, opened 2025-04-04
Proper per-field deletions (1: python part)

lferran:
### Description

### How was this PR tested?


nidx/nidx_text/src/lib.rs
```diff
@@ -99,7 +126,7 @@ impl TextIndexer {
     }
 
     pub fn deletions_for_resource(&self, resource: &nidx_protos::Resource) -> Vec<String> {
-        vec![resource.resource.as_ref().unwrap().uuid.clone()]
+        resource.texts_to_delete.clone()
```
javitonino, line 129: We cannot change this until after the nucliadb part is deployed that always fills in this field. Otherwise, we'll end up with duplicated fields.

So we may need to do it in two deploys.

nidx/src/import_export.rs
```diff
@@ -346,12 +340,6 @@ mod tests {
         let shard_dest = Shard::get(&meta_dest.pool, shard.id).await?;
         assert_eq!(shard_dest.kbid, shard.kbid);
         assert_eq!(Index::for_shard(&meta_dest.pool, shard.id).await?.len(), 3);
-        assert_eq!(
-            sqlx::query_scalar!("SELECT COUNT(*) AS \"cnt!\" FROM deletions")
-                .fetch_one(&meta_dest.pool)
```
javitonino, line 351: I want to keep this. This ensures deletions are exported+imported. We can just add some random deletion to the resource being indexed at the beginning of the test (maybe just set tests_to_delete to some random key).

javitonino (approved):

PR #3038, opened 2025-04-07
Bounded train resource cache to avoid memory issues

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/cache.py
+++ b/nucliadb/src/nucliadb/common/cache.py
@@ -21,6 +21,7 @@
 import asyncio
 from abc import ABC, abstractmethod
 from contextvars import ContextVar
+from dataclasses import dataclass
 from functools import cached_property
 from typing import Generic, Optional, TypeVar
 
@@ -28,14 +29,22 @@ from lru import LRU
 
 from nucliadb.ingest.orm.resource import Resource as ResourceORM
 from nucliadb_protos.utils_pb2 import ExtractedText
-from nucliadb_telemetry.metrics import Gauge
+from nucliadb_telemetry.metrics import Counter, Gauge
 
 # specific metrics per cache type
 cached_resources = Gauge("nucliadb_cached_resources")
 cached_extracted_texts = Gauge("nucliadb_cached_extracted_texts")
+resource_cache_ops = Counter("nucliadb_resource_cache_ops", labels={"type": ""})
+extracted_text_cache_ops = Counter("nucliadb_extracted_text_cache_ops", labels={"type": ""})
 
 
 T = TypeVar("T")
+
+
+@dataclass
+class CacheMetrics:
+    _cache_size: Gauge
+    ops: Counter
 
 
 class Cache(Generic[T], ABC):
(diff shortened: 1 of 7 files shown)


nucliadb/src/nucliadb/train/generators/utils.py
```diff
@@ -29,29 +31,44 @@
 from nucliadb.train.types import T
 from nucliadb_utils.utilities import get_storage
 
-rcache: ContextVar[Optional[dict[str, ResourceORM]]] = ContextVar("rcache", default=None)
+
+class ResourceCache:
+    def __init__(self, size: int = 1024):
```
lferran, line 36: I would set a lower size for the cache.

Bear in mind that in the resource object we're accumulating the extracted texts and field metadata behaviours, which can be 100s of MBs in some cases (large pdfs, for instance)

So I would rather get some cache misses than dying of OOM

jotare: What do you think is a good value then? Right now we have an unbounded cache and 600Mi, so I don't think is fair to assume all resources are 100s of MBs

nucliadb_telemetry/src/nucliadb_telemetry/metrics.py
lferran, line 1: can you extend the existing unit tests for metrics?

nucliadb/src/nucliadb/common/cache.py
```diff
@@ -0,0 +1,168 @@
...
+# Global caches (per asyncio task)
+
+rcache: ContextVar[Optional[ResourceCache]] = ContextVar("rcache", default=None)
+etcache: ContextVar[Optional[ExtractedTextCache]] = ContextVar("etcache", default=None)
+
+
+# Cache management
+
+
+# Get or create a resource cache specific to the current asyncio task
```
lferran, line 129: I would add to the comment something like: "... and all its subtasks. Make sure to set the context var in the parent task of the concurrent tasks you want to spawn."

nucliadb/src/nucliadb/search/search/cache.py
```diff
@@ -37,47 +43,13 @@
 
 logger = logging.getLogger(__name__)
 
-rcache: ContextVar[Optional[dict[str, ResourceORM]]] = ContextVar("rcache", default=None)
-etcache: ContextVar[Optional["ExtractedTextCache"]] = ContextVar("etcache", default=None)
+etcache: ContextVar[Optional[ExtractedTextCache]] = ContextVar("etcache", default=None)
```
lferran, line 46: this one can be removed no? it's already in common.cache

nucliadb/src/nucliadb/common/cache.py
```diff
@@ -0,0 +1,171 @@
...
+    def __del__(self):
+        # we want to clear the cache before deleting the object and set the
+        # metric appropriately
+        self.clear()
+
+    @abstractmethod
+    @cached_property
+    def _cache_size_metric(self) -> Gauge: ...
+
+    def _evicted_callback(self, key: str, value: T):
```
lferran, line 87: more than the gauges, I think i'd be more interesting to know when there are evictions.
Can you add a counter metric here?

lferran (approved):

PR #3051, opened 2025-04-10
Support graph search in search request

jotare:
### Description
Replace entities subgraph search for new graph search

### How was this PR tested?


nucliadb_protos/nodereader.proto
```diff
@@ -292,6 +292,8 @@ message GraphSearchResponse {
         uint32 relation = 2;
         uint32 destination = 3;
         optional utils.RelationMetadata metadata = 4;
+        optional string resource_field_id = 5;
+        repeated string facets = 6;
```
javitonino, line 296: Why are you adding facets here? Do you think you'll need them? I think we should not expose them (at least directly) to the user, this is an implementation detail on how we implement some of the graph filters, the user does not have to know about this.

javitonino (approved):

nucliadb_protos/nodereader.proto
jotare: I'm not exposing those to users, as serialization is done manually and not with the protobuf to dict. 
And I'm adding it because this response will replace EntitiesSubgraphResponse, that return IndexRelation that contain those fields

jotare:
[sc-11743]

PR #3075, opened 2025-04-23
Use python build system to compile protos

javitonino:
### Description
Instead of manually compiling protos from the Makefile, use the python build system.

### How was this PR tested?
Compiled package before and after. Confirmed they were identical.


lferran (approved):
seems like there is a semgrep complain, but apart from that lgtm!

PR #3087, opened 2025-04-25
Integrate graph search into /find endpoint

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/find_merge.py
+++ b/nucliadb/src/nucliadb/search/search/find_merge.py
@@ -68,6 +68,9 @@ FIND_FETCH_OPS_DISTRIBUTION = metrics.Histogram(
     buckets=[1, 5, 10, 20, 30, 40, 50, 60, 80, 100, 200],
 )
 
+# Constant score given to all graph results until we implement graph scoring
+FAKE_GRAPH_SCORE = 1.0
+
 
 @merge_observer.wrap({"type": "find_merge"})
 async def build_find_response(
@@ -103,14 +106,17 @@ async def build_find_response(
             search_response.vector.documents,
         )
     )
+    graph_results = graph_results_to_text_block_matches(search_response.graph)
 
     merged_text_blocks: list[TextBlockMatch]
-    if len(keyword_results) == 0:
-        merged_text_blocks = semantic_results
-    elif len(semantic_results) == 0:
+    if len(keyword_results) > 0 and len(semantic_results) == 0 and len(graph_results) == 0:
         merged_text_blocks = keyword_results
+    elif len(keyword_results) == 0 and len(semantic_results) > 0 and len(graph_results) == 0:
+        merged_text_blocks = semantic_results
+    elif len(keyword_results) == 0 and len(semantic_results) == 0 and len(graph_results) > 0:
+        merged_text_blocks = graph_results
     else:
-        merged_text_blocks = rank_fusion_algorithm.fuse(keyword_results, semantic_results)
+        merged_text_blocks = rank_fusion_algorithm.fuse(keyword_results, semantic_results, graph_results)
 
     # cut
     # we assume pagination + predict reranker is forbidden and has been already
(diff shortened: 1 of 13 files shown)


jotare:
[sc-11743]

nucliadb/src/nucliadb/search/search/rank_fusion.py
```diff
@@ -141,21 +152,27 @@ def __init__(
...
         self._semantic_boost = semantic_weight
+        self._graph_boost = graph_weight
 
     @rank_fusion_observer.wrap({"type": "reciprocal_rank_fusion"})
     def _fuse(
-        self, keyword: Iterable[TextBlockMatch], semantic: Iterable[TextBlockMatch]
+        self,
+        keyword: Iterable[TextBlockMatch],
+        semantic: Iterable[TextBlockMatch],
+        graph: Iterable[TextBlockMatch],
```
javitonino, line 162: At this point, maybe this should take an array of things to merge? Something like `list[tuple[Iterable[TextBlockMatch], float]]`? Just make this algorithm generic and when calling it, pass what's needed (no need to pass an empty list)

nucliadb/src/nucliadb/search/search/find_merge.py
```diff
@@ -103,14 +106,17 @@ async def build_find_response(
             search_response.vector.documents,
         )
     )
+    graph_results = graph_results_to_text_block_matches(search_response.graph)
 
     merged_text_blocks: list[TextBlockMatch]
-    if len(keyword_results) == 0:
-        merged_text_blocks = semantic_results
-    elif len(semantic_results) == 0:
+    if len(keyword_results) > 0 and len(semantic_results) == 0 and len(graph_results) == 0:
```
javitonino, line 112: What about something like?
```
features_with_results = [x for x in (a,b,c) if len(keyword_results, semantic_results, graph_results) > 0]
if len(features_with_results) == 1:
  merged_text_blocks = features_with_results[0]
else:
  merged_text_blocks = rank_fusion_algorithm.fuse(keyword_results, semantic_results, graph_results)
```

nucliadb/src/nucliadb/search/search/rank_fusion.py
jotare: For the difference it makes, I'd leave it like that. Otherwise, you'll need to know the boosting at call time, and it'd be a specific parameter for RRF and, even if it's the only one we have, I don't think it's worth it

nucliadb/src/nucliadb/search/search/rank_fusion.py
```diff
@@ -62,21 +62,28 @@ def window(self) -> int:
...
     ) -> list[TextBlockMatch]: ...
 
 
+# DEPRECATED: only used in tests!
```
javitonino, line 86: Consider moving it to the test directory then? To some helper in `tests/search/unit` or something?

We could also rewrite the tests to use RRF, but not sure if that makes sense. In any case, moving this out should be easy and it's less code to see here.

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1724,6 +1727,11 @@ class FindRequest(BaseSearchRequest):
     query_entities: SkipJsonSchema[Optional[list[KnowledgeGraphEntity]]] = Field(
         default=None, title="Query entities", description="Entities to use in a knowledge graph search"
     )
+    graph_query: SkipJsonSchema[Optional[GraphPathQuery]] = Field(
```
javitonino, line 1730: You are not propagating this to the GET endpoints. On purpose? It seems like we are likely to forget to add it later unless we make an explicit note to remember.

nucliadb_models/src/nucliadb_models/graph/requests.py
```diff
@@ -161,8 +160,16 @@ class GraphSearchRequest(BaseGraphSearchRequest):
             "Filtering examples can be found here: https://docs.nuclia.dev/docs/rag/advanced/search-filters "
         ),
     )
-    security: Optional[RequestSecurity] = SearchParamDefaults.security.to_pydantic_field()
-    show_hidden: bool = SearchParamDefaults.show_hidden.to_pydantic_field()
+    security: Optional[RequestSecurity] = Field(
```
javitonino, line 163: Why this change?

jotare: A problematic circular dependency between search and graph models. We could move SearchParamDefaults outside but still, feels kind of weird having a giant class with defaults for all endpoints

nucliadb/src/nucliadb/search/search/find_merge.py
jotare: Done and moved to rank fusion

nucliadb_models/src/nucliadb_models/search.py
javitonino: Yeah, it's weird, but it's what I ended up doing for filter expressions. :shrug: 

Maybe we should deprecate GETs

PR #3101, opened 2025-05-05
Rename /ask context to chat_history

jotare:
### Description

### How was this PR tested?


lferran (approved):
lgtm! since this is a param that is used a lot, maybe you could leave at least a test that checks that the former param still works?

PR #3113, opened 2025-05-12
automatic context switching

drf7:
### Description
Option on /ask to ignore the chat history if the last question is not related.
(it would impact both the "rephrasing" for the retrieval and the final context passed to the LLM)

### How was this PR tested?
Unit tests


nucliadb/src/nucliadb/search/predict.py
```diff
@@ -139,6 +140,12 @@ def prettify(self) -> str:
         }[self]
 
 
+@dataclass
+class RephraseResponse:
+    rephrased_query: str
+    use_chat_history: Optional[bool]
```
lferran, line 146: why is the Optional needed? should we simply default to False?

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -996,13 +996,48 @@ class ChatModel(BaseModel):
 class RephraseModel(BaseModel):
     question: str
     chat_history: list[ChatContextMessage] = []
+    context: list[Message] = Field(default_factory=list)
```
lferran, line 999: where is this used?

drf7: I saw the RephraseModel was out of sync with the Learning RephraseModel and I updated it: https://github.com/nuclia/learning/blob/ac726cf51ccedab669131d5a200686ffc2820b47/libraries/learning_generative/src/learning_generative/models.py#L156

nucliadb/src/nucliadb/search/predict.py
drf7: It is mapping a header coming from learning, this header is not always set, it is only set when `chat_history_relevance_threshold` is sent in the schema.
So here `use_chat_history ` can be None if the header is not set or a bool with the value of the header.

We could return False instead of None when the header is not set, but I don't think there's much of a difference.

nucliadb/src/nucliadb/search/predict.py
```diff
@@ -624,20 +631,23 @@ async def _parse_generative_chunks(gen):
...
         raise RephraseError(content[:-2])
     elif content.endswith("-2"):
         raise RephraseMissingContextError(content[:-2])
-    else:
-        # bw compatibility
-        return content
+
+    use_chat_history = None
+    if NUCLIA_LEARNING_CHAT_HISTORY_HEADER in resp.headers:
+        use_chat_history = resp.headers[NUCLIA_LEARNING_CHAT_HISTORY_HEADER] == "true"
```
lferran, line 652:
```suggestion
        use_chat_history = resp.headers[NUCLIA_LEARNING_CHAT_HISTORY_HEADER].lower() == "true"
```
to be safer

nucliadb/src/nucliadb/search/search/chat/ask.py
```diff
@@ -491,14 +491,19 @@ async def ask(
...
                     chat_history=chat_history,
                     query=user_query,
                     user_id=user_id,
                     user_context=user_context,
                     generative_model=ask_request.generative_model,
+                    chat_history_relevance_threshold=ask_request.chat_history_relevance_threshold,
                 )
+                rephrased_query = response.rephrased_query
+                if response.use_chat_history is False:
+                    # Ignored if the question is not relevant enough with the chat history
```
lferran, line 505: Maybe a logging info/debug is useful here, in case we need to debug it.

lferran (approved):
some nit comments

PR #3114, opened 2025-05-12
More generic metrics information

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/query.py
+++ b/nucliadb/src/nucliadb/search/search/chat/query.py
@@ -462,23 +462,22 @@ async def run_prequeries(
     results: list[PreQueryResult] = []
     max_parallel_prequeries = asyncio.Semaphore(settings.prequeries_max_parallel)
 
-    async def _prequery_find(
-        prequery: PreQuery,
-    ):
+    async def _prequery_find(prequery: PreQuery, index: int):
         async with max_parallel_prequeries:
+            prequery_id = prequery.id or str(index)
             find_results, _, _ = await find(
                 kbid,
                 prequery.request,
                 x_ndb_client,
                 x_nucliadb_user,
                 x_forwarded_for,
-                metrics=metrics,
+                metrics=metrics.child_span(prequery_id),
             )
             return prequery, find_results
 
     ops = []
-    for prequery in prequeries:
-        ops.append(asyncio.create_task(_prequery_find(prequery)))
+    for idx, prequery in enumerate(prequeries):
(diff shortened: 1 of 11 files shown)


nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -2077,7 +2082,7 @@ class SyncAskResponse(BaseModel):
     debug: Optional[dict[str, Any]] = Field(
         default=None,
         title="Debug information",
-        description="Debug information about predict",
+        description="Debug information about the ask operation. This is only available if the `debug` parameter is set to true in the request.",  # noqa: E501
```
javitonino, line 2085: Maybe we should add some notice here and in the `metrics` field that this information is subject to change at any point? I don't want people to start relying it and complain if we change the format of it (I don't want to do bw/c on these fields).

nucliadb/tests/nucliadb/integration/test_ask.py
```diff
@@ -224,9 +224,12 @@ async def test_sync_ask_returns_debug_mode(
         assert resp_data.prompt_context
         assert resp_data.predict_request
         assert isinstance(resp_data.predict_request, dict)
+        assert isinstance(resp_data.debug, dict)
+        assert resp_data.debug["metrics"]
```
javitonino, line 228: Maybe check that it actually contains some values as well? e.g: Check that `index_search` is > 0 or something like that. This way we make sure we are actually passing some data and not just an empty dict.

nucliadb/src/nucliadb/search/search/chat/query.py
```diff
@@ -455,32 +454,30 @@ async def run_prequeries(
...
     """
     results: list[PreQueryResult] = []
     max_parallel_prequeries = asyncio.Semaphore(settings.prequeries_max_parallel)
 
-    async def _prequery_find(
-        prequery: PreQuery,
-    ):
+    async def _prequery_find(prequery: PreQuery, index: int):
         async with max_parallel_prequeries:
+            prequery_id = prequery.id or str(index)
```
javitonino, line 467: Maybe `str(index)` should be `f"prequery-{index}"` or something like that, a bit more clear

javitonino (approved):

nucliadb_models/src/nucliadb_models/search.py
lferran: agree!

lferran:
[sc-12292]

PR #3123, opened 2025-05-16
add audit to predict chat proxy

drf7:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/predict_proxy.py
+++ b/nucliadb/src/nucliadb/search/search/predict_proxy.py
@@ -17,15 +17,34 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+import json
 from enum import Enum
 from typing import Any, Optional, Union
 
+import aiohttp
 from fastapi.datastructures import QueryParams
 from fastapi.responses import Response, StreamingResponse
+from multidict import CIMultiDictProxy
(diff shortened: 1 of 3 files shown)


nucliadb/src/nucliadb/search/search/predict_proxy.py
```diff
@@ -97,3 +160,94 @@ async def predict_proxy(
...
+
+            if is_json:
+                try:
+                    parsed_chunk = GenerativeChunk.model_validate(chunk)
+                    if isinstance(parsed_chunk, TextGenerativeResponse):
+                        text_answer += parsed_chunk.text
+                    elif isinstance(parsed_chunk, JSONGenerativeResponse):
+                        json_object = parsed_chunk.object
+                    elif isinstance(parsed_chunk, StatusGenerativeResponse):
+                        status_code = parsed_chunk.code
```
lferran, line 196: can you add an else clause here with a warning? Just in case we're missing something out in the future

lferran: I think you are missing the metadata one actually

drf7: Yeah, I'm missing many things, the reason is because I only want to process the chunks of data that contain relevant data for the audit logs, the rest is just proxied without processing it, to keep the endpoint behaviour without changes

lferran (approved):

PR #3128, opened 2025-05-19
Fix indexing logic to avoid duplicating text fields

lferran:
### Description

### How was this PR tested?


lferran:
[sc-11816]

nucliadb/src/nucliadb/ingest/orm/brain_v2.py
```diff
@@ -216,14 +216,15 @@ def generate_paragraphs(
         user_field_metadata: Optional[UserFieldMetadata],
         replace_field: bool,
         skip_index: Optional[bool],
+        skip_texts_index: Optional[bool],
```
jotare, line 219: Could you rename the one above, so we know which index are we skipping with each param?

jotare (approved):

PR #3146, opened 2025-05-27
Filter expression in train field streaming

lferran:
### Description
- Moved filter expresion parsing and InvalidQueryError  to nucliadb.common to be able to reuse in search and train
- Added support for filter expressions on nidx Document stream
- Adapted train field streaming endpoint to allow filter expression on the json payload

### How was this PR tested?
Integration tests


nucliadb/src/nucliadb/train/generators/field_classifier.py
```diff
@@ -38,6 +39,7 @@ def field_classification_batch_generator(
     kbid: str,
     trainset: TrainSet,
     shard_replica_id: str,
+    filter_expression: Optional[FilterExpression],
```
javitonino, line 42: Maybe thrown an error in all the trainsets where this is not used if it's not none? Something like `filter_expression` not supported for field classifier or whatever. Just to avoid surprises if someone tries to use this.

javitonino (approved):

PR #3148, opened 2025-05-28
Catalog title filtering

javitonino:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/pgcatalog.py
+++ b/nucliadb/src/nucliadb/search/search/pgcatalog.py
@@ -135,8 +135,10 @@ def _prepare_query_search(query: search_models.CatalogQuery, params: dict[str, A
     elif query.match == search_models.CatalogQueryMatch.StartsWith:
         params["query"] = query.query + "%"
         if query.field == search_models.CatalogQueryField.Title:
+            # Insensitive search supported by pg_trgm for title
             return f"{query.field.value} ILIKE %(query)s"
         else:
+            # Sensitive search for slug (btree does not support ILIKE and slugs are all lowercase anyway)
             return f"{query.field.value} LIKE %(query)s"
     # The rest of operators only supported by title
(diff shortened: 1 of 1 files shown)


nucliadb/src/nucliadb/search/search/query_parser/models.py
```diff
@@ -172,7 +172,7 @@ class Date:
 
 class CatalogQuery(BaseModel):
     kbid: str
-    query: str
+    query: Optional[search_models.CatalogQuery]
```
lferran, line 175: Query can be None when you simply want to paginate all, right? Just like before it was query = ""?

javitonino: Yes, the query planner converts this. I just feel None is clearer than an empty string.

nucliadb/src/nucliadb/search/search/pgcatalog.py
```diff
@@ -133,6 +128,39 @@ def _prepare_query_filters(catalog_query: CatalogQuery) -> tuple[str, dict[str,
...
+def _prepare_query_search(query: search_models.CatalogQuery, params: dict[str, Any]) -> str:
+    if query.match == search_models.CatalogQueryMatch.Exact:
+        params["query"] = query.query
+        return f"{query.field.value} = %(query)s"
+    elif query.match == search_models.CatalogQueryMatch.StartsWith:
+        params["query"] = query.query + "%"
+        if query.field == search_models.CatalogQueryField.Title:
+            return f"{query.field.value} ILIKE %(query)s"
+        else:
+            return f"{query.field.value} LIKE %(query)s"
```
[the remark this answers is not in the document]
javitonino: It's not because of that. It's just that the index I plan to use for the slug (next PR) won't support insensitive search (and it doesn't matter because slugs are all lowercase anyway).

PR #3159, opened 2025-06-03
Fix citations for neighbouring paragraphs

lferran:
### Description

By moving the neighbouring paragraphs logic to be a text block hydration option (not published on the api yet), we solve the problem where the context sent to the llm was different from the retrieval results shown to the user, which was making the citations not work.

### How was this PR tested?
Integration tests

### DIFF
--- a/nucliadb/src/nucliadb/ingest/fields/base.py
+++ b/nucliadb/src/nucliadb/ingest/fields/base.py
@@ -19,6 +19,7 @@
 #
 from __future__ import annotations
 
+import asyncio
 import enum
 import logging
 from datetime import datetime
@@ -99,7 +100,7 @@ class Field(Generic[PbType]):
         self.computed_metadata = None
         self.large_computed_metadata = None
         self.question_answers = None
-
+        self.locks: dict[str, asyncio.Lock] = {}
         self.id: str = id
         self.resource = resource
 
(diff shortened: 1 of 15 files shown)


nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -499,10 +500,13 @@ async def set_field_metadata(self, payload: FieldComputedMetadataWrapper) -> Fie
...
-            payload = await self.storage.download_pb(sf, FieldComputedMetadata)
-            if payload is not None:
-                self.computed_metadata = payload
+            async with self.locks.setdefault("field_metadata", asyncio.Lock()):
+                if self.computed_metadata is None:
+                    # Value can be fetch while waiting for the lock
+                    sf = self.get_storage_field(FieldTypes.FIELD_METADATA)
+                    payload = await self.storage.download_pb(sf, FieldComputedMetadata)
+                    if payload is not None:
+                        self.computed_metadata = payload
```
jotare, line 509: Not saying that we should fix it here, but the possibility of a value being None makes impossible to differentiate between None and not cached

lferran: it's ok, I guess it's a corner case and if it's none there's no big deal on fetching it multiple times.
But yeah, I agree it could be better

PR #3171, opened 2025-06-06
add x-show-consumption header

drf7:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/search/chat/ask.py
+++ b/nucliadb/src/nucliadb/search/search/chat/ask.py
@@ -84,6 +84,7 @@ from nucliadb_models.search import (
     ChatModel,
     ChatOptions,
     CitationsAskResponseItem,
+    ConsumptionResponseItem,
     DebugAskResponseItem,
     ErrorAskResponseItem,
     FindOptions,
@@ -107,6 +108,7 @@ from nucliadb_models.search import (
     StatusAskResponseItem,
     SyncAskMetadata,
     SyncAskResponse,
+    TokensDetail,
     UserPrompt,
     parse_custom_prompt,
     parse_rephrase_prompt,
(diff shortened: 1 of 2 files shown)


nucliadb_sdk/pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "httpx",
     "orjson",
     "pydantic>=2.6",
-    "nuclia-models>=0.24.2",
+    "nuclia-models>=0.43.0",
```
javitonino, line 29: This is not needed, is it? (nucliadb_sdk doesn't use the new things)

Just to be clear, no need to revert it.

nucliadb/tests/search/integration/api/v1/test_ask_security.py
```diff
@@ -42,7 +44,7 @@ async def test_ask_receives_injected_security_groups(
     )
     assert resp.status_code == 200
     spy.assert_called_once()
-    ask_request = spy.call_args[0][1]
+    ask_request = spy.call_args[1]["ask_request"]
```
javitonino, line 47: Why is this test changing?

nucliadb_sdk/pyproject.toml
drf7: Yeah, right, not really needed there

nucliadb/tests/search/integration/api/v1/test_ask_security.py
drf7: Because I changed how we called the `create_ask_response` passing kwargs instead of args. That change was not needed either but I think it is less error prone

https://github.com/nuclia/nucliadb/pull/3171/files#diff-ca0cdfdb1723a49271a9869120dfa20b781baf81714ec0a5283500a4d2d87d7bL100-R107

PR #3187, opened 2025-06-13
tweak combined rag strategies so paragraph is not overwritten

lferran:
### Description
Otherwise, the combo of hiearchy + neighbouring was overwritting the hierarchy paragraph

### How was this PR tested?


nucliadb/src/nucliadb/search/search/chat/prompt.py
```diff
@@ -668,7 +683,7 @@ def _get_paragraph_text(extracted_text: ExtractedText, pid: ParagraphId) -> str:
         if field_extracted_text is None:
             continue
         ptext = _get_paragraph_text(field_extracted_text, pid)
-        if ptext:
+        if ptext and pid.full() not in context:
```
lferran, line 686: here is the fix

PR #3192, opened 2025-06-17
Serialize default fields on SDK

carlesonielfa:
### Description
We were not serializing fields that contained default values. This caused a problem for features like RAG strategies, which use the `name` field as a discriminator (with  a default value set) to determine the specific strategy type. Because the name was not being sent, these strategies could not be used.

### How was this PR tested?
Added additional test


carlesonielfa:
Closing this as not serializing default fields was done on purpose for backwards compatibility

PR #3202, opened 2025-06-18
nucliadb_models typechecking

javitonino:
Make nucliadb_models pass typechecking with pyright & ty, fixing all typechecking warnings. This uncovered some actual bugs that could trigger when using the SDK.

### DIFF
--- a/nucliadb_models/tests/test_search.py
+++ b/nucliadb_models/tests/test_search.py
@@ -30,40 +30,40 @@ def test_filter_model_validator():
 def test_field_extension_strategy_fields_field_validator():
     search.FieldExtensionStrategy(
         name="field_extension",
-        fields={"f/myfield"},
+        fields=["f/myfield"],
     )
 
     # not a set of fields
     with pytest.raises(ValidationError):
         search.FieldExtensionStrategy(
             name="field_extension",
-            fields=0,
+            fields=0,  # type: ignore
         )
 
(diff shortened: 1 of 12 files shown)


nucliadb_models/tests/test_search.py
```diff
@@ -30,40 +30,40 @@ def test_filter_model_validator():
...
+        search.FindRequest(features=[search.SearchOptions.FULLTEXT])  # type: ignore
 
 
 def test_find_supports_search_options():
-    search.FindRequest(features=[search.SearchOptions.KEYWORD])
-    search.FindRequest(features=[search.SearchOptions.SEMANTIC])
-    search.FindRequest(features=[search.SearchOptions.RELATIONS])
+    search.FindRequest(features=[search.FindOptions.KEYWORD])
+    search.FindRequest(features=[search.FindOptions.SEMANTIC])
+    search.FindRequest(features=[search.FindOptions.RELATIONS])
```
jotare: This isn't a mistake. It is testing that you can actually pass `SearchOptions` to `FindRequest` as this was a legacy behavior we should support (at least for a while to not break users)

javitonino: Ah, I saw another test above testing this and got confused. Fixed it (adding a legitimate type: ignore comment)

jotare (approved):

PR #3215, opened 2025-07-10
Extract DataStore code

javitonino:
### Description

This PR extracts all logic related storing the vectors and associated metadata (id, labels, etc.) to a separate module. The purpose of this is to make it easier to change the data store format, not having to track the logic all over the code. We are doing this to be able to implement a new data store with better support for multi-vector representations (like colbert). The plan is to do this by converting `DataStore` and `Node` to traits and having two implementation of them.

Some significant changes:
- `Node` used to be a repository of functions operating on bytes. Now it wraps the bytes internally, which offers a safer/easier-to-use interface.
- We used to pass `nodes` (an `Mmap`) around. Now we pass `DataStore` which wraps it and exposes the needed functions.
- `Neighbour` used to copy data (passing `Vec<u8>`). Now it works on slices to avoid that copy. This should make search a tiny bit faster.

Review should be easier by doing it commit by commit. In fact, I think any individual commit can be merged without the others.

### DIFF
--- a/nidx/nidx_vector/src/data_store/store.rs
+++ b/nidx/nidx_vector/src/data_store/store.rs
@@ -18,11 +18,12 @@
 // along with this program. If not, see <http://www.gnu.org/licenses/>.
 //
 
-use std::fs::File;
-use std::io::{self, BufWriter, Seek, SeekFrom, Write};
-
+use super::node::Node;
 use crate::config::{VectorConfig, VectorType};
 use crate::data_types::usize_utils::*;
+use lazy_static::lazy_static;
+use std::fs::File;
+use std::io::{self, BufWriter, Seek, SeekFrom, Write};
 
 // A data store schema.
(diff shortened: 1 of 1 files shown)


nidx/nidx_vector/src/data_store/store.rs
```diff
@@ -71,13 +69,14 @@ pub fn stored_elements(x: &[u8]) -> usize {
...
+pub fn get_value(src: &[u8], id: usize) -> Node {
     let pointer = get_pointer(src, id);
-    Node.read_exact(&src[pointer..]).0
+    Node::new(&src[pointer..])
 }
 
 use lazy_static::lazy_static;
-use libc;
+
+use super::node::Node;
```
jotare: NIT: Can we move this `use` to the top?

jotare (approved):

PR #3217, opened 2025-07-15
Custom keyword query parser

jotare:
### Description
Replace the usage of tantivy's query parser for keyword search and start using our own. This will remove the leak of features from tantivy to our keyword search and will let us control our query grammar.

We use nom v7 as it's already a transitive dependency

Some of the fixes you'll find:
- No more quote everything if parsing fails.
  The query `(theres anything wrong` was converted to `"(theres anything wrong"` due to unbalanced parens. Not anymore

- Better fuzzy fallback with exact matches.
  We use to remove quoted content from the fallback query, thus `this "is important"` would end up being `this` with fuzzy and ` ` (yes, empty query) after stop words removal. We now keep the important content without fuzzy. This may end up with really similar queries, between keyword and fuzzy. Although this could be detected and improved, is better than search without those terms.

- Fuzzy only for longer words.
  We were applying fuzzy to all words with any length. We now reduce this to a minimum distance to avoid bad results (distance 1 of 1-3 words can bring a lot of nonsense)

- More exhaustive and clear tests

- A query like `-came` now will match `AllQuery + MustNot(came)` while before was matching `"-came"`. This a more consistent behavior with our grammar.

### DIFF
--- a/nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
+++ b/nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
@@ -116,3 +116,83 @@ pub fn parse_fuzzy_query(query: &[Token], term_collector: SharedTermC, last_lite
         Box::new(BooleanQuery::new(subqueries))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::search_query::TermCollector;
+
+    use super::*;
+
+    #[test]
+    fn test_short_literals_do_not_fuzzy() {
+        let term_collector = SharedTermC::from(TermCollector::new());
+
+        // literal shorter than MIN_FUZZY_LEN will become a TermQuery
+        let literal = "ab";
+        assert!(literal.len() < MIN_FUZZY_LEN);
+        let query = [Token::Literal(literal.into())];
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), false);
+        assert!(fuzzy.is::<TermQuery>());
+        let q = fuzzy.downcast::<TermQuery>().unwrap();
+        assert_eq!(q.term().value().as_str(), Some(literal));
+    }
+
+    #[test]
+    fn test_fuzzy_literals() {
+        let term_collector = SharedTermC::from(TermCollector::new());
+
+        let literal = "abcd";
+        assert!(literal.len() >= MIN_FUZZY_LEN);
+        let query = [Token::Literal(literal.into())];
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), false);
+        assert!(fuzzy.is::<FuzzyTermQuery>());
+        assert!(!fuzzy.downcast::<FuzzyTermQuery>().unwrap().is_prefix());
+    }
+
+    #[test]
+    fn test_fuzzy_prefix() {
+        let term_collector = SharedTermC::from(TermCollector::new());
+
+        // literals longer than the min fuzzy prefix become prefix if they are
+        // last and the flag is enabled
+
+        let literal = "abcd";
+        assert!(literal.len() >= MIN_FUZZY_PREFIX_LEN);
+        let query = [Token::Literal(literal.into())];
+
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), false);
+        assert!(fuzzy.is::<FuzzyTermQuery>());
+        assert!(!fuzzy.downcast::<FuzzyTermQuery>().unwrap().is_prefix());
+
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), true);
+        assert!(fuzzy.is::<FuzzyTermQuery>());
+        assert!(fuzzy.downcast::<FuzzyTermQuery>().unwrap().is_prefix());
+
+        // only the last term is fuzzy prefix
+
+        let query = [Token::Literal(literal.into()), Token::Literal(literal.into())];
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), true);
+        assert!(fuzzy.is::<BooleanQuery>());
+        let q = fuzzy.downcast::<BooleanQuery>().unwrap();
+        let clauses = q.clauses();
+        assert_eq!(clauses.len(), 2);
+        assert!(clauses[0].1.is::<FuzzyTermQuery>());
+        assert!(!clauses[0].1.downcast_ref::<FuzzyTermQuery>().unwrap().is_prefix());
+        assert!(clauses[1].1.is::<FuzzyTermQuery>());
+        assert!(clauses[1].1.downcast_ref::<FuzzyTermQuery>().unwrap().is_prefix());
+
+        // however, shorter terms won't become prefix
+
+        let literal = "abc";
+        assert!(literal.len() < MIN_FUZZY_PREFIX_LEN);
+        let query = [Token::Literal(literal.into())];
+
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), false);
+        assert!(fuzzy.is::<FuzzyTermQuery>());
+        assert!(!fuzzy.downcast::<FuzzyTermQuery>().unwrap().is_prefix());
+
+        let fuzzy = parse_fuzzy_query(&query, term_collector.clone(), true);
+        assert!(fuzzy.is::<FuzzyTermQuery>());
+        assert!(!fuzzy.downcast::<FuzzyTermQuery>().unwrap().is_prefix());
+    }
+}
--- a/nidx/nidx_paragraph/src/fuzzy_query.rs
+++ b/nidx/nidx_paragraph/src/fuzzy_query.rs
@@ -215,6 +215,11 @@ impl FuzzyTermQuery {
         }
     }
 
+    #[cfg(test)]
+    pub fn is_prefix(&self) -> bool {
+        self.prefix
+    }
+
     fn specialized_weight(&self) -> tantivy::Result<AutomatonWeight<DfaWrapper>> {
         // LEV_BUILDER is a HashMap, whose `get` method returns an Option
         match LEV_BUILDER.get(&(self.distance, self.transposition_cost_one)) {


nucliadb/tests/nucliadb/integration/test_find.py
```diff
@@ -492,7 +492,7 @@ async def test_find_highlight(
     match = body["resources"].popitem()[1]["fields"]["/a/summary"]["paragraphs"].popitem()[1]
     assert match["order"] == 0
     assert match["score_type"] == "BM25"
-    assert "<mark>Marcus</mark> Aurelius" in match["text"]
+    assert "<mark>Marcus</mark> <mark>Aurelius</mark>" in match["text"]
```
jotare, line 495: This is a side effect of a better parsing. We were not matching Aurelius because it had a `?`. However, I think it makes sense to do so

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
```diff
@@ -0,0 +1,137 @@
...
+        match item {
+            Token::Literal(literal) => {
+                let term = Term::from_field_text(schema.text, literal);
+                let distance = FUZZY_DISTANCE;
+                let transposition_cost_one = true;
+
+                let q: Box<dyn Query>;
+                if literal.len() < MIN_FUZZY_LEN {
+                    // to avoid noise, we don't want to match too short terms as fuzzy
+                    q = Box::new(TermQuery::new(term, IndexRecordOption::Basic));
```
jotare, line 82: This is new, before we were doing fuzzy on everything without control, potentially bringing bad results. Again, this can lead to less differences between keyword and fuzzy queries, but limits the amount of randomness

nidx/nidx_paragraph/src/search_query.rs
```diff
@@ -293,20 +123,21 @@ fn filter_query(
...
-    let mut term_collector = TermCollector::default();
-    let processed = preprocess_raw_query(text, &mut term_collector);
-    let query = parse_query(parser, &processed.regular_query);
-    let fuzzy_query = parse_query(parser, &processed.fuzzy_query);
-    let termc = SharedTermC::from(term_collector);
-    let mut fuzzies = fuzzied_queries(fuzzy_query, true, distance, termc.clone());
-    let mut originals = vec![(Occur::Must, query)];
+    let query = &request.body;
+    let parser_config = ParserConfig::with_schema(schema).last_fuzzy_query_literal_as_prefix();
+    let parsed = query_parser::parse_query(query, parser_config);
```
javitonino: It might be a bit more ergonomic to have the `parse_query` function as a method of the config:

`let parsed = ParserConfig::with_schema(schema).last_fuzzy_query_literal_as_prefix().parse(query)`

I also find `.last_fuzzy_query_literal_as_prefix()` a bit long, maybe `.last_fuzzy_term_as_prefix()`?

Anyway, these are two very nitpicky things, question of preference. Just found this couple of lines a bit long and java-ish. But it's personal preference and what's its already here is very easy to understand.

nidx/nidx_paragraph/src/search_query.rs
```diff
@@ -321,62 +152,63 @@ pub fn suggest_query(
...
         let term_query = TermQuery::new(term, IndexRecordOption::Basic);
-        fuzzies.push((Occur::Must, Box::new(term_query.clone())));
-        originals.push((Occur::Must, Box::new(term_query)))
+        fuzzy_subqueries.push((Occur::Must, Box::new(term_query.clone())));
+        keyword_subqueries.push((Occur::Must, Box::new(term_query)))
     }
 
-    if originals.len() == 1 && originals[0].1.is::<AllQuery>() {
-        let original = originals.pop().unwrap().1;
+    if keyword_subqueries.len() == 1 && keyword_subqueries[0].1.is::<AllQuery>() {
```
javitonino: Do you know why is this here and if it's needed?

nidx/nidx_paragraph/src/request_types.rs
```diff
@@ -45,3 +46,20 @@ pub struct ParagraphSuggestRequest {
...
+    // Helper method to extract valid facets from the request
+    pub fn facets(&self) -> Vec<String> {
+        self.faceted
+            .as_ref()
+            .iter()
+            .flat_map(|faceted| faceted.labels.iter())
+            .filter(|label| self.is_valid_facet(label))
+            .cloned()
+            .collect()
+    }
```
javitonino, line 59: This is converting to a `Facet` to check if it's valid and then returns the string. The return value of this is passed to `Searcher` which calls to `Facet::from_text` again, to convert to the facet needed for the query. Not sure if it's possible for this to return `Vec<Facet>` (not sure if those facets are used anywhere else).

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+/// searched with certain Levenshtein distance. Quoted and excluded terms will
+/// remain the same, but literals will be elegible to be fuzzy. All long enough
+/// literals will be searched as fuzzy. The last literal can be searched as a
+/// fuzzy prefix if a suggest-like feature is desired.
+///
+pub fn parse_fuzzy_query(query: &[Token], term_collector: SharedTermC, last_literal_as_prefix: bool) -> Box<dyn Query> {
+    // We use usize::MAX as a discriminant when we don't want/care about the
+    // last index. This value won't match as there's no query that long.
+    let last_literal_index = if last_literal_as_prefix {
+        query.iter().enumerate().fold(
```
javitonino: This can also be expressed as `query.iter().position(||...)` which returns `Option<usize>` with the index of the first element matching the predicate. It might be clearer than using `fold`.

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
```diff
@@ -0,0 +1,141 @@
...
+            Token::Literal(literal) => {
+                let term = Term::from_field_text(schema.text, literal);
+                let distance = FUZZY_DISTANCE;
+                let transposition_cost_one = true;
+
+                let q: Box<dyn Query>;
+                if literal.len() < MIN_FUZZY_LEN {
+                    // to avoid noise, we don't want to match too short terms as fuzzy
+                    q = Box::new(TermQuery::new(term, IndexRecordOption::Basic));
+                } else if last_literal_as_prefix && (i == last_literal_index) && literal.len() >= MIN_FUZZY_PREFIX_LEN {
```
javitonino: `literal.len() >= MIN_FUZZY_PREFIX_LEN` is guaranteed to always be `false` here (otherwise, it should have gone into the first branch of the if). You can remove that part of the condition.

javitonino: Ah no, `MIN_FUZZY_LEN` and `MIN_FUZZY_PREFIX_LEN` are different things.

jotare: Mm... Maybe the names are too similar haha

nidx/nidx_paragraph/src/search_query.rs
jotare: Makes sense, we could use a `Parser` directly instead of config and the function. And like the rename, I struggled with a name that was clear enough that we were talking about the fuzzy query

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
javitonino: We could even decide to use the same constant, I don't think the exact number matters too much anyway :stuck_out_tongue:

nidx/nidx_paragraph/src/query_parser/keyword_parser.rs
```diff
@@ -0,0 +1,189 @@
...
+        match item {
+            Token::Literal(literal) => {
+                let q: Box<dyn Query> = Box::new(TermQuery::new(
+                    Term::from_field_text(schema.text, literal),
+                    IndexRecordOption::Basic,
+                ));
+                subqueries.push((Occur::Should, q));
+            }
+            Token::Quoted(quoted) => {
+                let mut terms: Vec<Term> = quoted
```
javitonino: I was wondering if this part of the code can be shared with the fuzzy parser? If I'm reading correctly, Literals are different for keyword/fuzzy, but phrase and excluded looks like they are the same?

nidx/nidx_paragraph/src/request_types.rs
jotare: Now Searcher owns the `Vec<Facet>`. Unfortunately we still have to copy things around when building the response, but we do a bit less

nidx/nidx_paragraph/src/search_query.rs
jotare: Not sure...

nidx/nidx_paragraph/src/query_parser/stop_words.rs
```diff
@@ -90,6 +92,40 @@ pub fn is_stop_word(word: &str) -> bool {
...
+    let query_len = query.len();
+
+    // as we don't want to remove the last literal, we pop the last element and
+    // always put it in the filtered query. We don't really care if the last
+    // it's a literal or not.
+    let last: Token<'_> = query.pop().unwrap(); // safe as query is not empty
+
+    query
+        .into_iter()
+        .take(query_len - 1)
```
javitonino: I think this `take` should not be here? `query.pop()` has already removed the last element, so this skips over the penultimate element.

nidx/nidx_paragraph/src/query_parser/keyword_parser.rs
jotare: Yes, they are the same. I repeated the code as it was not too long, but I can extract some smaller functions and reuse the code

nidx/nidx_paragraph/src/query_parser/stop_words.rs
jotare: Good catch! It was a leftover of a previous implementation. I'll add a test to make sure this is better tested

jotare:
[sc-12555]

nidx/nidx_paragraph/src/query_parser/keyword_parser.rs
jotare: Done

nidx/nidx_paragraph/src/search_query.rs
jotare: I think it was a weird combination of things that could happen. An `AllQuery` was the fallback for queries failing tantivy's parsing. So, IMO it was a shortcut like: if we can't parse a query and use an `AllQuery`, fuzzy won't be needed so use an empty query. But in a weird fashion

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
```diff
@@ -0,0 +1,117 @@
...
+    // We use usize::MAX as a discriminant when we don't want/care about the
+    // last index. This value won't match as there's no query that long.
+    let last_literal_index = if last_literal_as_prefix {
+        query
+            .iter()
+            .position(|token| matches!(token, &Token::Literal(_)))
+            .unwrap_or(usize::MAX)
+    } else {
+        usize::MAX
+    };
```
javitonino: I think it may be clearer to set `last_literal_index` to an Option:

```suggestion
    let last_literal_index = if last_literal_as_prefix {
        query
            .iter()
            .position(|token| matches!(token, &Token::Literal(_)))
    } else {
        None
    };
```

and then change the else if check to:

`} else if matches!(last_literal_index, Some(idx) if idx == i) && literal.len() >= MIN_FUZZY_PREFIX_LEN {`

or even to a match:

```rust
match (literal.len(), last_literal_index) {
    (..MIN_FUZZY_LEN, _) => {}
    (MIN_FUZZY_PREFIX_LEN.., Some(idx)) if idx == i => {}
    _ => {}
}
```

javitonino (approved):

nidx/nidx_paragraph/src/query_parser/fuzzy_parser.rs
jotare: Changing this back. We want the last position not the first

PR #3219, opened 2025-07-16
Data store v2

javitonino:
### Description

A second version of the DataStore which stores vectors on a separate file (for similarity calculations with less IO) and supports multiple vectors per paragraph.

Some design notes:
- Instead of `Address`, we now have `VectorAddr` and `ParagraphAddr` to pass around identifiers to records.
- Since there are multiple data stores, they are used with generics/boxed trait objects as needed. The design here has been to use generic in everything performance-critical which, for us, means HNSW indexing. So things like the HNSW search algorithms/`Retriever`/etc are generic to benefit from optimizations. The final steps of the search (retrieving paragraph metadata) are not and use boxed traits & virtual dispatch. This way we get performance were it matters and simplicity everywhere else.
- V2 datastore is enabled in per-index flag (in postgres metadata config). If not enabled, everything is V1. If enabled, all new segments and all new merges will produce V2 segments. Merges that produce a v2 store can be merged from any DataStore (v1, v2 or combinations).

### How was this PR tested?


nidx/nidx_vector/src/data_store/v2/paragraph_store.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn key(&self) -> &str {
+        self.key
+    }
+
+    pub fn metadata(&self) -> &[u8] {
+        self.metadata
+    }
+
+    pub fn labels(&self) -> Vec<String> {
+        self.labels.iter().copied().map(str::to_string).collect()
```
jotare: Do we need this `copied` if you're converting to string?

nidx/nidx_vector/src/data_point/mod.rs
```diff
@@ -215,180 +270,129 @@ pub fn create(path: &Path, elems: Vec<Elem>, config: &VectorConfig, tags: HashSe
...
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
-pub struct Address(usize);
+pub struct Address(pub(super) usize);
 
-pub struct Retriever<'a> {
+pub struct Retriever<'a, DS: DataStore> {
     similarity_function: fn(&[u8], &[u8]) -> f32,
-    no_nodes: usize,
+    no_vectors: usize,
```
jotare: NIT: maybe `vector_count` for consistency with the changes above?

jotare (approved):

nidx/nidx_vector/src/data_store/v2/paragraph_store.rs
javitonino: It's needed because `labels.iter()` is `impl Iterator<Item = &&str>` and `str::to_string` takes `&str`. The `copied` is just dereferencing the borrow here.

Anyway, I changed it to a closure which is probably more readable here.

PR #3223, opened 2025-07-17
Allow title reset on resource reprocess

lferran:
### Description
Occasionally, when processing link files we hit a firewall or the server rate limits Nuclia, ending up with an extracted title for the 404 page.

Upon resolving these issues, the clients end up having a lot of resources with the correct link extracted metadata but with the wrong resource title.

This PR adds the `reset_title` param to the `/reprocess`  endpoints so that, when set to True, the resource title will be set to the resulting link or file computed title of the new processing attempt.

### How was this PR tested?
Integration tests


nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -166,6 +167,13 @@ async def set_basic(
 
             self.basic.MergeFrom(payload)
 
+            # Make sure title is cleared if the flag is set
+            if self.basic.title == CLEAR_TITLE_FLAG:
+                # We set the title to the resource UUID so that it is not empty.
+                # If there is a computed title in the link or file extracted metadata
+                # by processing, it will be updated later.
+                self.basic.title = self.uuid
```
javitonino: If this is actually setting the title to the UUID, maybe we could avoid the `CLEAR_TITLE_FLAG` and set the title directly to uuid in the message to index? Just a thought, not sure if it's actually a good idea.

lferran: Yeah, you're right. One less indirection

lferran: changed!

PR #3225, opened 2025-07-18
Better reset title

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/ingest/orm/resource.py
+++ b/nucliadb/src/nucliadb/ingest/orm/resource.py
@@ -146,6 +146,21 @@ class Resource:
         if basic_in_payload.HasField("metadata") and basic_in_payload.metadata.useful:
             current_basic.metadata.status = basic_in_payload.metadata.status
 
+    async def title_marked_for_reset(self) -> bool:
+        return await datamanagers.resources.get_title_reset_mark(
+            self.txn, kbid=self.kb.kbid, rid=self.uuid
+        )
+
+    async def mark_title_for_reset(self):
+        await datamanagers.resources.set_title_reset_mark(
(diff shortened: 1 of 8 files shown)


nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -146,6 +146,17 @@ def set_processing_status(self, current_basic: PBBasic, basic_in_payload: PBBasi
         if basic_in_payload.HasField("metadata") and basic_in_payload.metadata.useful:
             current_basic.metadata.status = basic_in_payload.metadata.status
 
+    async def title_marked_for_reset(self) -> bool:
+        basic = await self.get_basic()
+        if basic is None:
+            return False
+        return basic.reset_title
```
javitonino: you can simplify this in this case. Or maybe it's just my ruby brain speaking xD

```suggestion
        return basic and basic.reset_title
```

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -624,10 +635,12 @@ async def maybe_update_resource_title_from_link(self, link_extracted_data: LinkE
             self.basic.title.startswith("http")
             or self.basic.title == ""
             or self.basic.title == self.uuid
+            or await self.title_marked_for_reset()
```
javitonino, line 636: Are you not removing the uuid check?

lferran: need to keep it, otherwise if someone creates a resource with a link but no title, the title is not replaced with the computed one (because writer sets the uuid as title by default, if not set)

PR #3227, opened 2025-07-18
Ask / Find with query image

carlesonielfa:
### Description
- [x] Add query image as a parameter in Ask / Find
- [x] Send query image to query endpoint and use POST method
- [x] Don't add `extra_image_context` or Rag Images Strategy to images if query image is received

This PR requires that https://github.com/nuclia/learning/pull/2488 is promoted

### How was this PR tested?
Added ask and find tests

[sc-12259]

### DIFF
--- a/nucliadb/tests/nucliadb/integration/test_find.py
+++ b/nucliadb/tests/nucliadb/integration/test_find.py
@@ -598,8 +598,15 @@ async def test_find_with_generative_model(
         },
     )
     assert resp.status_code == 200
-
-    spy.assert_called_once_with(kbid, "whatever", None, "everest", False, None, None)
+    spy.assert_called_once()
+    assert kbid == spy.call_args[0][0]
+    item = spy.call_args[0][1]
+    assert item.text == "whatever"
+    assert item.semantic_models is None
+    assert item.generative_model == "everest"
+    assert item.rephrase is False
+    assert item.rephrase_prompt is None
+    assert item.query_image is None
 
 
 @pytest.mark.deploy_modes("standalone")


nucliadb/src/nucliadb/search/predict.py
```diff
@@ -364,11 +366,13 @@ async def query(
...
             params["generative_model"] = generative_model
+        if query_image is not None:
+            params["query_image"] = query_image.model_dump()
 
         resp = await self.make_request(
-            "GET",
+            "POST",
             url=self.get_predict_url(QUERY, kbid),
-            params=params,
+            json=params,
```
lferran: nit: as we're moving to a post with payload, we could have an internal pydantic model defined in "nucliadb_search/predict_models.py".

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1891,6 +1896,11 @@ class FindRequest(BaseSearchRequest):
         title="Generative model",
         description="The generative model used to rephrase the query. If not provided, the model configured for the Knowledge Box is used.",
     )
+    query_image: Optional[Image] = Field(
+        default=None,
+        title="Query image",
+        description="Image that will be used together with the query text for retrieval.",
```
lferran, line 1903: Can you add a note here that if this is provided, the image rag strategies and the extra image context are not applied?

lferran (approved):

nucliadb/src/nucliadb/search/predict.py
carlesonielfa: Sure! Added it

nucliadb_models/src/nucliadb_models/search.py
carlesonielfa: yes!! Added it

lferran (approved):

PR #3232, opened 2025-07-22
Add support for encrypt S3 buckets with custom KMS key

rastut:
This pull request introduces support for using AWS KMS (Key Management Service) encryption for S3 buckets in the `nucliadb_utils` library. The changes include adding a new optional parameter for KMS key IDs, updating the S3 bucket creation logic to enable server-side encryption with KMS, and modifying the related settings and tests accordingly.

### Enhancements to S3 Storage:

* Added a new `kms_key_id` parameter to the `S3Storage` class constructor in `nucliadb_utils/storages/s3.py`, and updated the `create_bucket` method to pass this parameter to the bucket creation logic.
* Updated the `create_bucket` function to configure server-side encryption for S3 buckets using the provided KMS key ID, if available.

### Updates to Configuration:

* Added a new `s3_kms_key_id` field to the `Settings` class in `nucliadb_utils/storages/settings.py` to allow specifying the KMS key ID in the configuration. 

### Updates to Tests:

* Modified the test settings in `nucliadb_utils/tests/s3.py` to include a fake KMS key ID for testing purposes.
* Updated the `s3_storage` test helper to pass the `kms_key_id` parameter when initializing the `S3Storage` instance.


nucliadb_utils/src/nucliadb_utils/storages/s3.py
```diff
@@ -356,6 +356,7 @@ def __init__(
         verify_ssl: bool = True,
         use_ssl: bool = True,
         region_name: Optional[str] = None,
+        kms_key_id: Optional[str] = None,
```
javitonino, line 359: Can you move this to be the last argument? This is in nucliadb_utils which is shared with other repos, and I don't know if someone might be calling this function relying on argument order. I mean, it would be a terrrible idea to do so, but someone might be doing it, so just in case.

javitonino (approved):

nucliadb_utils/src/nucliadb_utils/storages/s3.py
rastut: Sure, done!

PR #3238, opened 2025-07-24
split strategy wiring

Ciniesta:
### Description
split strategy wiring
### How was this PR tested?
Updates unit tests


jotare (approved):
Changes look good but semgrep complains seem a bit worrisome. 

I think this only affects onprem. In hosted, the communication from nucliadb to processing is done inside the cluster, so exposing data there shouldn't be that big of a problem. However, we should review this.

PR #3245, opened 2025-08-08
Extracted Text positions index

lferran:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/ingest/fields/base.py
+++ b/nucliadb/src/nucliadb/ingest/fields/base.py
@@ -596,3 +596,101 @@ class Field(Generic[PbType]):
 
     async def get_value(self) -> Any:
         raise NotImplementedError()
+
+
+class ExtractedTextIndex:
+    def __init__(self, et: ExtractedText):
+        self.et = et
+        self.step_size = 1024
+        self.index: dict[int, int] = self.create_index()
+
+    def create_index(self) -> dict[int, int]:
+        # We assume that the dictionary will preserve the order of insertion
+        # (Python 3.7+), so we can use it to create an index
+        # that maps the position of the text in the ExtractedText to the byte
+        # position in the serialized ExtractedText.
+        index: dict[int, int] = {}
+        encoded = self.et.SerializeToString()
+        # Initialize the index with the start position by searching for the first two characters.
+        # This is a trick to know where the protobuff header ends.
+        text_starts_at = encoded.index(self.et.text[:2].encode())
+        index[0] = text_starts_at
+
+        for i in range(self.step_size, len(self.et.text), self.step_size):
+            text_slice = self.et.text[0:i]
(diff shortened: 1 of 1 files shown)


nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -596,3 +596,101 @@ async def set_value(self, value: Any):
...
+        index: dict[int, int] = {}
+        encoded = self.et.SerializeToString()
+        # Initialize the index with the start position by searching for the first two characters.
+        # This is a trick to know where the protobuff header ends.
+        text_starts_at = encoded.index(self.et.text[:2].encode())
+        index[0] = text_starts_at
+
+        for i in range(self.step_size, len(self.et.text), self.step_size):
+            text_slice = self.et.text[0:i]
+            index[i] = len(text_slice.encode()) + text_starts_at
```
javitonino, line 621: Here you are encoding the whole text up to this point. It would be better if you only encoded the current chunk (`[i-step_size : i]`) and then added the position of the previous chunk to the position of this one. This way, you only encoded each chunk once (right now you are encoding each part of the text chunks² times on average)

nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -596,3 +596,101 @@ async def set_value(self, value: Any):
...
+
+        # Add the end position of the text to the index
+        last_pos = len(self.et.text) - 1
+        if last_pos not in index:
+            index[last_pos] = len(self.et.text.encode()) + text_starts_at
```
javitonino, line 626: I think this is equivalent to `index[last_pos] = encoded.len()`, assumming the protobuf does not have more fields, but not sure if that's a fair assumption.

If not, same comment as above, try to avoid reencoding the entire text once more.

nucliadb/src/nucliadb/ingest/fields/base.py
```diff
@@ -596,3 +596,101 @@ async def set_value(self, value: Any):
...
+
+        assert len(index) > 0, "Index is empty, cannot create ExtractedTextIndex."
+        return index
+
+    def get_start(self, position: int) -> tuple[int, int]:
```
javitonino, line 631: `get_start` and `get_end` can be combined into a single function (`get_range`?) in order to save some iteration time over the index. It's not very relevant since the index would be small, but still.

Something like this (untested):

```python
def get_range(start_char, end_char):
    range_start = None
    for ((prev_char, prev_byte), (cur_char, cur_byte)) in itertools.pairwise(index.items()):
        if not range_start:
            if start_char == cur_char:
                range_start = (cur_char, cur_byte)
            elif start_char > prev_char:
                range_start = (prev_char, prev_byte)
        
        if range_start:
            if cur_char >= end_char:
                range_end = (cur_char, cur_byte)
                return (range_start, range_end)
    
    raise Exception("Positions outside index")
```

PR #3253, opened 2025-08-19
Fix reader select for update

lferran:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/common/maindb/pg.py
```diff
@@ -31,6 +30,7 @@
 
 from nucliadb.common.maindb.driver import DEFAULT_SCAN_LIMIT, Driver, Transaction
 from nucliadb.common.maindb.exceptions import ConflictError
+from nucliadb.ingest.settings import DriverSettings
```
jotare: You don't need to create an instance of the settings, you can directly
```suggestion
from nucliadb.ingest.settings import settings
```

jotare (approved):

PR #3266, opened 2025-08-28
Fix empty quotes token bug

jotare:
### Description
Our new tokenizer removes empty quotes (or full with whitespaces). However, quotes with only special characters can become empty due to tantivy retokenization step. We fix it by checking if we have emptied the token before adding it to the new token list

### How was this PR tested?
New unit tests


jotare:
[sc-12887]

PR #3273, opened 2025-09-04
Hydration API

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/hydrate.py
+++ b/nucliadb/src/nucliadb/search/api/v1/hydrate.py
@@ -1,0 +1,701 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+import base64
+from typing import Literal, Optional, Union, cast, overload
+
+from fastapi import Request, Response
+from fastapi_versioning import version
+
+from nucliadb.common.ids import FIELD_TYPE_STR_TO_NAME, FieldId, ParagraphId
+from nucliadb.common.maindb.utils import get_driver
+from nucliadb.common.models_utils import from_proto
+from nucliadb.ingest.fields.base import Field
+from nucliadb.ingest.fields.file import File
+from nucliadb.ingest.orm.resource import Resource
+from nucliadb.ingest.serialize import get_orm_resource
+from nucliadb.search import SERVICE_NAME
+from nucliadb.search.api.v1.router import KB_PREFIX, api
+from nucliadb.search.search import paragraphs
+from nucliadb.search.search.cache import request_caches
+from nucliadb.search.search.hydrator import hydrate_field_text
+from nucliadb_models import hydration as hydration_models
+from nucliadb_models.common import FieldTypeName
+from nucliadb_models.hydration import Hydrated, HydrateRequest, Hydration
+from nucliadb_models.resource import NucliaDBRoles
+from nucliadb_models.search import Image
+from nucliadb_models.security import ResourceSecurity
+from nucliadb_protos import resources_pb2
+from nucliadb_utils.authentication import requires
+from nucliadb_utils.utilities import get_storage
+
+
+@api.post(
+    f"/{KB_PREFIX}/{{kbid}}/hydrate",
+    status_code=200,
+    summary="",  # TODO
+    description="",  # TODO
+    response_model_exclude_unset=True,
+    tags=["Hydration"],
+)
+@requires(NucliaDBRoles.READER)
+@version(1)
+async def hydrate_endpoint(
+    request: Request,
+    response: Response,
+    kbid: str,
+    item: HydrateRequest,
+) -> Hydrated:
+    with request_caches():
+        return await Hydrator(kbid, item.hydration).hydrate(item.data)
+
+
+class HydratedBuilder:
+    """Builder class to construct an Hydrated payload."""
+
+    def __init__(self) -> None:
+        self._resources: dict[str, hydration_models.HydratedResource] = {}
+        self._fields: dict[
+            str,
+            Union[
+                hydration_models.HydratedTextField,
+                hydration_models.HydratedFileField,
+                hydration_models.HydratedLinkField,
+                hydration_models.HydratedConversationField,
+                hydration_models.HydratedGenericField,
+            ],
+        ] = {}
+        self._paragraphs: dict[str, hydration_models.HydratedParagraph] = {}
+
+    @property
+    def resources(self) -> dict[str, hydration_models.HydratedResource]:
+        return self._resources
+
+    @property
+    def fields(
+        self,
+    ) -> dict[
+        str,
+        Union[
+            hydration_models.HydratedTextField,
+            hydration_models.HydratedFileField,
+            hydration_models.HydratedLinkField,
+            hydration_models.HydratedConversationField,
+            hydration_models.HydratedGenericField,
+        ],
+    ]:
+        return self._fields
+
+    @property
+    def paragraphs(self) -> dict[str, hydration_models.HydratedParagraph]:
+        return self._paragraphs
+
+    def build(self) -> Hydrated:
+        return Hydrated(
+            resources=self._resources,
+            fields=self._fields,
+            paragraphs=self._paragraphs,
+        )
+
+    def add_resource(self, rid: str, slug: str) -> hydration_models.HydratedResource:
+        hydrated = hydration_models.HydratedResource(id=rid, slug=slug)
+        self._resources[rid] = hydrated
+        return hydrated
+
+    @overload
+    def add_field(
+        self, field_id: FieldId, field_type: Literal[FieldTypeName.TEXT]
+    ) -> hydration_models.HydratedTextField: ...
+
+    @overload
+    def add_field(
+        self, field_id: FieldId, field_type: Literal[FieldTypeName.FILE]
+    ) -> hydration_models.HydratedFileField: ...
+
+    @overload
+    def add_field(
+        self, field_id: FieldId, field_type: Literal[FieldTypeName.LINK]
+    ) -> hydration_models.HydratedLinkField: ...
+
+    @overload
+    def add_field(
+        self, field_id: FieldId, field_type: Literal[FieldTypeName.CONVERSATION]
+    ) -> hydration_models.HydratedConversationField: ...
+
+    @overload
+    def add_field(
+        self, field_id: FieldId, field_type: Literal[FieldTypeName.GENERIC]
+    ) -> hydration_models.HydratedGenericField: ...
+
+    def add_field(self, field_id: FieldId, field_type: FieldTypeName):
+        hydrated: Union[
+            hydration_models.HydratedTextField,
+            hydration_models.HydratedFileField,
+            hydration_models.HydratedLinkField,
+            hydration_models.HydratedConversationField,
+            hydration_models.HydratedGenericField,
+        ]
+
+        if field_type == FieldTypeName.TEXT:
+            hydrated = hydration_models.HydratedTextField(
+                id=field_id.full(),
+                resource=field_id.rid,
+                field_type=field_type,
+            )
+
+        elif field_type == FieldTypeName.FILE:
+            hydrated = hydration_models.HydratedFileField(
+                id=field_id.full(),
+                resource=field_id.rid,
+                field_type=field_type,
+            )
+
+        elif field_type == FieldTypeName.LINK:
+            hydrated = hydration_models.HydratedLinkField(
+                id=field_id.full(),
+                resource=field_id.rid,
+                field_type=field_type,
+            )
+
+        elif field_type == FieldTypeName.CONVERSATION:
+            hydrated = hydration_models.HydratedConversationField(
+                id=field_id.full(),
+                resource=field_id.rid,
+                field_type=field_type,
+            )
+
+        elif field_type == FieldTypeName.GENERIC:
+            hydrated = hydration_models.HydratedGenericField(
+                id=field_id.full(),
+                resource=field_id.rid,
+                field_type=field_type,
+            )
+
+        else:
+            # This is a trick so mypy generates an error if this branch can be reached,
+            # that is, if we are missing some ifs
+            _a: int = "a"
+
+        self._fields[field_id.full()] = hydrated
+        return hydrated
+
+    def add_paragraph(self, paragraph_id: ParagraphId) -> hydration_models.HydratedParagraph:
+        hydrated = hydration_models.HydratedParagraph(
+            id=paragraph_id.full(),
+            field=paragraph_id.field_id.full(),
+            resource=paragraph_id.rid,
+        )
+        self._paragraphs[paragraph_id.full()] = hydrated
+        return hydrated
+
+    def add_page_preview(self, paragraph_id: ParagraphId, page: int, image: Image) -> str:
+        field_id = paragraph_id.field_id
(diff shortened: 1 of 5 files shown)


nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+        elif field_type == FieldTypeName.LINK:
+            await self._hydrate_link_field(resource, field_id)
+
+        elif field_type == FieldTypeName.CONVERSATION:
+            await self._hydrate_conversation_field(resource, field_id)
+
+        elif field_type == FieldTypeName.GENERIC:
+            await self._hydrate_generic_field(resource, field_id)
+
+        else:
```
lferran: you can add `# pragma: no cover` here

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+
+
+@api.post(
+    f"/{KB_PREFIX}/{{kbid}}/hydrate",
+    status_code=200,
+    summary="Hydrate a set of paragraphs",
+    description="Internal API endpoint to hydrate a set of paragraphs",
+    response_model_exclude_unset=True,
+    tags=["Hydration"],
+)
```
lferran, line 64: if you don't want it to appear in the openapi specs, you should add `include_in_schema=False` here

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+    def add_field(self, field_id: FieldId, field_type: FieldTypeName):
+        hydrated: Union[
+            hydration_models.HydratedTextField,
+            hydration_models.HydratedFileField,
+            hydration_models.HydratedLinkField,
+            hydration_models.HydratedConversationField,
+            hydration_models.HydratedGenericField,
+        ]
+
+        if field_type == FieldTypeName.TEXT:
```
lferran: As all parameters are the same for all fields, it would be less verbose if you did something like
```python
 hydrated_cls = {
   FieldTypeName.TEXT: hydration_models.HydratedTextField,
   FieldTypeName.FILE: hydaration_models.HydratedFileField,
   ...
   }[field_type]
   
   hydrated = hydrated_cls(
       id=field_id.full(),
       resource=field_id.rid,
       field_type=field_type,
    )
```
But feel free to ignore the comment, this is just a nit.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+        assert paragraph.page is not None, "should already be set"
+        paragraph.page.page_preview_ref = preview_id
+
+        return preview_id
+
+    def add_table_page_preview(self, paragraph_id: ParagraphId, page: int, image: Image) -> str:
+        field_id = paragraph_id.field_id
+        field = self._fields[field_id.full()]
+        # TODO: implement link previews
+        assert isinstance(field, hydration_models.HydratedFileField), (
```
lferran: This will fail right now no? Why not simply returning and leave a warning, so we make sure we implement this at some point?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+            resource=paragraph_id.rid,
+        )
+        self._paragraphs[paragraph_id.full()] = hydrated
+        return hydrated
+
+    def add_page_preview(self, paragraph_id: ParagraphId, page: int, image: Image) -> str:
+        field_id = paragraph_id.field_id
+        field = self._fields[field_id.full()]
+        # TODO: implement link previews
+        assert isinstance(field, hydration_models.HydratedFileField), (
```
lferran: same, what about a warning + return instead of raising AssertionError when it is not a file?

nucliadb_models/src/nucliadb_models/hydration.py
```diff
@@ -0,0 +1,411 @@
...
+        description="Field hydration options",
+    )
+    paragraph: ParagraphHydration = Field(
+        default_factory=ParagraphHydration,
+        description="Paragraph hydration options",
+    )
+
+
+class HydrateRequest(BaseModel, extra="forbid"):
+    data: list[str]
```
lferran: Can you add field descriptions?
Also, should we validate this a bit more?
Like, making sure the paragraphs have the right format, adding limits on number of text blocks to hydrate in one request, etc.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+        self.field_paragraphs: dict[FieldId, ParagraphIndex] = {}
+
+    async def hydrate(self, paragraph_ids: list[str]) -> Hydrated:
+        driver = get_driver()
+        db_resources = {}
+        async with driver.ro_transaction() as txn:
+            for raw_paragraph_id in paragraph_ids:
+                # TODO: remove this HACK skip
+                if raw_paragraph_id.startswith(":"):
+                    continue
```
lferran: What is this hack about?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+
+                # TODO if we accept user ids, we must validate them
+                paragraph_id = ParagraphId.from_string(raw_paragraph_id)
```
lferran: yup, agree! Ideally the validation is done at the pydantic/fastapi layer

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+        self.hydrated = HydratedBuilder()
+
+        # cached paragraphs per field
+        self.field_paragraphs: dict[FieldId, ParagraphIndex] = {}
+
+    async def hydrate(self, paragraph_ids: list[str]) -> Hydrated:
+        driver = get_driver()
+        db_resources = {}
+        async with driver.ro_transaction() as txn:
+            for raw_paragraph_id in paragraph_ids:
```
lferran: This hydrates one paragraph after the other. Why not doing it in parallel, similarly to what we are doing on find or ask.

This look could still be "registering" the stuff that needs to be fetched, but then the actual fetching of the data could be done completely in parallel (with some limits, if you want).

lferran (commented):
reviewed the first layers so far.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+                    continue
+
+                # TODO if we accept user ids, we must validate them
+                paragraph_id = ParagraphId.from_string(raw_paragraph_id)
+                field_id = paragraph_id.field_id
+                rid = paragraph_id.rid
+
+                if rid not in db_resources:
+                    resource = await get_orm_resource(txn, self.kbid, rid, service_name=SERVICE_NAME)
+                    assert resource is not None
```
lferran: If resource is None, you should skip it

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+            security = await resource.get_security()
+            hydrated.security = ResourceSecurity(access_groups=[])
+            if security is not None:
+                for group_id in security.access_groups:
+                    hydrated.security.access_groups.append(group_id)
+
+        if self.config.resource.origin:
+            origin = await resource.get_origin()
+            if origin is not None:
+                # TODO: we want a better hydration than proto to JSON
```
lferran: why? what did you have in mind otherwise?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+                if rid not in db_resources:
+                    resource = await get_orm_resource(txn, self.kbid, rid, service_name=SERVICE_NAME)
+                    assert resource is not None
+                    db_resources[rid] = resource
+                resource = db_resources[rid]
+
+                if rid not in self.hydrated.resources:
+                    await self._hydrate_resource(resource, rid)
+
+                if field_id.full() not in self.hydrated.fields:
```
lferran: are fields always hydrated though? Some may only need the paragraph, maybe? Or at least the option to disable it.

Same goes with resources.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+            # that is, if we are missing some ifs
+            _a: int = "a"
+
+    async def _hydrate_text_field(
+        self, resource: Resource, field_id: FieldId
+    ) -> hydration_models.HydratedTextField:
+        hydrated = self.hydrated.add_field(field_id, FieldTypeName.TEXT)
+
+        if not self.config.field.text:
+            return hydrated
```
lferran: nit: imo this would be cleaner if moved at the `_hydrate_field` method, on the if checks.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+            if origin is not None:
+                # TODO: we want a better hydration than proto to JSON
+                hydrated.origin = from_proto.origin(origin)
+
+        return hydrated
+
+    async def _hydrate_field(self, resource: Resource, field_id: FieldId):
+        field_type = FIELD_TYPE_STR_TO_NAME[field_id.type]
+
+        if field_type == FieldTypeName.TEXT:
```
lferran: what about 
```python
if field_type == FieldTypeName.TEXT and self.config.field.text:
```

nucliadb/src/nucliadb/search/api/v1/hydrate.py
lferran: otherwise you are adding a field to the payload that won't be hydrated no?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+        hydrated = self.hydrated.add_field(field_id, FieldTypeName.TEXT)
+
+        if not self.config.field.text:
+            return hydrated
+
+        config = self.config.field.text
+
+        if config.extracted_text:
+            field_text = await hydrate_field_text(self.kbid, field_id)
+            assert field_text is not None
```
lferran: if text is not found, simply ignore instead of assertion error

lferran: same applies to all other asserts

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+            (_, text) = field_text
+            hydrated.extracted = hydration_models.FieldExtractedData(text=text)
+
+        return hydrated
+
+    async def _hydrate_paragraph(self, resource: Resource, paragraph_id: ParagraphId):
+        if paragraph_id.full() in self.hydrated.paragraphs:
+            hydrated = self.hydrated.paragraphs[paragraph_id.full()]
```
lferran: in this case, will the paragraph be hydrated twice?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+
+            if config.related.replacements:
+                hydrated.related.replacements = []
+                for replacement_id in field_paragraphs.replacements(paragraph_id):
+                    await self._hydrate_related_paragraph(
+                        ParagraphId.from_string(replacement_id),
+                    )
+                    hydrated.related.replacements.append(replacement_id)
+
+        paragraph = field_paragraphs.get(paragraph_id)
```
lferran: what if paragraph is None?

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+                    # Paragraphs extracted from an image store its original image
+                    # representation in the reference file. The path is incomplete
+                    # though, as it's stored in the `generated` folder
+                    image = await download_image(
+                        self.kbid,
+                        field_id,
+                        f"generated/{source_image}",
+                        # TODO: we assume all reference files are PNG images, but
+                        # this actually depends on learning so it's a dangerous
+                        # assumption. We should check it by ourselves
```
lferran: I would say it's ok. just another implicit contract. I don't think we can do anything on our side. Leaving a comment is good enough at this point, imo !

nucliadb_models/src/nucliadb_models/hydration.py
```diff
@@ -0,0 +1,411 @@
...
+        description="Resource hydration options",
+    )
+    field: FieldHydration = Field(
+        default_factory=FieldHydration,
+        description="Field hydration options",
+    )
+    paragraph: ParagraphHydration = Field(
+        default_factory=ParagraphHydration,
+        description="Paragraph hydration options",
+    )
```
lferran: should these be optional? meaning, some may want to skip hydrating resources, for example. I agree that by default all should be not None though.

nucliadb/src/nucliadb/search/api/v1/hydrate.py
jotare: I just prefer to stay away from the proto-to-json, as it couples internal models with public APIs

nucliadb/src/nucliadb/search/api/v1/hydrate.py
jotare: There's some dependency between paragraphs and fields. A paragraph with visual can trigger a field to be hydrated with pages. That's why on a first approach, I hydrated everything. I'll do it more granular once I flip the order of hydration

nucliadb/src/nucliadb/search/api/v1/hydrate.py
jotare: This was intentional, as some paragraphs will later require the field to exist. I moved the if to `_hydrate_field` and added some hacky add to refactor that part

nucliadb/src/nucliadb/search/api/v1/hydrate.py
jotare: Yes, if the paragraph has been hydrated due to a neighbour but it's also in the list, we want to end the hydration. I'll change the order of things so it gets more clear

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,758 @@
...
+from nucliadb_models.search import Image
+from nucliadb_models.security import ResourceSecurity
+from nucliadb_protos import resources_pb2
+from nucliadb_protos.resources_pb2 import FieldComputedMetadata
+from nucliadb_utils.authentication import requires
+from nucliadb_utils.utilities import get_storage
+
+
+@api.post(
+    f"/{KB_PREFIX}/{{kbid}}/hydrate",
```
lferran, line 57: For this to work in the cluster, you'll need to add these to the virtual service.

I wonder if we should have the endpoint in the reader component instead, as it is purely fetching stuff from maindb and storage. wdyt?

jotare: I don't think it matters much. We don't have problems in any of those components. All are scalable and search is doing more or less the same in other endpoints

lferran: I was more concerned in terms of memory usage. but we can work on this once it is a problem

lferran: you will need the virtual service though

nucliadb/src/nucliadb/search/api/v1/hydrate.py
```diff
@@ -0,0 +1,317 @@
...
+                        self.hydrated.add_table_page_preview(paragraph_id, page_number, preview)
+
+        return self.hydrated.build()
+
+    # TODO: proper typing
+    async def _limited_concurrency(self, aw: Awaitable):
+        async with self.max_ops:
+            return await aw
+
+    @alru_cache(maxsize=None)
```
lferran, line 326: This will store the downloaded pages forever in memory, consider using a cache object or a dict in the hydrate function directly. wdyt?

jotare: No, it doesn't. As the hydrator is an instance, the cache is local to that hydrator instance, not global

nucliadb_models/src/nucliadb_models/hydration.py
jotare: Resource is optional now. Paragraph left as it's kind of weird to ask for paragraphs for the other things. We can tune it once we allow selecting things other than paragraphs.

PR #3281, opened 2025-09-15
Retry bucket creation errors

lferran:
### Description
Some additional fly-by changes removing repeated session initialization checks

### How was this PR tested?
Existing tests should cover us. I'm also testing against the actual cluster.

### DIFF
--- a/nucliadb_utils/src/nucliadb_utils/storages/gcs.py
+++ b/nucliadb_utils/src/nucliadb_utils/storages/gcs.py
@@ -43,7 +43,6 @@ from nucliadb_utils import logger
 from nucliadb_utils.storages import CHUNK_SIZE
 from nucliadb_utils.storages.exceptions import (
     CouldNotCopyNotFound,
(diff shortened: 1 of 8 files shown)


nucliadb_utils/src/nucliadb_utils/storages/gcs.py
```diff
@@ -484,15 +470,15 @@ def __init__(
         self.source = CloudFile.GCS
         self.deadletter_bucket = deadletter_bucket
         self.indexing_bucket = indexing_bucket
-        self.bucket = bucket
+        self.bucket = bucket or ""
```
jotare: Is this safe? In which case we need a GCSStorage instance without knowing the bucket?

lferran: yeah I guess it doesn't make sense. it was like that already but mypy started complaining.
I will check if its simple to address , and if so, add to this PR

PR #3301, opened 2025-10-01
Do not index the full extracted text of a conversation

lferran:
### Description

Conversational fields can grow to a very large number of messages, being each message potentially as big as a regular field (text, file, link).

Right now, we were appending all the extracted texts of all messages into a single tantivy document on the texts index. This definitely does not scale, so we decided to not index the fulltext of the conversation fields, with the only caveat of not being able to do keyword filtering on these. For keyword filtering to work, we would need to rework the prefiltering on nidx to be done with the paragraphs index, for instance. 

All other filters will work normally as despite not indexing the fulltext, we are indexing the labels on the texts index.

[sc-12775]

### How was this PR tested?
Existing tests + extended conversation integration tests

### DIFF
--- a/nidx/nidx_paragraph/src/resource_indexer.rs
+++ b/nidx/nidx_paragraph/src/resource_indexer.rs
@@ -82,10 +82,9 @@ pub fn index_paragraphs(
             let split = &p.split;
 
             // The text of the paragraph can be specified in the paragraph message, otherwise slice it from the full field text.
-            let text: String;
-            let text_ref: &str;
+            let text_ref: String;
             if let Some(t) = &p.text {
-                text_ref = t;
+                text_ref = t.to_string();
             } else {
                 let lower_bound = std::cmp::min(start_pos as usize, chars.len());
                 let upper_bound = std::cmp::min(end_pos as usize, chars.len());
@@ -93,8 +92,8 @@ pub fn index_paragraphs(
                     warn!("Skipping paragraph with wrong start and end positions");
                     continue;
                 }
-                text = chars[lower_bound..upper_bound].iter().collect();
-                text_ref = &text;
+                let slice = &chars[lower_bound..upper_bound];
+                text_ref = slice.iter().collect();
             }
             let facet_field = format!("/{field}");
             let paragraph_labels = p
--- a/nidx/nidx_paragraph/src/resource_indexer.rs
+++ b/nidx/nidx_paragraph/src/resource_indexer.rs
@@ -79,9 +79,18 @@ pub fn index_paragraphs(
             let end_pos = p.end as u64;
             let index = p.index;
             let split = &p.split;
-            let lower_bound = std::cmp::min(start_pos as usize, chars.len());
-            let upper_bound = std::cmp::min(end_pos as usize, chars.len());
-            let text: String = chars[lower_bound..upper_bound].iter().collect();
+
+            // The text of the paragraph can be specified in the paragraph message, otherwise slice it from the full field text.
+            let text: String;
+            let text_ref: &str;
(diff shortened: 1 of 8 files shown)


nidx/nidx_paragraph/src/resource_indexer.rs
```diff
@@ -79,9 +79,18 @@ pub fn index_paragraphs(
             let end_pos = p.end as u64;
             let index = p.index;
             let split = &p.split;
-            let lower_bound = std::cmp::min(start_pos as usize, chars.len());
-            let upper_bound = std::cmp::min(end_pos as usize, chars.len());
-            let text: String = chars[lower_bound..upper_bound].iter().collect();
+
+            // The text of the paragraph can be specified in the paragraph message, otherwise slice it from the full field text.
+            let text: String;
```
javitonino: This variable does not appear to be in use, remove it?

nidx/nidx_paragraph/src/resource_indexer.rs
```diff
@@ -79,9 +79,18 @@ pub fn index_paragraphs(
...
+            let text_ref: &str;
+            if let Some(t) = &p.text {
+                text_ref = t;
+            } else {
+                let lower_bound = std::cmp::min(start_pos as usize, chars.len());
+                let upper_bound = std::cmp::min(end_pos as usize, chars.len());
```
javitonino, line 90: This is silently eating errors. If start_pos and end_pos are wrong, we will try to index an empty string. Maybe it's better to just skip indexing the paragraph in this case.

nidx/nidx_vector/src/searcher.rs
```diff
@@ -667,6 +669,7 @@ mod tests {
             split: "".to_string(),
             repeated_in_field: false,
             metadata: None,
+            ..Default::default()
```
javitonino, line 672: It would be nice to have a test somewhere that uses the `text` field to make sure it works.

nidx/nidx_paragraph/src/resource_indexer.rs
```diff
@@ -79,9 +80,21 @@ pub fn index_paragraphs(
...
+                text_ref = t.to_string();
+            } else {
+                let lower_bound = std::cmp::min(start_pos as usize, chars.len());
+                let upper_bound = std::cmp::min(end_pos as usize, chars.len());
+                if lower_bound >= upper_bound {
+                    warn!("Skipping paragraph with wrong start and end positions");
+                    continue;
+                }
+                let slice = &chars[lower_bound..upper_bound];
+                text_ref = slice.iter().collect();
```
javitonino, line 96: With the `min(start_pos, chars.len)` thing we still could ignore invalid ranges. Maybe move it to a check in the if, for example:

```suggestion
                if start_pos >= end_pos || end_pos > chars.len() {
                    warn!("Skipping paragraph with wrong start and end positions");
                    continue;
                }
                let slice = chars[start_pos..end_pos];
                text_ref = slice.iter().collect();
```

nidx/nidx_paragraph/tests/reader.rs
```diff
@@ -505,3 +505,94 @@ fn test_query_parsing_weird_stuff() -> anyhow::Result<()> {
...
+        order: None,
+        result_per_page: 20,
+        only_faceted: false,
+        ..Default::default()
+    };
+
+    search.body = "meaning of life".to_string();
+    let result = paragraph_reader_service.search(&search, &PrefilterResult::All).unwrap();
+    assert_eq!(result.total, 1);
+    Ok(())
```
javitonino, line 597: cool :sparkles:

PR #3303, opened 2025-10-02
Index only appended messages

lferran:
### Description

When appending messages to a conversation, we were reindexing the whole field always in all indexes (fulltext, relations, paragraphs and vectors).

This PR improves the logic so at least we don't reindex the paragraphs and vectors for all messages every time a new message is appended.

[sc-12775]

### How was this PR tested?
Existing tests modified

### DIFF
--- a/nucliadb/src/migrations/0039_backfill_converation_splits_metadata.py
+++ b/nucliadb/src/migrations/0039_backfill_converation_splits_metadata.py
@@ -47,11 +47,11 @@ async def migrate(context: ExecutionContext) -> None: ...
 
 async def migrate_kb(context: ExecutionContext, kbid: str) -> None:
     BATCH_SIZE = 100
-    async with context.kv_driver.rw_transaction() as txn:
-        txn = cast(PGTransaction, txn)
-        start = ""
-        while True:
-            to_fix: list[tuple[str, str]] = []
+    start = ""
+    while True:
+        to_fix: list[tuple[str, str]] = []
+        async with context.kv_driver.rw_transaction() as txn:
+            txn = cast(PGTransaction, txn)
             async with txn.connection.cursor() as cur:
                 # Retrieve a bunch of conversation fields
                 await cur.execute(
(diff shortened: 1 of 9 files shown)


nucliadb/tests/ingest/unit/orm/test_brain.py
```diff
@@ -202,19 +207,88 @@ def test_apply_field_paragraphs_populates_page_number():
...
+        field_key,
+        field_computed_metadata=fcm,
+        page_positions={},
+        extracted_text=extracted_text,
+        user_field_metadata=None,
+        replace_field=False,
+        skip_paragraphs=False,
+        append_splits=None,
+    )
+    assert len(br.brain.paragraphs[field_key].paragraphs) == 1
```
javitonino, line 250: Maybe it would be good to check that the brain does not contain deletions. Not sure if it can be checked easily (calling `apply_field_paragraphs` might not be enough for this to be meaningful), but we have had lots of problems with deletions so I think it might be worth being extra careful.

nucliadb/src/nucliadb/ingest/orm/index_message.py
```diff
@@ -214,6 +218,21 @@ async def for_processor_bm(
...
             if fieldid in message.delete_fields:
                 continue
+
+            # For conversation fields, we only replace the full field if it is not an append messages operation.
+            # All other fields are always replaced upon modification.
+            replace_field = True
+            modified_splits: Optional[list[str]] = None
+            if fieldid.field_type == FieldType.CONVERSATION:
+                modified_splits = await get_bm_modified_split_ids(fieldid, message, self.resource)
+                stored_splits = await get_stored_split_ids(fieldid, self.resource)
```
javitonino, line 229: This can be a bit heavy (retrieve the full extracted text) and kind of defeats the purpose of the PR. That is, we are trying to make conversation fields more scalable, but we are loading all messages just to check if this is an append.

Can we figure out another way to do this check? A quick idea I had, is check if the split numbers in the Broker Message start at 1 or not. But not sure if that works. Another way would be to store some maindb metadata about the number of splits in a field, that we might want to do for other parts of this features as well?

Anyway, not sure about this, it just feels like there must be a better way to do this check.

lferran: changed, let me know what you think of the new approach

nucliadb/src/migrations/0039_backfill_converation_splits_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+async def migrate(context: ExecutionContext) -> None: ...
+
+
+async def migrate_kb(context: ExecutionContext, kbid: str) -> None:
+    BATCH_SIZE = 100
+    async with context.kv_driver.rw_transaction() as txn:
```
javitonino: This transaction should probably be read-only and be inside the loop (same level as the cursor), to avoid a long-lived transaction.

PR #3310, opened 2025-10-07
Rerank using RaBitQ error bounds

javitonino:
### Description

RaBitQ provides error bound that can be used to stop reranking vectors early. If the upper bound of the score estimate is worse than the best reranked vector so far, we can skip calculating the actual similarity score. This reduces the number of raw vectors loaded from disk significantly. In a quick experiment with the current setting this reduces the loaded vectors by ~50% while keeping the exact same results.

### How was this PR tested?


nidx/nidx_vector/src/hnsw/ops_hnsw.rs
```diff
@@ -67,6 +67,28 @@ pub trait Hnsw {
     fn get_layer(&self, i: usize) -> Self::L;
 }
 
+#[derive(Clone, Copy)]
+pub struct ScoreBound {
```
jotare: NIT: What do you think about `EstimatedScore` or something like that? It reads a bit better than `ScoreBound`, as it's a score with an error more than a bound on a score

PR #3588, opened 2026-04-09
Store extracted text in nidx

jotare:
### Description
Object storage response time for extracted text blobs is around 50-200ms depending on the region, activity... This is a first approach towards storing the text in nidx, which provides response times in the order or µs inside nidx and ~10ms for a gRPC call (mostly network).

This is a quite naive implementation and many things can be improved

### How was this PR tested?

### DIFF
--- a/nidx/nidx_protos/nidx.proto
+++ b/nidx/nidx_protos/nidx.proto
@@ -66,7 +66,7 @@ message ExtractedTextsRequest {
         string rid = 1;
         string field_type = 2;
         string field_name = 3;
-        optional string split = 4;
+        // optional string split = 4;
     }
 
     // equivalent to {rid}/{field_type}/{field_id}[/{split}]/{paragraph_start}-{paragraph_end}
@@ -74,7 +74,7 @@ message ExtractedTextsRequest {
         string rid = 1;
         string field_type = 2;
         string field_name = 3;
-        optional string split = 4;
+        // optional string split = 4;
         uint32 paragraph_start = 5;
         uint32 paragraph_end = 6;
     }
--- a/nidx/nidx_protos/nidx.proto
+++ b/nidx/nidx_protos/nidx.proto
@@ -69,7 +69,7 @@ message ExtractedTextsRequest {
         optional string split = 4;
     }
 
-    // equivalent to {rid}/{field_type}/{field_id}/{paragraph_start}-{paragraph_end}
+    // equivalent to {rid}/{field_type}/{field_id}[/{split}]/{paragraph_start}-{paragraph_end}
     message ParagraphId {
         string rid = 1;
         string field_type = 2;
--- a/nidx/nidx_protos/nidx.proto
+++ b/nidx/nidx_protos/nidx.proto
@@ -22,6 +22,8 @@ service NidxSearcher {
   rpc GraphSearch(nodereader.GraphSearchRequest) returns (nodereader.GraphSearchResponse) {}
   rpc Suggest(nodereader.SuggestRequest) returns (nodereader.SuggestResponse) {}
 
+  rpc ExtractedTexts(ExtractedTextsRequest) returns (ExtractedTextsResponse) {}
+
   // Streams
   rpc Paragraphs(nodereader.StreamRequest) returns (stream nodereader.ParagraphItem) {}
   rpc Documents(nodereader.StreamRequest) returns (stream nodereader.DocumentItem) {}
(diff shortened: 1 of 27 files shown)


nidx/nidx_protos/nidx.proto
```diff
@@ -57,3 +59,36 @@ message ShardConfig {
...
     bool prewarm_enabled = 2;
 }
+
+message ExtractedTextsRequest {
+    // equivalent to {rid}/{field_type}/{field_id}[/{split}]
+    message FieldId {
+        string rid = 1;
+        string field_type = 2;
+        string field_name = 3;
+        // optional string split = 4;
```
javitonino: Still TODO?

nucliadb/src/nucliadb/common/cluster/rollover.py
```diff
@@ -165,6 +165,9 @@ async def create_rollover_shards(
                         nucliadb_index_config_to_nidx(model.vectorset_index_config)
                     )
 
+            if has_feature(const.Features.NIDX_AS_EXTRACTED_TEXT_STORAGE):
+                req.features.append(const.Features.NIDX_AS_EXTRACTED_TEXT_STORAGE)
```
javitonino, line 169: No KBID/account passed to `has_feature`? We probably want something like this to scope to some test KBs. As it is now, it's only per environment.

nucliadb/src/nucliadb/common/cache.py
```diff
@@ -116,23 +121,72 @@ def __init__(self, cache_size: int) -> None:
...
-                    "Error getting extracted text for field. Retrying",
-                    exc_info=True,
-                    extra={
-                        "kbid": kbid,
-                        "resource_id": field_id.rid,
-                        "field": f"{field_id.type}/{field_id.key}",
-                    },
+                text = extracted_texts.fields.get(field_id.full_without_subfield(), "")
+
+                logger.info(
```
javitonino: Maybe replace this log for a metric? (same in other places)

nucliadb_utils/src/nucliadb_utils/featureflagging.py
```diff
@@ -57,16 +68,44 @@ class Settings(pydantic_settings.BaseSettings):
...
-            self.flag_service = mrflagly.FlagService(data=json.dumps(DEFAULT_FLAG_DATA))
+            self.flag_service = mrflagly.FlagService(data=json.dumps(DEFAULT_FLAG_DATA))  # type: ignore[attr-defined]
         else:
-            self.flag_service = mrflagly.FlagService(url=settings.flag_settings_url)
+            self.flag_service = mrflagly.FlagService(url=settings.flag_settings_url)  # type: ignore[attr-defined]
+
+        # We are transitioning from mr. flaggly to Flipt. Meanwhile, we'll have
+        # both clients and check both places
+        self.flipt_enabled = settings.flipt_token is not None
+        print("FLIPT ENABLED", self.flipt_enabled)
```
javitonino: print! (and more in flipt code)

nidx/nidx_protos/nidx.proto
jotare: Yeah, I was thinking on adding support for splits in the next iteration, but I can do it now

nucliadb/src/nucliadb/common/cluster/rollover.py
jotare: My bad! Good catch!

nucliadb_utils/src/nucliadb_utils/featureflagging.py
jotare: I know I know, it was to test a bit in stage too with some extra debugging. I can change them for logger.debug/logger.info

nucliadb/src/nucliadb/common/cache.py
jotare: Sure, these are only debug logs while developing. I can remove them now or in a next PR

nucliadb_utils/src/nucliadb_utils/featureflagging.py
jotare: I left some better logs useful while deploying in different envs. We can remove them once we deploy to prod

nucliadb/src/nucliadb/common/cache.py
jotare: Removed

nidx/nidx_protos/nidx.proto
jotare: Added support

nucliadb_utils/src/nucliadb_utils/featureflagging.py
```diff
@@ -57,16 +69,51 @@ class Settings(pydantic_settings.BaseSettings):
...
+                try:
+                    evaluation = self.client.evaluate_boolean(
+                        flag_key=flag_key,
+                        entity_id=self.entity_id,
+                        context=context,
+                    )
+                except EvaluationError as exc:
+                    logger.exception("Flipt FF evaluation failed", exc_info=exc)
+                else:
+                    logger.info(f"Flipt evaluation of {flag_key} for {context} was {evaluation}")
```
javitonino: This log is going to be noisy, maybe remove it or downgrade to debug?

nucliadb_utils/src/nucliadb_utils/featureflagging.py
```diff
@@ -57,16 +69,51 @@ class Settings(pydantic_settings.BaseSettings):
...
     def enabled(self, flag_key: str, default: bool = False, context: dict | None = None) -> bool:
         if context is None:
             context = {}
         context["environment"] = running_settings.running_environment
         context["zone"] = nuclia_settings.nuclia_zone
-        return self.flag_service.enabled(flag_key, default=default, context=context)
+
+        enabled = False
+        if self.flipt_enabled:
+            if flag_key in const._FliptFeatures:
```
javitonino: I would:
- Collapse these two ifs into one.
- Return directly the results of the evaluation.

Something like:

```python
if self.flipt_enabled and flag_key in const._FliptFeature:
  return evaluate_boolean(...)
else:
  return self.flag_service_enabled(...)
```

Currently, if a feature returns false from flipt, it's evaluated with mr.flagly which is probably not what we want.

With my suggestion, we only fallback to mrflagly if flipt is disabled, which is probably what we want (I think?).

javitonino (approved):

jotare:
[sc-14317]

PR #3589, opened 2026-04-10
Fix nidx binding build

javitonino:
Switching to python 3.10 (from 3.9 default). Latest manylinux image removed 3.9 since it's EOL.


lferran (approved):
tyvm!

PR #3592, opened 2026-04-10
Support OTLP exporter in nucliadb_telemetry

javitonino:
If OTLP_COLLECTOR_ENDPOINT is set, use the OTLP exporter. If not, and JAEGER_* are set, use the old code for Jaeger exporter.


nucliadb_telemetry/src/nucliadb_telemetry/utils.py
```diff
@@ -46,67 +50,80 @@ def get_telemetry(service_name: str | None = None) -> AsyncTracerProvider | None
...
+    if not telemetry_settings.tracing_enabled():
         return None
 
-    tracer_provider = AsyncTracerProvider(
-        active_span_processor=AsyncMultiSpanProcessor(),  # type: ignore
-        resource=Resource.create({SERVICE_NAME: service_name}),
-    )
+    resource = Resource.create({SERVICE_NAME: service_name})
+    if telemetry_settings.otlp_collector_endpoint is not None:
+        tracer_provider = TracerProvider(resource=resource)
```
jotare, line 59: Why using a sync tracer provider here instead of the async?

javitonino: Because there is no async provider in the SDK. The async provider is something we have in nucliadb_telemetry but not the official sdk. From previous research, it looks like we copied and adapted some code from the SDK into nucliadb_telemetry to make it work better with asyncio.

3 years later, this is no longer needed, we can directly use the SDK itself. Internally, the SDK always uses sync code but runs it in a separate thread (exporting a span just puts it in a queue, the thread will eventually pick it up, build a batch, and send it). 

Our implementation uses an asyncio task for this, which is probably a better fit. But you know, I don't think it justifies using something custom instead of the sdk standard that everyone else uses.

jotare (approved):

PR #3595, opened 2026-04-15
Store extracted text in nidx (2)

jotare:
### Description
Parsing rids as UUIDs and serializing them returns them with `-`, while the Python side compares them as strings without `- `.
Also disable conversation fields from nidx as we don't store them properly (yet)

### How was this PR tested?
Now we enable flipt FF for all tests, so all tests are now using nidx as extracted text storage


lferran (approved):

jotare:
[sc-14317]

PR #3596, opened 2026-04-15
Fix and explain nidx_paragraph indexing regex

jotare:
### Description
Fix an undocumented regex from the beginning of the times that now was in the wrong place. This regex searches for explicitly written escape sequences (with multiple chars) and replaces them for a whitespace. This improves search in case the extracted text contains this sequences, as tantivy tokenizer doesn't split it.

However, the regex was replacing 2 chars by 1, which made every match on the regex shift the indexed content compared with the actual paragraph.

This affects ALL keyword search on documents matching the regex. It makes keyword search worse and, depending on the skew, return wrong paragraphs (which matched a different text)

### How was this PR tested?
Locally with a huge document containing several occurrences of this.

### DIFF
--- a/nidx/nidx_paragraph/src/resource_indexer.rs
+++ b/nidx/nidx_paragraph/src/resource_indexer.rs
@@ -30,7 +30,9 @@ use tantivy::doc;
 use tantivy::schema::Facet;
 
 lazy_static::lazy_static! {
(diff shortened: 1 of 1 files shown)


lferran (approved):
lgtm. you rock! should we do a reindex?

javitonino (approved):

jotare:
[sc-14341]

PR #3597, opened 2026-04-16
Optimizations for nidx extracted text API

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nidx/nidx_text/src/reader.rs
+++ b/nidx/nidx_text/src/reader.rs
@@ -510,7 +510,7 @@ impl TextReaderService {
     pub fn get_paragraphs_text(
         &self,
         paragraph_uids: Vec<ParagraphUid>,
-    ) -> anyhow::Result<HashMap<ParagraphUid, Option<String>>> {
+    ) -> anyhow::Result<HashMap<ParagraphUid, String>> {
         let mut field_paragraph_ids = HashMap::new();
         for paragraph_id in paragraph_uids {
             let field_id = FieldUid::from(paragraph_id.clone());
@@ -600,7 +600,7 @@ impl TextReaderService {
     fn extract_paragraphs(
         ids: impl Iterator<Item = ParagraphUid>,
         mut text: std::str::Chars<'_>,
-    ) -> HashMap<ParagraphUid, Option<String>> {
+    ) -> HashMap<ParagraphUid, String> {
         let mut paragraphs = HashMap::new();
 
         // sort paragraph_ids by (start, end) to avoid the need of already read chars from the text
@@ -636,7 +636,7 @@ impl TextReaderService {
                     // clamp to chunk size, we don't have more text
                     let end = std::cmp::min((id.paragraph_end - window.start) as usize, chunk.len());
                     let paragraph: String = chunk[start..end].iter().collect();
-                    paragraphs.insert(id, Some(paragraph));
+                    paragraphs.insert(id, paragraph);
                 }
 
                 // As the new paragraph could overlap with future ones, we reset the window with it
@@ -657,7 +657,7 @@ impl TextReaderService {
             let start = (id.paragraph_start - window.start) as usize;
             let end = std::cmp::min((id.paragraph_end - window.start) as usize, chunk.len());
             let paragraph: String = chunk[start..end].iter().collect();
-            paragraphs.insert(id, Some(paragraph));
+            paragraphs.insert(id, paragraph);
         }
 
         paragraphs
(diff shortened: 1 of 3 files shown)


nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +525,92 @@ impl TextReaderService {
...
+                split: parts.get(3).map(|x| x.to_string()),
+            };
+
+            if let Some(paragraphs) = field_paragraph_ids.remove(&field_uid) {
+                // iterate the text by unicode characters only once, reusing the same iterator for
+                // all paragraphs on the field. This is more useful for multiple paragraphs per
+                // field on a large text
+                let mut paragraph_chars = text.chars();
+                let mut skip = 0;
+                for paragraph_id in paragraphs.into_iter().sorted() {
```
javitonino: I think the sort can be more efficient by checking just paragraph_start (maybe sort_by_key(paragraph_start)? Probably minimal impact since this is a short array.

This also requires that paragraphs are not overlapping, so we may want to check this here and return an error if it's not the case.

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +525,92 @@ impl TextReaderService {
...
+                    skip = paragraph_id.paragraph_start as usize - skip;
```
javitonino: This doesn't look correct. For the paragraphs: 10-20, 40-50, this will:

1st
skip = 10
take = 10
paragraph_text = text[10:20]
iterator_position = 20

2nd
skip = 30 (par_start-skip = 40-10)
take = 10
paragraph_text = text[50:60] (starts at position+skip), incorrect
iterator_position = 60

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +525,92 @@ impl TextReaderService {
...
+                    self.schema.encoded_field_id_bytes,
+                    &encode_field_id_bytes(
+                        Uuid::parse_str(&field_uid.rid)?,
+                        &format!("{}/{}", field_uid.field_type, field_uid.field_name),
+                    ),
+                ),
+                IndexRecordOption::Basic,
+            )));
+        }
+        let query: Box<dyn Query> = Box::new(BooleanQuery::union(subqueries));
```
javitonino: Maybe the query building above can be extracted and shared with get_fields_text?

javitonino (changes requested):

nidx/nidx_text/src/reader.rs
jotare: Ups, I missed to update skip afterwards

nidx/nidx_text/src/reader.rs
jotare: I think we should handle overlapping paragraphs. Although not necessary by /find, /augment can need multiple overlapping paragraphs. Not sure if this should be handled here or above though...

jotare: I just implemented a method to get overlapping paragraphs. The use case for overlapping paragraphs is an augment of nearby paragraphs. For example, a neighboring paragraphs strategy for paragraphs close enough so the augmented text overlaps.

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +506,162 @@ impl TextReaderService {
...
+        let Some(first) = ids.next() else {
+            return paragraphs;
+        };
+        let mut window = std::ops::Range {
+            start: first.paragraph_start,
+            end: first.paragraph_end,
+        };
+        let mut window_paragraphs = vec![first];
+
+        let mut skip = 0;
```
javitonino: I think this should be named something like `position`. It tracks the character index the iterator is currently at.

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +506,162 @@ impl TextReaderService {
...
+        }
+        // we store a doc per field, so we expect at most the number of unique fields
+        let limit = subqueries.len();
+        let query: Box<dyn Query> = Box::new(BooleanQuery::union(subqueries));
+        let collector = TopDocs::with_limit(limit).order_by_score();
+        let results = searcher.search(&query, &collector)?;
+        Ok(results)
+    }
+
+    fn extract_paragraphs(
```
javitonino, line 604: Very cool!

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +506,162 @@ impl TextReaderService {
...
+        ids: impl Iterator<Item = ParagraphUid>,
+        mut text: std::str::Chars<'_>,
+    ) -> HashMap<ParagraphUid, Option<String>> {
```
javitonino: I think this can never return None as a value of the HashMap. Do we need an Option here? (and in the caller)

nidx/nidx_text/src/reader.rs
```diff
@@ -525,6 +506,162 @@ impl TextReaderService {
...
+                        &format!("{}/{}", field_uid.field_type, field_uid.field_name),
+                    ),
+                ),
+                IndexRecordOption::Basic,
+            )));
+        }
+        // we store a doc per field, so we expect at most the number of unique fields
+        let limit = subqueries.len();
+        let query: Box<dyn Query> = Box::new(BooleanQuery::union(subqueries));
+        let collector = TopDocs::with_limit(limit).order_by_score();
```
javitonino, line 599: limit=0 panics tantivy. Probably not an issue since we control how this is called on the python side. But maybe take a look to ensure this cannot happen.

javitonino (approved):

nidx/nidx_text/src/reader.rs
jotare: Oh true, I forgot to do the rename

nidx/nidx_text/src/reader.rs
jotare: I'll do a check to make sure

jotare:
[sc-14317]

PR #3602, opened 2026-04-22
Use nidx as extracted text storage in augmentor

jotare:
### Description
- Use nidx as extracted text storage in augmentor
- Implements an augmentor cache that groups requests to nidx in order to optimize the amount of requests done
- Conversation fields are still out of scope for nidx. In this case (and any problem with nidx), we fallback to object storage

### How was this PR tested?
Existing test suite with nidx as extracted text storage enabled

### DIFF
--- a/nucliadb/src/nucliadb/common/cache.py
+++ b/nucliadb/src/nucliadb/common/cache.py
@@ -28,13 +28,10 @@ from typing import Generic, TypeVar
 
 import backoff
 from async_lru import _LRUCacheWrapper, alru_cache
-from nidx_protos.nidx_pb2 import ExtractedTextsRequest
 from typing_extensions import ParamSpec
 
-from nucliadb.common import datamanagers
 from nucliadb.common.ids import FieldId
 from nucliadb.common.maindb.utils import get_driver
-from nucliadb.common.nidx import get_nidx_searcher_client
 from nucliadb.ingest.fields.base import FieldTypes
 from nucliadb.ingest.orm.knowledgebox import KnowledgeBox as KnowledgeBoxORM
 from nucliadb.ingest.orm.resource import Resource as ResourceORM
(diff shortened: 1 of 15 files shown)


nucliadb/src/nucliadb/search/augmentor/extracted_text.py
```diff
@@ -0,0 +1,144 @@
...
+        resource_nidx_shard = {}
+
+        for field_id in fields:
+            rid = field_id.rid
+            if rid not in resource_nidx_shard:
+                resource_shard_id = await datamanagers.resources.get_resource_shard_id(
+                    txn, kbid=kbid, rid=rid
+                )
+                if resource_shard_id is None:
+                    return None
```
lferran: should we log and continue instead?
It could be that a single resource was deleted and still being returned in search results, and this will return all the function

nucliadb/src/nucliadb/search/augmentor/extracted_text.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+
+        for paragraph_id in paragraphs:
+            rid = paragraph_id.rid
+            if rid not in resource_nidx_shard:
+                resource_shard_id = await datamanagers.resources.get_resource_shard_id(
+                    txn, kbid=kbid, rid=rid
+                )
+                if resource_shard_id is None:
+                    return None
```
lferran: same

nucliadb/src/nucliadb/search/search/paragraphs.py
```diff
@@ -71,70 +44,12 @@ async def get_paragraph_from_full_text(
...
-            if log_on_missing_field:
-                logger.warning(
-                    "Extracted_text for field does not exist on DB. This should not happen.",
-                    extra={
-                        "field_id": field.resource_unique_id,
-                        "kbid": field.kbid,
-                    },
-                )
-            return ""
+    from nucliadb.search.augmentor.paragraphs import get_paragraph_text
```
lferran: normally the imports inside functions are indicators of a bad module/package structure. Can you check if there is another way to fix this? (same above)

jotare: Uuh, thanks, it was a leftover while refactoring that stuff

nucliadb/src/nucliadb/common/cache.py
```diff
@@ -120,69 +116,23 @@ def __init__(self, cache_size: int) -> None:
...
-                        ],
-                    )
+            storage = await get_storage()
+            try:
+                sf = storage.file_extracted(
+                    kbid, field_id.rid, field_id.type, field_id.key, FieldTypes.FIELD_TEXT.value
+                )
+                return await storage.download_pb(sf, ExtractedText)
+            except Exception:
+                logger.warning(
```
lferran, line 126: is this try/catch + warning really needed, if we are raising the exception anyway?

jotare: I guess it's to log the warning, it was already like this before

lferran (approved):

jotare:
[sc-14317]

PR #3613, opened 2026-04-28
File duplication checks

lferran:
### Description
These are all the aggregated changes of a series of PRs that I will push.

The plan is to enable the flags it in 3 steps

- 1st:
  - Table creation migration
  - Writes and deletes to md5 table at ingest time active. (activate WRITE flag)

- 2nd:
  - Backfill migration

- 3rd:
  - Check for file duplication using the md5 table (activate READ flag)
  - Do not reuse uuids on resource creation, anywhere.

[sc-14410]

### How was this PR tested?
integration tests


nucliadb/src/nucliadb/reader/api/v1/file_check.py
```diff
@@ -0,0 +1,49 @@
...
+from fastapi_versioning import version
+
+from nucliadb.common import file_md5
+from nucliadb.reader.api.v1.router import KB_PREFIX, api
+from nucliadb_models.resource import NucliaDBRoles
+from nucliadb_utils.authentication import requires_one
+
+
+@api.head(
+    f"/{KB_PREFIX}/{{kbid}}/file-exists/{{md5}}",
```
jotare: Why are we exposing an endpoint?

nucliadb_sdk/src/nucliadb_sdk/v2/sdk.py
```diff
@@ -480,6 +480,12 @@ class SdkEndpointDefinition:
         method="POST",
         path_params=("kbid",),
     ),
+    # File deduplication check
+    "file_exists": SdkEndpointDefinition(
```
jotare: And an SDK function?

nucliadb/src/nucliadb/common/file_md5.py
```diff
@@ -0,0 +1,86 @@
...
+from typing import cast
+
+from nucliadb.common.maindb.driver import Transaction
+from nucliadb.common.maindb.pg import PGDriver, PGTransaction
+from nucliadb.common.maindb.utils import get_driver
+from nucliadb_telemetry import metrics
+
+logger = logging.getLogger(__name__)
+
+observer = metrics.Observer("file_md5", labels={"op": ""})
```
jotare: Metrics should be prefixed by `nucliadb_

nucliadb/src/migrations/0042_backfill_file_md5.py
```diff
@@ -0,0 +1,104 @@
...
+                    )
+                    if payload is None or len(payload) == 0:
+                        # No payload found for this field, skip it
+                        continue
+                    value = FieldFile()
+                    value.ParseFromString(payload)
+                    if not value.file.md5:
+                        # No md5 hash set to backfill for this field
+                        continue
+                    await file_md5.set(txn, kbid=kbid, md5=value.file.md5, rid=rid, field_id=field_id)
```
jotare, line 90: If at some point we want to extend this to conversation attachment dedup or other use cases, it'd be better to use the field type as part of the field id too

nucliadb/src/nucliadb/writer/api/v1/upload.py
```diff
@@ -860,23 +863,18 @@ async def validate_field_upload(
...
         # we are going to create a new resource and a field
         if md5 is not None:
-            exists = await datamanagers.atomic.resources.resource_exists(kbid=kbid, rid=md5)
+            exists = await file_md5.exists(kbid=kbid, md5=md5)
             if exists:
-                raise HTTPConflict("A resource with the same uploaded file already exists")
-            rid = md5
-        else:
-            rid = uuid.uuid4().hex
+                raise HTTPConflict("A file with the same MD5 hash already exists in this knowledge box")
```
jotare: Do we have to explain the details of our dedup internals?

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -452,6 +455,25 @@ async def apply_fields(self, message: BrokerMessage):
...
             self.modified = True
 
+    async def set_file_field_md5(self, field_id: str, file: CloudFile):
+        """
+        Record file MD5 hashes for deduplication checks.
+        """
+        if not file.md5:
+            # To be safe, delete any existing MD5 record for this field if the new file doesn't have an MD5.
+            # This is for PUT operations where the file is being replaced but the new md5 is not provided.
+            await self.delete_file_field_md5(field_id)
```
jotare, line 470: In this case then we don't store the MD5? Shouldn't we compute it?

lferran: I don't want to do it right now, as this would probably involve downloading the file in memory.

The idea with this PR is to replace the existing feature with a better implementation. The existing feature is: if the user passes a x-md5 header, we check if the file already exists in the KB and return conflict if so.

nucliadb/src/nucliadb/reader/api/v1/file_check.py
lferran: I think it can be useful. Right now, we only check for file duplicates in the api/v1/kb/kbid/tusupload endpoint.

Serving this in an endpoint allows the api users to do this check on their end easily without us having to implement the check in all our endpoints.

But maybe we should only expose this in an endpoint if we make sure to calculate md5 hashes always for all files. I'm ok removing it for now, in fact.

nucliadb_sdk/src/nucliadb_sdk/v2/sdk.py
lferran: it was implemented below, but I will remove it for now

nucliadb/src/migrations/0042_backfill_file_md5.py
lferran: field_id is already an arbitrary string, so we can put any prefix we want in it. 
Note that I am using it with the "f/" prefix

nucliadb/tests/nucliadb/integration/common/test_file_md5.py
```diff
@@ -148,7 +148,7 @@ async def test_delete_by_kb(driver: PGDriver):
     assert await file_md5.exists(kbid=other_kb, md5="ccc") is True
 
 
-@pytest.mark.skip(reason="Only for local testing")
+#@pytest.mark.skip(reason="Only for local testing. Uncomment line to run")
```
jotare: AI slop? If you comment the skip it will actually run and uncomment will do a skip and not run :)

lferran: nah, sorry. this was a dirty local change that I pushed accidentally

nucliadb/src/migrations/0042_backfill_file_md5.py
jotare: Is it using the `f/` prefix? If the split is done on `/`, `field_id` is only the field name, not with the type prefix. Am I missing something?

nucliadb/tests/nucliadb/integration/common/test_file_md5.py
```diff
@@ -0,0 +1,287 @@
...
+        yield pg_maindb_driver
+
+
+def kbid() -> str:
+    return str(uuid.uuid4())
+
+
+def rid() -> str:
+    return str(uuid.uuid4())
+
```
jotare, line 45: Do you need a function for this?

nucliadb_utils/src/nucliadb_utils/featureflagging.py
```diff
@@ -65,6 +65,14 @@ class Settings(pydantic_settings.BaseSettings):
...
         "variants": {"environment": ["local"]},
     },
+    const.Features.FILE_MD5_WRITES: {
+        "rollout": 0,
+        "variants": {"environment": ["local"]},
+    },
+    const.Features.FILE_MD5_READS: {
+        "rollout": 0,
+        "variants": {"environment": ["local"]},
+    },
```
jotare, line 75: You can also use Flipt FF if you want to try it

nucliadb/src/migrations/0042_backfill_file_md5.py
lferran: right now this is only used for files here:

https://github.com/nuclia/nucliadb/pull/3613/changes#diff-598885ee452ffda6ee778ab290ca0630c75801b4cf1501c6d85aaddfb3ddb3ffR473

support for conversation field attachments is out of scope of this PR. I think we can implement/extend  it easliy later.

jotare (approved):

PR #3615, opened 2026-04-29
Nidx json

javitonino:
### Description

### How was this PR tested?


nidx/nidx_json/src/schema.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub json: Field,
+}
+
+impl JsonSchema {
+    pub fn new() -> Self {
+        let mut sb = Schema::builder();
+
+        let rid = sb.add_bytes_field("uuid", STORED | INDEXED);
+        let encoded_rid = sb.add_u64_field("encoded_field_id", FAST);
```
jotare: Shouldn't this be called resource?
```suggestion
        let encoded_rid = sb.add_u64_field("encoded_resource_id", FAST);
```

nidx/nidx_json/src/search.rs
```diff
@@ -0,0 +1,385 @@
...
+    pub json_path: String,
+    pub predicate: JsonPredicate,
+}
+
+#[derive(Clone)]
+pub enum JsonPredicate {
+    Text(String),
+
+    IntRange { lower: Option<i64>, upper: Option<i64> },
+    FloatRange { lower: Option<f64>, upper: Option<f64> },
```
jotare, line 39: Should we consider excluding ranges too? Right now is [lower, upper] but we may also want [lower, upper) or other combinations.

nidx/nidx_protos/noderesources.proto
```diff
@@ -10,6 +10,10 @@ message TextInformation {
     repeated string labels = 2;
 }
 
+message JsonInformation {
```
jotare: NIT: JsonValue instead of Information?

nidx/nidx_types/src/prefilter.rs
```diff
@@ -36,3 +40,58 @@ pub enum PrefilterResult {
     All,
     Some(Vec<FieldId>),
 }
+
+impl PrefilterResult {
+    /// Combine `self` (the text/security prefilter result) with a set of
+    /// resource UUIDs from the JSON prefilter using either OR
+    /// (`filter_or = true`) or AND (`filter_or = false`) semantics.
```
jotare: An enum would probably read better for callers and inside the function. Although I know we have plenty of similar enums, so feel free to ignore

nidx/nidx_types/src/prefilter.rs
```diff
@@ -36,3 +40,58 @@ pub enum PrefilterResult {
...
+        match (self, filter_or) {
+            (PrefilterResult::All, true) => PrefilterResult::All,
+            (PrefilterResult::None, true) => PrefilterResult::Some(to_field_ids(&resource_uuids)),
+            (PrefilterResult::Some(fields), true) => {
+                // All resources from the JSON prefilter pass at resource-level.
+                // Fields only in the text prefilter (not covered by resource_uuids)
+                // are kept at their original field-level granularity.
+                let mut merged = to_field_ids(&resource_uuids);
+                merged.extend(fields.into_iter().filter(|f| !resource_uuids.contains(&f.resource_id)));
+                if merged.is_empty() {
```
jotare: This can never be empty, we already checked resource_uuids is not empty at the beginning

nidx/nidx_json/src/search.rs
javitonino: I wanted to keep it simple for now, we can always add more things later.

jotare (approved):

PR #3616, opened 2026-04-29
Nucliadb kv

javitonino:
### Description

### How was this PR tested?


nucliadb_utils/src/nucliadb_utils/featureflagging.py
```diff
@@ -65,6 +65,10 @@ class Settings(pydantic_settings.BaseSettings):
         "rollout": 0,
         "variants": {"environment": ["local"]},
     },
+    const.Features.KEY_VALUE_FIELDS: {
+        "rollout": 0,
+        "variants": {"environment": ["local"]},
+    },
```
jotare: You shouldn't need this, there's a fixture that enables all flipt FFs by default in all tests

nucliadb/src/nucliadb/reader/api/v1/resource.py
```diff
@@ -418,6 +418,15 @@ async def _get_resource_field(
                 value = await field.get_value()
                 resource_field.value = from_proto.field_link(value)
 
+            if isinstance(value, resources_pb2.FieldKeyValue):
+                import json
+                from typing import Any
+
+                resource_field.value = cast(Any, json.loads(value.data) if value.data else {})
```
jotare: Imports should be at the top. Why this cast to Any? We know the type is at least a dict

nucliadb/src/nucliadb/writer/api/v1/services.py
```diff
@@ -231,3 +234,89 @@ async def delete_search_configuration(request: Request, kbid: str, config_name:
...
+    async with datamanagers.with_transaction() as txn:
+        if not await datamanagers.kb.exists_kb(txn, kbid=kbid):
+            raise HTTPException(status_code=404, detail="Knowledge Box does not exist")
+
+        if await datamanagers.kv_schemas.get(txn, kbid=kbid, name=item.name) is not None:
+            raise HTTPException(status_code=409, detail="KV schema already exists")
+
+        schema = KVSchema(name=item.name, description=item.description, fields=item.fields)
+        await datamanagers.kv_schemas.set(txn, kbid=kbid, schema=schema)
+        await txn.commit()
```
jotare: Now that we're adding this feature, we should limit the amount of schemas from the beginning. Otherwise this key can grow as much as users want

nucliadb/src/nucliadb/common/datamanagers/kv_schemas.py
```diff
@@ -0,0 +1,62 @@
...
+async def get_all(txn: Transaction, *, kbid: str) -> KBKVSchemas:
+    key = KB_KV_SCHEMAS.format(kbid=kbid)
+    data = await txn.get(key, for_update=False)
+    if not data:
+        return KBKVSchemas()
+    return KBKVSchemas.model_validate_json(data)
+
+
+async def get(txn: Transaction, *, kbid: str, name: str) -> KVSchema | None:
+    schemas = await get_all(txn, kbid=kbid)
```
jotare, line 35: Storing all schemas in a single key can become a really big value

nucliadb/src/nucliadb/ingest/fields/key_value.py
```diff
@@ -0,0 +1,71 @@
...
+    schema_fields = {f.key: f for f in schema.fields}
+    for key, value in data.items():
+        check_kv_type(schema.name, key, value, schema_fields[key].type)
+
+
+def _validate_keys(data: dict, schema: KVSchema) -> None:
+    schema_keys = {f.key for f in schema.fields}
+    unknown = set(data.keys()) - schema_keys
+    if unknown:
+        raise HTTPException(422, detail=f"Unknown keys for schema {schema.name!r}: {sorted(unknown)}")
```
jotare: We shouldn't raise HTTPExceptions here!

nucliadb/src/nucliadb/ingest/orm/brain_v2.py
```diff
@@ -511,6 +512,13 @@ def delete_field(self, field_key: str):
         self.brain.paragraphs_to_delete.append(full_field_id)
         self.brain.vectors_to_delete_in_all_vectorsets.append(full_field_id)
         self.brain.relation_fields_to_delete.append(field_key)
+        if ftype == "k":
+            self.brain.json_fields_to_delete.append(full_field_id)
+
+    def generate_json(self, field_id: str, kv_data: dict) -> None:
+        from nidx_protos.noderesources_pb2 import JsonInformation
```
jotare: Import inside function

nucliadb_models/src/nucliadb_models/kv_schemas.py
```diff
@@ -0,0 +1,82 @@
...
+    key: str = Field(pattern=r"^[^/.]{1,64}$")
+    type: KVFieldType
+    description: str = ""
+    required: bool = True
+
+
+class KVSchema(BaseModel):
+    name: str = Field(pattern=r"^[^/.]{1,64}$")
+    description: str = ""
+    fields: list[KVSchemaField] = Field(default_factory=list)
```
jotare: We should limit here the number of fields too

nucliadb/src/nucliadb/writer/resource/field.py
```diff
@@ -230,6 +231,69 @@ async def parse_fields(
...
+    a 422 to the HTTP client before the BrokerMessage is committed.
+    The entire data dict is JSON-encoded into the proto's single string data field.
+    No NLP processing is needed for KV fields.
+    """
+    import json
+
+    from fastapi import HTTPException
+
+    from nucliadb.ingest.fields.key_value import validate_kv_data
+
```
jotare: More imports inside functions...

nucliadb/src/nucliadb/ingest/serialize.py
```diff
@@ -299,6 +302,18 @@ async def serialize_resource(
...
+                    resource.data.key_values = {}
+                if field.id not in resource.data.key_values:
+                    resource.data.key_values[field.id] = KeyValueFieldData()
+                if include_value and value is not None:
+                    # data is a single JSON string containing the entire KV payload
+                    resource.data.key_values[field.id].value = (
+                        json.loads(value.data) if value.data else {}
+                    )
+                if include_errors:
+                    await serialize_field_errors(field, resource.data.key_values[field.id])
```
jotare, line 316: If we're adding support for key-value fields here, we may want to support them in the /augment API too. Not necessarily in this PR though

nucliadb_models/src/nucliadb_models/filters.py
```diff
@@ -303,6 +303,75 @@ class Status(FilterProp, extra="forbid"):
...
+    """Matches a key-value field where the given key falls within a numeric range"""
+
+    op: Literal["range"] = "range"
+    field_id: str = pydantic.Field(description="The KV field/schema name, e.g. 'product'")
+    key: str = pydantic.Field(description="The key within the KV data, e.g. 'price'")
+    gte: float | int | None = pydantic.Field(default=None, description="Greater than or equal to")
+    lte: float | int | None = pydantic.Field(default=None, description="Less than or equal to")
+
+    @model_validator(mode="after")
+    def check_bounds(self) -> "KVRange":
```
jotare, line 328: We should also check that the bounds limit something, i.e., lte <= gte

nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -124,6 +130,128 @@ async def parse_expression(
...
+        return nodereader_pb2.JsonFilterExpression(path=path)
+    elif isinstance(expr, KVBoolMatch):
+        path = nodereader_pb2.JsonFieldPathFilter(
+            field_id=f"k/{expr.field_id}",
+            json_path=expr.key,
+        )
+        path.boolean = expr.value
+        return nodereader_pb2.JsonFilterExpression(path=path)
+    else:
+        raise ValueError(f"Unknown KVFilterExpression type: {type(expr)}")
```
jotare: Can't we use an `assert_never` here?

nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -124,6 +130,128 @@ async def parse_expression(
...
+            )
+        elif isinstance(expr, KVBoolMatch) and schema_field.type != "boolean":
+            raise InvalidQueryError(
+                "key_value",
+                f"Key '{expr.key}' in schema '{expr.field_id}' is of type '{schema_field.type}', "
+                f"but 'bool_match' requires type 'boolean'",
+            )
+        return parse_kv_filter_expression(expr)
+    else:
+        raise ValueError(f"Unknown KVFilterExpression type: {type(expr)}")
```
jotare: Same (assert_never)

nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -124,6 +130,128 @@ async def parse_expression(
...
+        )
+        path.boolean = expr.value
+        return nodereader_pb2.JsonFilterExpression(path=path)
+    else:
+        raise ValueError(f"Unknown KVFilterExpression type: {type(expr)}")
+
+
+def _parse_kv_filter_expression(
+    expr: KVFilterExpression,
+    all_schemas: "KBKVSchemas",
```
jotare: No need of string type as it's imported
```suggestion
    all_schemas: KBKVSchemas,
```

nucliadb/tests/nucliadb/integration/test_key_value_fields.py
```diff
@@ -0,0 +1,478 @@
...
+                    },
+                }
+            },
+        },
+    )
+    assert resp.status_code == 201, resp.text
+    rid2 = resp.json()["uuid"]
+
+    # Indexing is awaited by the standalone_knowledgebox / nucliadb_reader fixtures
+    await asyncio.sleep(1)
```
jotare: Why not using `wait_for_sync`?

nucliadb/src/nucliadb/writer/api/v1/services.py
```diff
@@ -231,3 +234,95 @@ async def delete_search_configuration(request: Request, kbid: str, config_name:
...
+        if item.name in existing.schemas:
+            raise HTTPException(status_code=409, detail="KV schema already exists")
+
+        await datamanagers.kv_schemas.set(txn, kbid=kbid, schema=item)
+        await txn.commit()
+
+    return item
+
+
+@api.put(
```
jotare, line 272: Just as a comment, updates are a tricky operation. One can update their schema but indexed data will remain the same. If a type change for the same key is done, the index can end up with a field with different types on the same attribute

jotare (approved):

nucliadb/src/nucliadb/ingest/serialize.py
javitonino: yeah, another PR for how we use this. I just wrote some basics to be able to test this, but we have to figure out exactly the APIs where this can be used

nucliadb/src/nucliadb/writer/api/v1/services.py
javitonino: Yes. But it kind of works. There should be some tests in nidx_json. tantivy basically creates an index per column and type, so you can only get results of the type in the query. It's not pretty and will need documentation, but I think we can support this with some caveats.

nucliadb/tests/nucliadb/integration/test_key_value_fields.py
javitonino: debugging stuff, It should not even be needed, since there is an implicit wait_for in the search

PR #3621, opened 2026-05-01
Fix link extracted title indexing

jotare:
### Description
LinkExtractedData can come with a better title than a URL for a link field. This was updated on the db but not properly propagated to the index. In addition, a bug in the apply vectors iterations has been fixed too.
[sc-6088]

### How was this PR tested?
Improved the test that in the day was used to validate this fix


jotare:
[sc-6088]

carlesonielfa (approved):

PR #3627, opened 2026-05-06
Add RAO to reporting 

Ciniesta:
Add RAO to reporting services


VERSION
```diff
@@ -1 +1 @@
-6.13.0
\ No newline at end of file
+6.13.1
```
javitonino: Please revert this change. This is technically not needed in nucliadb_protos (we publish all protos with a suffix, so you can always refer to that 6.13.0.postXXXX).

If you really want to bump, we use this action: https://github.com/nuclia/nucliadb/actions/workflows/bump-version.yml that bumps all nucliadb components, docs, public docker images, so we try to be a bit careful with it.

javitonino (changes requested):

VERSION
Ciniesta: I did it out of habit since that's what we do with our protos, thanks for the warning

Ciniesta: done

jotare (approved):

javitonino (approved):

PR #3638, opened 2026-05-11
Allow to define the UUID on local on prem NucliaDB

bloodbare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb_models/src/nucliadb_models/resource.py
+++ b/nucliadb_models/src/nucliadb_models/resource.py
@@ -78,6 +78,11 @@ class ExtractedDataTypeName(str, Enum):
 
 
 class KnowledgeBoxConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


nucliadb_models/src/nucliadb_models/resource.py
```diff
@@ -78,6 +78,11 @@ class ExtractedDataTypeName(str, Enum):
 
 
 class KnowledgeBoxConfig(BaseModel):
+    uuid: str | None = Field(
```
jotare: Could you use `uuid.UUID` type instead of `str` so we have UUID validation? You can then do `str(item.uuid)` to convert it to string inside the enpdoint

bloodbare: Yups no prob

jotare (approved):

PR #3656, opened 2026-05-15
ty in nucliadb

javitonino:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
```diff
@@ -196,7 +197,7 @@ async def get_node_index_counts(kbid: str) -> tuple[IndexCounts, list[str]]:
 
     counts = IndexCounts(fields=0, paragraphs=0, sentences=0, size_bytes=0)
     for shard in results:
-        if isinstance(shard, Exception):
+        if isinstance(shard, BaseException):
```
lferran, line 200: why BaseException?

lferran (approved):
lgtm ! thanks for all the cleanup work

nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
javitonino: Because it's the top-level exception https://docs.python.org/3/library/exceptions.html#exception-hierarchy and at this point, you don't know that it's not one of the other non-exception types.

In the old case, if a non-Exception error was raised (which I'm not even sure it can practically happen), it would try to parse it as a result and fail uncontrolled.

PR #3657, opened 2026-05-15
Add endpoint to delete messages in a conversation

lferran:
### Description

[sc-14709]

Important considerations:
- Processed data (vectors, metadata, etc) of deleted splits is not really deleted for now.
- Conversation pages are not consolidated, we just remove their text.
- For simplicify, when there are message deletions on a conversation field, the whole field is reindexed.
- Added a migration to populate the index to store page for every split, for easier deletion.

### How was this PR tested?
Integration tests.

### DIFF
--- a/nucliadb/src/nucliadb/ingest/fields/conversation.py
+++ b/nucliadb/src/nucliadb/ingest/fields/conversation.py
@@ -267,46 +267,24 @@ class Conversation(Field[PBConversation]):
         metadata = await self.get_metadata()
         splits_metadata = await self.get_splits_metadata()
         idents_to_delete = set(message_idents) - set(splits_metadata.deleted_splits)
+        idents_to_delete = idents_to_delete & set(splits_metadata.metadata.keys())
 
-        group_by_page: dict[int, set[str]] = {}
-        idents_without_page = set()
+        page_idents: dict[int, set[str]] = {}
         for ident in idents_to_delete:
             split_meta = splits_metadata.metadata.get(ident)
             if split_meta is not None:
-                group_by_page.setdefault(split_meta.page, set()).add(ident)
-            else:
-                idents_without_page.add(ident)
+                page_idents.setdefault(split_meta.page, set()).add(ident)
 
-        for page, idents in group_by_page.items():
+        for page, idents in page_idents.items():
             some_deleted_in_page = False
             page_obj = await self.db_get_value(page)
             for message in page_obj.messages:
-                if message.ident in idents or message.ident in idents_without_page:
+                if message.ident in idents:
                     message.content.text = ""
                     total_deleted += 1
                     some_deleted_in_page = True
                     splits_metadata.deleted_splits.append(message.ident)
                     idents_to_delete.remove(message.ident)
-
-            if some_deleted_in_page:
-                # If we deleted a message in this page, we need to update it in the database.
-                await self.db_set_value(page_obj, page)
-
-        # The remaining messages may not have the page metadata yet, so we need to delete by iterating all the pages.
-        for page in range(1, metadata.pages + 1):
-            if page in group_by_page:
-                # We already visited this page, so we know the message is not in it.
-                continue
-            some_deleted_in_page = False
-            page_obj = await self.db_get_value(page)
-            for message in page_obj.messages:
-                if message.ident in idents_without_page:
-                    message.content.text = ""
-                    total_deleted += 1
-                    some_deleted_in_page = True
-                    splits_metadata.deleted_splits.append(message.ident)
-                    idents_to_delete.remove(message.ident)
-
             if some_deleted_in_page:
                 # If we deleted a message in this page, we need to update it in the database.
                 await self.db_set_value(page_obj, page)
--- a/nucliadb/src/nucliadb/ingest/fields/conversation.py
+++ b/nucliadb/src/nucliadb/ingest/fields/conversation.py
@@ -18,8 +18,9 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 import uuid
-from typing import Any
+from typing import Any, Iterable
 
+from nucliadb.ingest import logger
 from nucliadb.ingest.fields.base import Field
 from nucliadb_protos.resources_pb2 import CloudFile, FieldConversation, SplitsMetadata
 from nucliadb_protos.resources_pb2 import Conversation as PBConversation
@@ -115,7 +116,14 @@ class Conversation(Field[PBConversation]):
         while True:
             # Fit the messages in the last page
             available_space = metadata.size - len(last_page.messages)
-            last_page.messages.extend(messages[:available_space])
+            page_messages = messages[:available_space]
+
+            # Add the page to the split metadata, so we can track in which page each message is for efficient deletion later.
+            for message in page_messages:
+                split_meta = self._splits_metadata.metadata.get_or_create(message.ident)
+                split_meta.page = metadata.pages
+
+            last_page.messages.extend(page_messages)
 
             # Save the last page
             await self.db_set_value(last_page, metadata.pages)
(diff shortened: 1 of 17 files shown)


nucliadb/tests/nucliadb/integration/test_conversation.py
```diff
@@ -966,3 +991,238 @@ async def test_append_messages_to_non_existent_conversation_field_creates_it(
...
+    kbid = standalone_knowledgebox
+
+    msg1_text = "Hello from the user -- common text"
+    msg2_text = "Hi there from the assistant -- common text"
+    msg3_text = "How are you doing today -- common text"
+    msg1_vector = [-1.0] * 512
+    msg2_vector = [0.0] * 512
+    msg3_vector = [1.0] * 512
+
+    # Create a resource with a conversation field containing 3 messages
```
jotare: Could you reuse the lambs test resource instead of creating a new one? The test will look cleaner and the lambs resource is a more complete conversation

nucliadb/src/nucliadb/ingest/fields/conversation.py
```diff
@@ -243,3 +251,56 @@ async def set_splits_metadata(self, payload: SplitsMetadata) -> None:
...
+
+        for page, idents in page_idents.items():
+            some_deleted_in_page = False
+            page_obj = await self.db_get_value(page)
+            for message in page_obj.messages:
+                if message.ident in idents:
+                    message.content.text = ""
+                    total_deleted += 1
+                    some_deleted_in_page = True
+                    splits_metadata.deleted_splits.append(message.ident)
```
jotare, line 286: We only append stuff to deleted_splits but never remove. If a user deletes a message, that identifier will be inconsistent. The message will be created but we'll skip indexing and a bunch of stuff. I'm not sure if that's the intention

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -362,6 +362,14 @@ async def delete_field(self, type: FieldType.ValueType, key: str):
                 self.all_fields_keys.remove(field)
         await field_obj.delete()
 
+    async def delete_splits(self, payload: writer_pb2.DeleteSplits) -> None:
```
jotare: Do we really want the "split" concept in a function of a resource?

nucliadb/src/nucliadb/reader/api/v1/resource.py
```diff
@@ -437,7 +437,15 @@ async def _get_resource_field(
...
                 if value is not None:
+                    splits_metadata = await field.get_splits_metadata()
+                    deleted_messages = set(splits_metadata.deleted_splits)
                     resource_field.value = from_proto.conversation(value)
+                    # Skip deleted messages
+                    resource_field.value.messages = [
+                        msg
+                        for msg in (resource_field.value.messages or [])
+                        if msg.ident not in deleted_messages
+                    ]
```
jotare, line 448: We're missing the integration with the augmentor. We can do the optimization to use the new splitmetadata in a following PR, but conversation value serialization should be fixed as you've done with here

nucliadb/src/nucliadb/ingest/fields/conversation.py
lferran: Yes, this is the intention for now. 

What we discussed with Javi is that, if we later see this becoming a problem, we can have a background task consolidating conversations that have more than X deletions.

The idea was that, to keep things simple, if the user deletes a message with an ident X, that ident can no longer be reused (unless the consolidation happens)

nucliadb/src/nucliadb/ingest/orm/resource.py
lferran: It's here only as a helper function of resource.apply_fields. Are you ok if I make it a private method with the '_' prefix? Renaming it to `_apply_field_split_deletions` for instance

nucliadb/src/nucliadb/reader/api/v1/resource.py
lferran: Sure! Good point. I'll add it in this PR too. There is no rush though, no one is deleting messages in prod yet unless we publish the docs

nucliadb/tests/nucliadb/integration/test_conversation.py
lferran: 👍🏼  I will try

nucliadb/src/nucliadb/reader/api/v1/resource.py
lferran: done!

nucliadb/tests/nucliadb/integration/test_conversation.py
lferran: done! thanks!

nucliadb/src/nucliadb/ingest/fields/conversation.py
```diff
@@ -243,3 +251,56 @@ async def set_splits_metadata(self, payload: SplitsMetadata) -> None:
...
+
+    async def delete_messages(self, message_idents: Iterable[str]) -> int:
+        """Delete messages by their ident from conversation pages.
+
+        Messages are removed in-place from their pages without restructuring
+        pages, so that page numbering stays stable.
+
+        Returns the number of messages actually deleted.
+
+        Processed metadata for the deleted messages (e.g: text, vectors, etc) is not deleted immediately.
```
jotare: Careful with misleading comments, we're not deleting them at all as far as I've seen, right?

nucliadb/src/nucliadb/ingest/orm/index_message.py
```diff
@@ -445,11 +488,13 @@ async def get_bm_modified_split_ids(
...
     resource: Resource,
-) -> set[str]:
+) -> tuple[set[str], set[str]]:
     fid = conversation_field_id
     conv: Conversation = await resource.get_field(fid.field, fid.field_type, load=False)
     splits_metadata = await conv.get_splits_metadata()
-    return set(splits_metadata.metadata)
+    deleted_splits = set(splits_metadata.deleted_splits)
+    non_deleted_splits = set(splits_metadata.metadata.keys()) - deleted_splits
+    return non_deleted_splits, deleted_splits
```
jotare: NIT
```suggestion
    existing_splits = set(splits_metadata.metadata.keys()) - deleted_splits
    return existing_splits, deleted_splits
```

nucliadb/src/nucliadb/writer/api/v1/field.py
```diff
@@ -686,3 +689,110 @@ async def reprocess_file_field(
         )
 
     return ResourceUpdated(seqid=processing_info.seqid)
+
+
+DELETE_CONVERSATION_MESSAGES_DESCRIPTION = (
+    "Delete a message from a conversation field by its ident. "
+    "Note: deleting too many messages in a single conversation can degrade the performance "
+    "of the conversation field. This endpoint is not meant to be a mutability primitive, "
+    "but is rather a convenience endpoint for occasional deletions."
```
jotare: I think we should add here that the message identifier won't be usable anymore (even if we don't check it). If at some point we clean up the stuff, we can change the description again

nucliadb/src/nucliadb/writer/api/v1/field.py
```diff
@@ -686,3 +689,110 @@ async def reprocess_file_field(
...
+    field_id: FieldIdString,
+    message_ident: MessageIdent,
+) -> Response:
+    rid = await get_rid_from_slug_or_raise_error(kbid, rslug)
+    return await _delete_conversation_messages(kbid, rid, field_id, [message_ident])
+
+
+@api.delete(
+    f"/{KB_PREFIX}/{{kbid}}/{RESOURCE_PREFIX}/{{rid}}/conversation/{{field_id}}/messages/{{message_ident}}",
+    status_code=200,
```
jotare: Most of our endpoints return 204 on DELETE, I think we should use it here too for consistency with the rest of the API

nucliadb/src/nucliadb/writer/api/v1/field.py
lferran: Ok, I'll add the comment. We do check it though!

nucliadb/src/nucliadb/ingest/fields/conversation.py
```diff
@@ -243,3 +251,54 @@ async def set_splits_metadata(self, payload: SplitsMetadata) -> None:
...
+
+        splits_metadata = await self.get_splits_metadata()
+        idents_to_delete = set(message_idents) - set(splits_metadata.deleted_splits)
+        idents_to_delete = idents_to_delete & set(splits_metadata.metadata.keys())
+
+        page_idents: dict[int, set[str]] = {}
+        for ident in idents_to_delete:
+            split_meta = splits_metadata.metadata.get(ident)
+            if split_meta is not None:
+                page_idents.setdefault(split_meta.page, set()).add(ident)
```
javitonino: Just FYI. You can use defaultdict from the collections module which is a bit more ergonomic. Something like:

```
page_idents: dict[int, set[str]] = defaultdict(set)

page_idents[split_meta.page].add(ident)
```

nucliadb/src/nucliadb/ingest/fields/conversation.py
```diff
@@ -243,3 +251,54 @@ async def set_splits_metadata(self, payload: SplitsMetadata) -> None:
...
+
+        for page, idents in page_idents.items():
+            some_deleted_in_page = False
+            page_obj = await self.db_get_value(page)
+            for message in page_obj.messages:
+                if message.ident in idents:
```
javitonino: If you delete an ident that was already deleted, we will add it to deleted_splits again. Maybe we should guard against this? i.e: Check `if message.ident in idents and not message.ident in splits_metadata.deleted_splits`. Or maybe better in the previous loop, so we can skip loading the page.

javitonino (approved):

nucliadb/src/nucliadb/ingest/fields/conversation.py
lferran: I'm already removing the deleted from the idents_to_delete set in the beginning of the function, but I will add this as a safe-guard

javitonino: Ah, I didn't see that :see_no_evil: Then there is no change needed!

PR #3658, opened 2026-05-18
Implement more generic operators for key-value JSON filtering

jotare:
### Description
Replace type-specific operators (bool_match, exact_match) for more generic operators (eq, gte, lte) in key-value JSON filtering. 

### How was this PR tested?
Integration tests

### DIFF
--- a/nucliadb/src/nucliadb/common/filter_expression.py
+++ b/nucliadb/src/nucliadb/common/filter_expression.py
@@ -19,6 +19,7 @@
 #
 
 from datetime import datetime
+from typing import Type
 
 from nidx_protos import nodereader_pb2
 from nidx_protos.nodereader_pb2 import FilterExpression as PBFilterExpression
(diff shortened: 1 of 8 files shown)


nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -131,142 +131,126 @@ async def parse_expression(
...
-        assert_never(expr)
+    """Convert a key-value filter expression tree into it's proto
+    representation, validating types against the corresponding key-value schemas
+    used
 
+    """
+    async with datamanagers.with_ro_transaction() as txn:
+        all_schemas = await datamanagers.kv_schemas.get_all(txn, kbid=kbid)
+    oxpr = _parse_kv_expression(expr, all_schemas)
+    print(oxpr)
```
javitonino:
```suggestion
```

nucliadb_models/src/nucliadb_models/filters.py
```diff
@@ -303,98 +303,41 @@ class Status(FilterProp, extra="forbid"):
...
-    if isinstance(v, And):
-        return "and"
-    elif isinstance(v, Or):
-        return "or"
-    elif isinstance(v, Not):
-        return "not"
-    else:
-        return getattr(v, "op", None)
 
+class Lte(KVFilter):
```
javitonino: We talked about merging gte & lte so you could specify both at the same time. Thoughts?

nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -131,142 +131,126 @@ async def parse_expression(
...
+            else:
+                assert_never(expr.eq)
+
+        elif isinstance(expr, Gte):
+            if isinstance(expr.gte, float):
+                if schema_field.type != KVFieldType.FLOAT:
+                    raise InvalidKVType(expr.schema_id, schema_field, KVFieldType.FLOAT)
+                json_filter.path.float_range.lower = expr.gte
+            elif isinstance(expr.gte, int):
+                if schema_field.type != KVFieldType.INTEGER:
```
javitonino: We probably want to accept FLOAT fields here.

But in general, I feel we should try to do something more generic using the KEY_VALUE_ALLOWED_TYPES definition above, hopefully reducing the amount of nested ifs here.

jotare: Agree with the FLOAT.

For the generic thing, I think ee could add the expected types and schema types to the dict and avoid so many ifs but we'll lose the type checks. Right now, adding a new type to an operator raises a mypy error due to the `assert_never`, which I think is worth the ifs

nucliadb/src/nucliadb/common/filter_expression.py
jotare: Thanks

nucliadb_models/src/nucliadb_models/filters.py
jotare: What do you think of a Between operator with gte and lte inside? At some point we may want Gt and Lt, so we'll have a bunch of comparison operators anyway. I'm thinking more about the expression building in Python than in the API, a Between would enforce both and we won't have an operator with a range that can be None on either side

javitonino (approved):

PR #3671, opened 2026-05-22
Add integer range to key-value

jotare:
### Description
- Adding a new range type for integer ranges. The type has custom serialization in order to store a different representation in the database and decouple it from the models (it'd be too bad if a pydantic model change can break bw/c with stored data)
- Changed the field value serialization to include schema_id, otherwise, we can't know the schema from a GET
- Adding the `contains` operator to check if an integer is in a range

### How was this PR tested?
New tests

### DIFF
--- a/nucliadb/src/nucliadb/common/filter_expression.py
+++ b/nucliadb/src/nucliadb/common/filter_expression.py
@@ -158,7 +158,7 @@ KEY_VALUE_ALLOWED_TYPES: dict[Type[Eq] | Type[Inequalities] | Type[Contains], se
         KVFieldType.DATE,
     },
     Inequalities: {KVFieldType.INTEGER, KVFieldType.FLOAT, KVFieldType.DATE},
-    Contains: {KVFieldType.RANGE},
+    Contains: {KVFieldType.INTEGER, KVFieldType.FLOAT, KVFieldType.DATE},
 }
 
 # Map from Python types a value can have and which key-value field types can be
@@ -167,7 +167,7 @@ KV_VALUE_FIELD_TYPES: dict[type, list[KVFieldType]] = {
     str: [KVFieldType.TEXT],
     bool: [KVFieldType.BOOLEAN],
     float: [KVFieldType.FLOAT],
-    int: [KVFieldType.INTEGER, KVFieldType.FLOAT, KVFieldType.RANGE],
+    int: [KVFieldType.INTEGER, KVFieldType.FLOAT],
     datetime: [KVFieldType.DATE],
 }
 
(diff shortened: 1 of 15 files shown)


nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -277,6 +283,23 @@ def _parse_kv_expression(
...
 
+    elif isinstance(expr, Contains):
+        _validate_kv_schema(schemas, expr)
+
+        lower_bound = nodereader_pb2.JsonFilterExpression()
+        upper_bound = nodereader_pb2.JsonFilterExpression()
+
+        lower_bound.path.field_id = f"k/{expr.schema_id}"
+        upper_bound.path.field_id = f"k/{expr.schema_id}"
+        lower_bound.path.json_path = f"{expr.key}._range_min"
```
javitonino: Since this is internal and we are not letting users write their own, using `min` and `max` might be fine if we prefer. Nothing wrong with these names either.

jotare: Sure, I'll change it

nucliadb_models/src/nucliadb_models/kv_schemas.py
```diff
@@ -56,6 +56,10 @@ class KVSchemaField(BaseModel):
     type: KVFieldType
     description: str = ""
     required: bool = True
+    range: bool = Field(
```
javitonino, line 56: Maybe add a validation that this is false for string/bool, etc.

nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -191,6 +197,14 @@ def _validate_kv_schema(
             f"but '{type(expr).__name__.lower()}' requires type {allowed}",
         )
 
+    # validate range operations
+    if schema_field.range is False and isinstance(expr, (Contains,)):
```
javitonino: We also have to check the reverse, eq/inequalities probably should not be used with ranges (or if they do, they need specific implementation).

nucliadb/src/nucliadb/ingest/fields/key_value.py
```diff
@@ -97,3 +109,19 @@ def check_kv_type(schema_name: str, key: str, value: object, expected: KVFieldTy
         raise ValueError(
             f"Key {key!r} in schema {schema_name!r} expects type {expected.value!r}, got {type(value).__name__}"
         )
+
+
+def is_datetime(value: Any) -> bool:
```
javitonino, line 107: This is my fault, but extracting this function (❤️) made me realize, maybe this check should be like the one in str:

```
    if expected is KVFieldType.TEXT:
        if isinstance(value, str):
            try:
                dt = datetime.fromisoformat(value)
                # Tantivy's JSON indexer auto-parses strings as DateTime only when
                # they parse as RFC 3339, which requires both a time component and a
                # timezone offset (Z or ±HH:MM).
                ok = dt.tzinfo is None
            except ValueError:
                ok = True  # not parseable as a date at all, safe
        else:
            ok = False
```

and use the same check (inverted) for both places. So, something like

```
def is_datetime(value: Any) -> bool:
    # Dates must be stored as RFC 3339, which requires both a time component and a
    # timezone offset (Z or ±HH:MM).
    if isinstance(value, datetime):
        ok = dt.tzinfo is not None
    elif isinstance(value, str):
        try:
            datetime.fromisoformat(value)
            ok = dt.tzinfo is not None
        except ValueError:
            ok = False
    else:
        ok = False
    return ok
```

javitonino (approved):

PR #3680, opened 2026-05-27
fix: Add missing Search Configuration migrations

carlesonielfa:
### Description
Uses #3331 as inspiration

### How was this PR tested?
Created test for the migration

### DIFF
--- a/nucliadb/tests/nucliadb/migrations/test_migration_0044.py
+++ b/nucliadb/tests/nucliadb/migrations/test_migration_0044.py
@@ -28,6 +28,7 @@ from nucliadb_models.configuration import (
     AskSearchConfiguration,
     FindConfig,
     FindSearchConfiguration,
+    SearchConfiguration,
 )
 from tests.nucliadb.migrations import get_migration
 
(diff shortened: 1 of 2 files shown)


nucliadb/tests/nucliadb/migrations/test_migration_0044.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+        await datamanagers.search_configurations.set(
+            txn,
+            kbid=kbid,
+            name="gemini-2.0-flash-lite",
+            config=AskSearchConfiguration(
+                kind="ask", config=AskConfig(generative_model="gemini-2.0-flash-lite")
+            ),
+        )
+        await txn.commit()
```
jotare: NIT:
Could you simplify this as a list so it's a bit more readable. Something like:
```
deprecated_models = [
    ("gemini-2.0-flash-lite", AskConfig)
    ...
]
```
and inside the transaction iterate and do the search configuration set.

lferran (approved):

jotare (approved):

nucliadb/tests/nucliadb/migrations/test_migration_0044.py
carlesonielfa: perfect, simplified.

PR #3693, opened 2026-06-01
Support slug on resource field prefix filter

lferran:
### Description
[sc-14833]

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/common/filter_expression.py
+++ b/nucliadb/src/nucliadb/common/filter_expression.py
@@ -115,13 +115,16 @@ async def parse_expression(
     elif isinstance(expr, ResourceFieldPrefix):
         if expr.resource_id:
             f.resource_field_prefix.resource_id = expr.resource_id
-        if expr.resource_slug:
+        elif expr.resource_slug:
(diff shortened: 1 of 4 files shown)


nucliadb/src/nucliadb/common/filter_expression.py
```diff
@@ -113,7 +113,18 @@ async def parse_expression(
...
+        elif expr.resource_slug:
+            rid = await datamanagers.atomic.resources.get_resource_uuid_from_slug(
+                kbid=kbid, slug=expr.resource_slug
+            )
+            if rid is None:
+                raise InvalidQueryError("slug", f"Cannot find slug {expr.resource_slug}")
+            f.resource_field_prefix.resource_id = rid
+        else:
+            # Cannot happen due to model validation
+            raise ValueError("ResourceFieldPrefix needs resource_id or resource_slug")
```
jotare, line 127: Could you add a unit test in nucliadb_models for this? Just to make sure we never have a regression. Something as simple as a:
```python
with pytest raises(ValueError):
    filters.ResourceFieldPrefix(resource_id=None, resource_slug=None)
```

jotare (approved):

PR #3704, opened 2026-06-02
Clean vector result text fetching

jotare:
### Description
The code to obtain the paragraph text from vector results in /search was rather legacy and abandoned. This PR is an attempt to fix it and hopefully not breaking anything.

There are a couple of points to keep in mind for the review:
- In an old PR (https://github.com/nuclia/nucliadb/pull/1285) we changed the metadata indexed for each vector from the start-end of the vector to the start-end of the paragraph. This, as far as I can think, can only be useful in the context of multi-vector paragraphs, where the start-end wasn't the same as the paragraph stored in the keyword index.
- As we've done many refactors since then, we can assume all indexes have this metadata properly populated (so no need for the bw/c branch) 
- In the `get_text_sentence` there was a `start - 1` that doesn't make sense with the paragraph indexes
- The comment on the `get_text_sentence` saying we don't understand the differences was just scary

So, my hypothesis is the following: we can get rid of the bw/c code and the `start - 1` is a bug as things are right now.

### How was this PR tested?
Our current test suite.

In addition, I did some local tests to find the same paragraph in a /search with keyword and semantic. The results show how the text from semantic search has 1 char more than the keyword search one.

### DIFF
--- a/nucliadb/src/nucliadb/search/search/merge.py
+++ b/nucliadb/src/nucliadb/search/search/merge.py
@@ -35,7 +35,7 @@ from nidx_protos.nodereader_pb2 import (
     VectorSearchResponse,
 )
 
-from nucliadb.common.ids import FieldId, ParagraphId
+from nucliadb.common.ids import FieldId, ParagraphId, VectorId
 from nucliadb.common.models_utils import from_proto
 from nucliadb.common.models_utils.from_proto import (
     RelationNodeTypeMap,
@@ -79,7 +79,7 @@ from nucliadb_models.search import (
 from nucliadb_protos.utils_pb2 import RelationNode
 
 from .metrics import merge_observer
-from .paragraphs import get_paragraph_text, get_text_sentence
+from .paragraphs import get_paragraph_text
 
 Bm25Score = tuple[float, float]
 TimestampScore = datetime.datetime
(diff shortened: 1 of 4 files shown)


nucliadb/src/nucliadb/search/search/merge.py
```diff
@@ -258,49 +258,36 @@ async def merge_vectors_results(
...
-                index_int = -1
-        text = await get_text_sentence(
-            rid, field_type, field, kbid, index_int, start_int, end_int, subfield
+        vector_id = VectorId.from_string(result.doc_id.id)
+        # In case we have multiple vectors per paragraph, the vector id will
+        # have its start-end referencing a portion of the paragraph. However, we
+        # are interested in the whole paragraph, not only the portion, so we'll
+        # use the metadata.position to get the actual paragraph start-end
+        # positions
+        paragraph_id = ParagraphId(
```
javitonino, line 267: You may want to use the same as in find?

vector_id = VectorId.from_string(item.doc_id.id)
ParagraphId.from_vector_id(vector_id)

At least this way it is consistent, isn't it?

jotare: Sure, but we have to overwrite the start-end anyways

javitonino (approved):

jotare:
[sc-14908]

PR #3705, opened 2026-06-02
Use augmentor for /search merge (paragraphs and vectors)

jotare:
### Description
Leverage augmentor inside paragraph and vector merge to use nidx as extracted text storage when possible and more concurrency for vectors

### How was this PR tested?
Existing test suite


nucliadb/src/nucliadb/search/search/merge.py
```diff
@@ -380,25 +417,25 @@ async def merge_paragraph_results(
...
+    text = augmented.text or "" if augmented else ""
+    if text and highlight:
+        try:
+            text = highlight_paragraph(
+                text,
+                words=result.matches,  # type: ignore[arg-type,ty:invalid-argument-type]
+                ematches=ematches,
+            )
+        except Exception as ex:
+            errors.capture_exception(ex)
```
lferran: careful, this could easily fill up our sentry quota

nucliadb/src/nucliadb/search/search/merge.py
```diff
@@ -342,27 +353,53 @@ async def merge_paragraph_results(
...
 
-    result_resource_ids = []
+    augmented_paragraphs: dict[ParagraphId, AugmentedParagraph] = await augment_paragraphs(
+        kbid,
+        given=[
+            AugmentorParagraph(id=paragraph_id, metadata=Metadata.from_paragraph_result(result))
+            for paragraph_id, result, _ in raw_paragraph_list
+        ],
+        select=[ParagraphText()],
+        concurrency_control=asyncio.Semaphore(20),
```
lferran: I think we can put a higher value here no? 50? I don't know if it's even needed, now that we are querying nidx?

lferran (approved):

nucliadb/src/nucliadb/search/search/merge.py
jotare: This was already like this, but I'll change it for a warning

nucliadb/src/nucliadb/search/search/merge.py
jotare: I just added a shared semaphore between paragraph and vector merges of 50

jotare:
[sc-14908]

PR #3712, opened 2026-06-04
Skip useless object storage trips

jotare:
### Description
In /search and /suggest merge, we are doing a bunch of lookups to maindb/storage that are actually not needed as the information is already returned by the index. In addition, we were not properly utilizing request cache in fetch.py, concurrency in key points and some other improvements around.

### How was this PR tested?


lferran (approved):

jotare:
[sc-14908]

PR #3715, opened 2026-06-04
Suggest optimizations

jotare:
### Description
- We had a top 20 hardcoded in nidx but only 10 results were actually used in the Python side (introduced here https://github.com/nuclia/nucliadb/pull/235)
- Use augmentor to get paragraph text instead of one-by-one from storage

### How was this PR tested?
Existing test suite

### DIFF
--- a/nucliadb/src/nucliadb/search/search/merge.py
+++ b/nucliadb/src/nucliadb/search/search/merge.py
@@ -209,11 +209,7 @@ async def merge_suggest_paragraph_results(
         text = augmented.text or "" if augmented else ""
         if text and highlight:
             try:
-                text = highlight_paragraph(
-                    text,
-                    words=match.matches,  # type: ignore[arg-type,ty:invalid-argument-type]
-                    ematches=ematches,
-                )
+                text = highlight_paragraph(text, words=match.matches, ematches=ematches)
             except Exception as exc:
                 logger.warning("Error highlighting paragraph", extra={"kbid": kbid}, exc_info=exc)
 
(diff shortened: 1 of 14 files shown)


nucliadb/src/nucliadb/search/search/merge.py
```diff
@@ -173,61 +174,79 @@ async def merge_suggest_paragraph_results(
...
+        ematches.extend(suggest_response.ematches)
+
+    # merge while keeping sorted order by score
+    merged = heapq.merge(
+        *(response.results for response in suggest_responses),
+        reverse=True,
+        key=lambda x: (x.score.bm25, x.score.booster),
+    )
+    # cut the best results
+    matches = list(itertools.islice(merged, MAX_SUGGEST_RESULTS))
```
javitonino: Should top_k passed here to get an accurate cut?

nucliadb/src/nucliadb/search/search/merge.py
```diff
@@ -173,61 +174,79 @@ async def merge_suggest_paragraph_results(
...
+
+    results = []
+    for match in matches:
+        augmented = augmented_paragraphs.get(paragraph_id)
+        text = augmented.text or "" if augmented else ""
+        if text and highlight:
+            try:
+                text = highlight_paragraph(
+                    text,
+                    words=match.matches,  # type: ignore[arg-type,ty:invalid-argument-type]
```
javitonino: You can probably fix this properly by changing the type of the argument in highlight_paragraph to be Sequence[X] instead of list[X]

lferran (approved):

jotare:
[sc-14908]

PR #3741, opened 2026-06-17
Refactor basic metadata handling — move logic from Resource to Processor

lferran:
### Description
Ingest ORM resource layer was not consistent, and included some processor business logic.
This PR refactors the code and moves the business logic to the processor code, making our path easier towards having:

- faster saving of extracted data to storage: we can parallelize it now.
- faster saving of field values to pg: can could potentially merge all of the updates into a single update operation.
  

### How was this PR tested?
- Existing integration tests should cover us
- Will add new tests aswell


nucliadb/src/nucliadb/ingest/orm/processor/processor.py
```diff
@@ -586,50 +599,79 @@ async def external_index_add_resource(
     async def deadletter(self, message: writer_pb2.BrokerMessage, partition: str, seqid: int) -> None:
         await self.storage.deadletter(message, 0, seqid, partition)
 
+    async def _get_resource_filenames(self, resource: Resource) -> list[str]:
```
jotare: This should be either a function or a method on Resource, but it's not related with Processor at all (you're not using self)

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -485,32 +485,31 @@ async def apply_field_extracted_data(self, message: BrokerMessage) -> None:
...
         question answers, field metadata, large metadata). Does not touch basic.
         """
+        tasks = []
         for question_answers in message.question_answers:
-            await self._apply_question_answers(question_answers)
+            tasks.append(self._apply_question_answers(question_answers))
 
         for field_id in message.delete_question_answers:
-            await self._delete_question_answers(field_id)
+            tasks.append(self._delete_question_answers(field_id))
```
jotare: A concurrent set and delete on the same field is not safe. In fact, the code as it was could delete a question answer if processing is sending a set+delete as an overwrite. We should delete and set (or only delete the ones we won't set)

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -485,32 +485,31 @@ async def apply_field_extracted_data(self, message: BrokerMessage) -> None:
...
 
-        tasks = []
         for extracted_text in message.extracted_text:
             tasks.append(self._apply_extracted_text(extracted_text))
+
```
jotare, line 496: Could you add a comment explaining why those 2 separate concurrent groups of operations can't be done together?

nucliadb/src/nucliadb/ingest/orm/resource.py
```diff
@@ -194,11 +194,13 @@ async def set_user_relations(self, payload: PBRelations):
         self.user_relations = payload
 
     # Fields
-    async def get_fields(self, force: bool = False) -> dict[tuple[FieldType.ValueType, str], Field]:
+    async def get_fields(
+        self, force: bool = False, load_values: bool = False
```
jotare, line 199: NIT: it's not obvious from the caller what the boolean parameters do. A `(..., load_values=True, force=False)` would be hard to understand is not reusing the cached list of field ids and loading the value only when is not cached.

nucliadb/src/nucliadb/ingest/orm/resource.py
jotare: (A couple of lines below the one I marked)

nucliadb/src/nucliadb/ingest/orm/resource.py
lferran: I know.. I want to get rid of all of this anyway with the new implementation

jotare (approved):

PR #3752, opened 2026-06-22
Multi-shard search in nidx

jotare:
### Description
This is the first part of a series of PRs to improve how internal gRPC search requests are done through the cluster. We currently send 1 gRPC request per shard, but we could instead send 1 per searcher (containing a KB shard).

In this PR, we implement multi-shard requests and shard response merging in Rust, although Python will still request 1 per shard.

The change on the proto is bw/c compatible (see https://protobuf.dev/programming-guides/proto3/#wire-safe-changes). Python sends 1 shard id string as a string/list and nidx parses 1 shard id string as a string/list. 

### How was this PR tested?
New Rust tests on merging

### DIFF
--- a/nidx/src/searcher/grpc.rs
+++ b/nidx/src/searcher/grpc.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::{pin::Pin, sync::Arc};
 
 use futures::Stream;
@@ -22,6 +22,7 @@ use nidx_protos::nidx::nidx_searcher_server::{NidxSearcher, NidxSearcherServer};
 use nidx_protos::*;
 use tokio::sync::RwLock;
 use tokio::task::JoinSet;
+use tonic::Code;
 use tonic::service::Interceptor;
 use tonic::service::interceptor::InterceptedService;
 use tonic::transport::Channel;
@@ -163,7 +164,7 @@ impl NidxSearcher for SearchServer {
 
         let mut shards = vec![];
         for shard_id in &request.shard_ids {
-            let shard_id = uuid::Uuid::parse_str(shard_id).map_err(NidxError::from)?;
+            let shard_id = Uuid::parse_str(shard_id).map_err(NidxError::from)?;
             shards.push(shard_id);
         }
 
@@ -185,7 +186,7 @@ impl NidxSearcher for SearchServer {
 
     async fn suggest(&self, request: Request<SuggestRequest>) -> Result<Response<SuggestResponse>> {
         let message = request.get_ref();
-        let shard_id = uuid::Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
+        let shard_id = Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
         shard_request! {
             "suggest",
             self,
@@ -198,7 +199,7 @@ impl NidxSearcher for SearchServer {
 
     async fn graph_search(&self, request: Request<GraphSearchRequest>) -> Result<Response<GraphSearchResponse>> {
         let message = request.get_ref();
-        let shard_id = uuid::Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
+        let shard_id = Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
         shard_request! {
             "graph_search",
             self,
@@ -214,7 +215,7 @@ impl NidxSearcher for SearchServer {
         request: Request<ExtractedTextsRequest>,
     ) -> Result<Response<ExtractedTextsResponse>> {
         let message = request.get_ref();
-        let shard_id = uuid::Uuid::parse_str(&message.shard_id).map_err(NidxError::from)?;
+        let shard_id = Uuid::parse_str(&message.shard_id).map_err(NidxError::from)?;
         shard_request! {
             "extracted_texts",
             self,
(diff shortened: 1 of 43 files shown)


nidx/src/searcher/grpc.rs
```diff
@@ -152,14 +153,37 @@ macro_rules! shard_request {
...
-        let shard_id = uuid::Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
-        shard_request! {
-            "search",
-            self,
-            request,
-            shard_id,
-            LOCAL => shard_search::search(Arc::clone(&self.index_cache), message.clone()).await,
-            REMOTE => search
+
+        if !message.multi_shard_search {
```
javitonino: Is this really needed? I think we could just do the else case always and it should be the same

nidx/src/searcher/shard_merge.rs
```diff
@@ -0,0 +1,1074 @@
...
+    merged.results.truncate(limit.0);
+
+    merged
+}
+
+fn sort_documents(items: &mut [DocumentResult], order_by: OrderBy) {
+    match order_by.expr {
+        SortExpr::Score => {
+            debug_assert!(order_by.descending, "order by ascending score is not implemented");
+            items.sort_unstable_by(|a, b| match (a.sort_value, b.sort_value) {
```
javitonino: We may need stable sort here to guarantee result consistency if we care about that. This is just a note to think about this.

nidx/src/searcher/shard_merge.rs
```diff
@@ -0,0 +1,1074 @@
...
+            });
+        }
+    }
+}
+
+fn merge_vector_responses(responses: Vec<VectorSearchResponse>, limit: Limit) -> VectorSearchResponse {
+    debug_assert!(!responses.is_empty(), "must pass at least 1 shard response");
+    let mut merged = VectorSearchResponse::default();
+
+    // TODO: sort-merge and stop when limit is reached would be faster
```
javitonino: I think we can use itertools' [merge_join_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge_join_by) to do this quite easily.

nidx/src/searcher/grpc.rs
jotare: Agree but @lferran suggested a FF mechanism so we could rollback to the original implementation

nidx/src/searcher/shard_merge.rs
jotare: I'm not sure we care that much. I choose an unstable sort as I thought it'd be really rare that two paragraphs match with the same BM25 score (it probably happens with identical paragraphs, but in that case we don't care on order). On another hand, we query and merge shards unordered, so the order across shards would depend on shard position in the list too.
In any case, we can use a stable sort and at least be more consistent

nidx/src/searcher/grpc.rs
```diff
@@ -152,14 +153,29 @@ macro_rules! shard_request {
...
-        let shard_id = uuid::Uuid::parse_str(&message.shard).map_err(NidxError::from)?;
-        shard_request! {
-            "search",
-            self,
-            request,
-            shard_id,
-            LOCAL => shard_search::search(Arc::clone(&self.index_cache), message.clone()).await,
-            REMOTE => search
+
+        if message.shard_ids.len() == 1 {
```
javitonino: Maybe this should only be for the merge itself?

like, always iterate all queries and then do an if to run the merge code or just return the first result?

nidx/src/searcher/grpc.rs
```diff
@@ -152,14 +153,29 @@ macro_rules! shard_request {
...
+            let shard_id = uuid::Uuid::parse_str(&message.shard_ids[0]).map_err(NidxError::from)?;
+            self.shard_search(shard_id, &request).await
+        } else {
+            // Distributed query across multiple shards
+            let mut responses = Vec::with_capacity(message.shard_ids.len());
+
+            for shard_id in &message.shard_ids {
```
javitonino: We probably want to run all task concurrently (with a JoinSet or similar). Otherwise the query time is linear to the number of shards. I think Python already sent all queries in parallel, so this would be a regression.

jotare: Next PR. We can merge them and have it all here

nidx/src/searcher/grpc.rs
```diff
@@ -241,3 +261,205 @@ impl NidxSearcher for SearchServer {
...
+            shard_merge::merge(
+                responses,
+                OrderBy::from(&request),
+                Limit(request.result_per_page as usize),
+            )
+        };
+        Ok(merged)
+    }
+
+    async fn node_query(
```
javitonino, line 355: Looks good. An alternative if we want to avoid the task that returns an async block pattern would be to only pass what's needed here, since the index_cache and tonic clients are already inside an Arc:

```rust
    async fn node_query(
        index_cache: Arc<IndexCache>,
        clients: Arc<RwLock<HashMap<String, SearcherClient>>>,
        node: SearcherNode,
        shard_ids: Vec<uuid::Uuid>,
        request: SearchRequest,
    ) -> NidxResult<PartialResponse<SearchResponse>> {
```

and call it with 
```rust
                let index_cache = Arc::clone(&self.index_cache);
                let clients = Arc::clone(&self.clients);
                let request = request.clone();
                tasks.spawn(async move { Self::node_query(index_cache, clients, node, shard_ids, request).await });
```

jotare: Yeah, I also thought about that but for now I prefer this pattern. If we want to do this more generic, we'll have to change how this parameters are passed anyway

javitonino (dismissed):

jotare:
[sc-14854]

javitonino (approved):

PR #3759, opened 2026-06-24
Distributed multi-shard query

jotare:
### Description
The next step towards moving multi-shard querying to Rust. This PR implements forwarding of search requests with grouped shards between nidx-searcher nodes

### How was this PR tested?
New integration tests

### DIFF
--- a/nidx/src/searcher/grpc.rs
+++ b/nidx/src/searcher/grpc.rs
@@ -21,6 +21,7 @@ use nidx_protos::nidx::nidx_searcher_client::NidxSearcherClient;
 use nidx_protos::nidx::nidx_searcher_server::{NidxSearcher, NidxSearcherServer};
 use nidx_protos::*;
 use tokio::sync::RwLock;
+use tokio::task::JoinSet;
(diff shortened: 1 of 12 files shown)


nidx/src/searcher/grpc.rs
```diff
@@ -267,19 +270,175 @@ impl NidxSearcher for SearchServer {
...
+
+                let node = nodes.remove(0);
+                shard_groups
+                    .entry(node)
+                    .and_modify(|shards: &mut Vec<uuid::Uuid>| shards.push(*shard_id))
+                    .or_insert(vec![*shard_id]);
+            }
+
+            let mut tasks = JoinSet::new();
+            for (node, shard_ids) in shard_groups {
```
javitonino, line 317: I'm getting lost reading this function since it's quite long and nested. Maybe the contents of this loop can be extracted?

jotare: I'm not sure extracting the whole loop would make it more readable. I'll see what I can do

jotare:
[sc-14854]

PR #3790, opened 2026-07-09
Sanitize user data sent to audit

lferran:
### Description

[sc-14950]

### How was this PR tested?


drf7 (dismissed):

nucliadb/src/nucliadb/ingest/consumer/auditing.py
```diff
@@ -103,6 +103,9 @@ async def handle_message(self, msg: Any) -> None:
 
     @metrics.handler_histo.wrap({"type": "audit_counter"})
     async def process_kb(self, kbid: str) -> None:
+        if not await datamanagers.atomic.kb.exists_kb(kbid=kbid):
+            logger.info(f"KB does not exist, skipping counter audit", extra={"kbid": kbid})
```
jotare: This should be a warning. On cloud this should never happen

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -149,6 +155,52 @@ def enqueue_pending(self, context: RequestContext):
...
+        return kls.model_validate(json_body).model_dump_json(exclude_unset=True)
+    except ValidationError as exc:
+        logger.warning(f"Invalid request body for audit: {request.url.path} - {exc}")
+        return None
+    except json.JSONDecodeError as exc:
+        logger.warning(f"Invalid JSON for audit: {request.url.path} - {exc}")
+        return None
+    except KeyError as exc:
+        logger.warning(f"Invalid endpoint for audit: {request.url.path} - {exc}")
+        return None
```
jotare: You could remove this by adding more typing info on things (e.g., instead of string a union of literals of the endpoints). That way ty/mypy would complain if we add a new endpoint at some point

nucliadb_utils/src/nucliadb_utils/audit/stream.py
```diff
@@ -149,6 +155,52 @@ def enqueue_pending(self, context: RequestContext):
...
+
+
+async def valid_query_params(request: Request, endpoint: str) -> str | None:
+    query_params = dict(request.query_params)
+    try:
+        kls = endpoint_validators[endpoint]
+    except KeyError:
+        logger.warning(f"Invalid endpoint for audit: {request.url.path}")
+        return None
```
jotare: Same as above (use more typing)

nucliadb_utils/tests/unit/audit/test_stream.py
```diff
@@ -164,3 +166,156 @@ async def test_chat(audit_storage: StreamAuditStorage, nats):
...
     assert pb.generative_answer_first_chunk_time == 1
     assert pb.generative_reasoning_first_chunk_time == 1
+
+
+# ---------------------------------------------------------------------------
+# Tests for valid_payload
+# ---------------------------------------------------------------------------
+
+
+class TestValidUserRequest:
```
jotare, line 176: You could really reduce boilerplate with a pytest.mark.parametrize, should be quite easy

jotare (approved):

PR #3800, opened 2026-07-10
Validate KB and backup bucket exist before a restore

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/backups/restore.py
+++ b/nucliadb/src/nucliadb/backups/restore.py
@@ -18,7 +18,6 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
-
 import asyncio
 import functools
 import json
@@ -32,6 +31,7 @@ from pydantic import TypeAdapter
 from nucliadb.backups.const import MaindbKeys, StorageKeys
 from nucliadb.backups.models import RestoreBackupRequest
 from nucliadb.backups.settings import settings
+from nucliadb.common import datamanagers
 from nucliadb.common.context import ApplicationContext
 from nucliadb.export_import.utils import (
     import_binary,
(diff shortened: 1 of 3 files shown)


nucliadb/src/nucliadb/backups/restore.py
```diff
@@ -74,6 +74,21 @@ async def restore_kb(context: ApplicationContext, kbid: str, backup_id: str):
...
     """
+    async with context.kv_driver.ro_transaction() as txn:
+        kb_exists = await datamanagers.kb.exists_kb(txn, kbid=kbid)
+        if not kb_exists:
+            logger.warning(
+                "Trying to restore on a KB that doesn't exist",
+                extra={"kbid": kbid, "backup_id": backup_id},
+            )
+            return
+    if not await context.blob_storage.bucket_exists(backup_id):
```
lferran: this check is not correct. instead we should check if an object inside the backup folder in the backups bucket exists.

confusing naming xD

nucliadb/src/nucliadb/backups/restore.py
```diff
@@ -74,6 +74,21 @@ async def restore_kb(context: ApplicationContext, kbid: str, backup_id: str):
     """
     Downloads the backup files from the cloud storage and imports them into the KB.
     """
+    async with context.kv_driver.ro_transaction() as txn:
```
lferran: you can also use datamanager.atomic.kb.exists_kb(), maybe its cleaner?

jotare: If so, why aren't we using the global storage too? The rest of the code gets the utilities from the application context

lferran: I think it's simply different patterns we've wanted to use over time. it's ok like this too

nucliadb/src/nucliadb/backups/restore.py
```diff
@@ -74,6 +74,22 @@ async def restore_kb(context: ApplicationContext, kbid: str, backup_id: str):
...
     Downloads the backup files from the cloud storage and imports them into the KB.
     """
+    kb_exists = await datamanagers.atomic.kb.exists_kb(kbid=kbid)
+    if not kb_exists:
+        logger.warning(
+            "Trying to restore on a KB that doesn't exist",
+            extra={"kbid": kbid, "backup_id": backup_id},
+        )
+        return
+    if not await context.blob_storage.exists_object(
```
lferran: Can you add a test for this?

I'm not sure it will work when it's a folder what you are checking.

lferran: maybe a generic test in the storage layer is good enough

lferran (dismissed):

lferran (approved):

PR #3810, opened 2026-07-14
Validate resource and kb ids.

lferran:
### Description

This PR adds proper validation early on in the system, rather than in the datamanager layer.

If they are not uuids, return 404 to keep backward compatibility -- before, we were not validating and simply returning a kb or resource not found.

### How was this PR tested?
Existing tests and added new tests

### DIFF
--- a/nucliadb/src/nucliadb/common/fastapi.py
+++ b/nucliadb/src/nucliadb/common/fastapi.py
@@ -1,0 +1,60 @@
+# Copyright (C) 2021 Bosutech XXI S.L.
+#
+# nucliadb is offered under the AGPL v3.0 and as commercial software.
+# For commercial licensing, contact us at info@nuclia.com.
+#
+# AGPL:
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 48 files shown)


nucliadb/src/nucliadb/common/ids.py
```diff
@@ -310,3 +311,23 @@ def extract_data_augmentation_id(generated_field_id: str) -> str | None:
         return None
 
     return parts[1] or None
+
+
+def valid_kbid(kbid: str) -> bool:
+    # Must be a UUID in canonical dashed format.
+    if len(kbid) != 36 or kbid.count("-") != 4:
+        return False
```
jotare, line 319: Do we really want to validate the UUID format and have it different for kbid and rid? I know we currently use it like this, but seems an arbitrary choice

nucliadb/src/nucliadb/common/fastapi.py
```diff
@@ -0,0 +1,58 @@
...
+def _uuid_path_param(name: str, not_found_detail: str, validator_func: Callable[[str], bool]) -> Any:
+    """Return a FastAPI ``Depends`` callable that validates *name* as a UUID."""
+
+    def _validate(value: str = Path(alias=name)) -> str:
+        if not validator_func(value):
+            raise HTTPException(status_code=404, detail=not_found_detail)
+        return value
+
+    _validate.__name__ = f"validate_{name}"
+    return Depends(_validate)
```
jotare, line 53: I found this too hacky to return 404 for invalid UUIDs...
Also, using our own validated string parameters avoids using UUIDs internally, but maybe that's what we want (although it'd be a bigger change).

jotare (approved):

PR #3813, opened 2026-07-15
validate labelset kind

lferran:
### Description
[sc-15102]
### How was this PR tested?

### DIFF
--- a/nucliadb_models/src/nucliadb_models/labels.py
+++ b/nucliadb_models/src/nucliadb_models/labels.py
@@ -110,7 +110,11 @@ class LabelSet(BaseModel):
     )
     color: str | None = "blue"
     multiple: bool = True
-    kind: list[LabelSetKind] = []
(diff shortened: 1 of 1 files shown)


nucliadb_models/src/nucliadb_models/labels.py
```diff
@@ -110,7 +110,11 @@ class LabelSet(BaseModel):
     )
     color: str | None = "blue"
     multiple: bool = True
-    kind: list[LabelSetKind] = []
+    kind: list[LabelSetKind] = Field(
+        default_factory=lambda: [LabelSetKind.RESOURCES],
```
jotare: You don't need a default_factory, pydantic already does deep copy of unhashable values (like lists): https://pydantic.dev/docs/validation/latest/concepts/fields/#mutable-default-values

jotare (approved):

PR #3816, opened 2026-07-17
Nidx multi-shard for suggest and graph search

jotare:
### Description

### How was this PR tested?

### DIFF
--- a/nidx/tests/common/services.rs
+++ b/nidx/tests/common/services.rs
@@ -63,11 +63,7 @@ impl NidxFixture {
                 telemetry: Default::default(),
                 work_path: None,
                 control_socket: None,
-                searcher: Some(SearcherSettings {
-                    // tests can take advantage of faster refresh intervals
-                    metadata_refresh_interval: 0.1,
-                    ..Default::default()
-                }),
+                searcher: Some(SearcherSettings::default()),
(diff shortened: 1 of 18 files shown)


nidx/tests/common/services.rs
```diff
@@ -138,6 +134,13 @@ impl NidxFixture {
 
     pub async fn wait_sync(&self) {
         // TODO: Check the searcher has synced? For now, waiting twice the sync interval
-        tokio::time::sleep(Duration::from_secs(2)).await;
+        tokio::time::sleep(Duration::from_secs_f32(
+            2.0 * self
```
javitonino, line 138: Does this help?

I was playing around with trying to do this based on the sync channel, but it ended up being messy and no better.

javitonino (dismissed):

nidx/tests/common/services.rs
jotare: It's the same as it was, but that way it changes with the configuration. I had some issues with sqlx tests and pg pool timings running cluster tests in parallel. What actually fixed it was going back to a bigger metadata refresh interval

javitonino (approved):

jotare:
[sc-14854]

PR #3818, opened 2026-07-20
Stable booster

javitonino:
Changes how results are merged to guarantee a consistent order. Order is:
- Score
- Shard
- Booster (which is now defined as segment_id + doc_id)

Also add a missing test that I had not git add'ed in a previous PR https://github.com/nuclia/nucliadb/pull/3727

About booster remove:
- If nucliadb is updated first. It will try to read fields that are not set and will default to 0, which will cause random order.
- If nidx is updated first, old booster field is not set and nucliadb will read it as 0.0, which will cause random order.

I think this is fine for a temporary condition (and not much worse than the old booster anyway)

### DIFF
--- a/nidx/nidx_vector/tests/test_paragraph_merge.rs
+++ b/nidx/nidx_vector/tests/test_paragraph_merge.rs
@@ -65,14 +65,8 @@ fn build_resource(uuid: &str, vector0: Vec<f32>, vector1: Vec<f32>) -> Resource 
     }
 }
 
-/// Verifies that merging two paragraph segments—each carrying a self-referencing
-/// deletion for its own field—keeps both vectors alive in the merged output.
-///
-/// The deletion pattern (a deletion at the same seq as the segment) mirrors
-/// what happens during a resource update: the old version of a field is marked
-/// for deletion atomically with the new version being written.  When the two
-/// segments belong to *different* resources the deletions are disjoint and
-/// neither removes the other resource's vectors.
+/// Verifies that merging two segments that update a resource each are applied correctly.
+/// This requires that deletions are applied in the correct order during merging.
 #[test]
 fn test_paragraph_merge_with_deletions() -> anyhow::Result<()> {
     let config = VectorConfig::for_paragraphs(VectorType::DenseF32 { dimension: DIMENSION });
--- a/nidx/nidx_vector/tests/test_paragraph_merge.rs
+++ b/nidx/nidx_vector/tests/test_paragraph_merge.rs
@@ -1,0 +1,180 @@
+// Copyright 2021 Bosutech XXI S.L.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+mod common;
+
+use nidx_protos::{IndexParagraph, IndexParagraphs, Resource, ResourceId, VectorSentence};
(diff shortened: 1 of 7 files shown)


nidx/nidx_vector/tests/test_paragraph_merge.rs
```diff
@@ -0,0 +1,180 @@
...
+            make_field(uuid, "a/title", vector0),
+            make_field(uuid, "t/body", vector1),
+        ]
+        .into(),
+        ..Default::default()
+    }
+}
+
+/// Verifies that merging two paragraph segments—each carrying a self-referencing
+/// deletion for its own field—keeps both vectors alive in the merged output.
```
jotare: Can you remove this weirdly formatted stuff?
```suggestion
/// Verifies that merging two paragraph segments, each carrying a self-referencing
/// deletion for its own field, keeps both vectors alive in the merged output.
```

nidx/src/searcher/shard_merge.rs
```diff
@@ -55,12 +56,24 @@ pub fn merge(shard_responses: Vec<SearchResponse>, order_by: OrderBy, limit: Lim
...
+            for result in &mut document.results {
+                result.shard_id = shard_id.clone();
+            }
             document_responses.push(document);
         }
-        if let Some(paragraph) = response.paragraph {
+        if let Some(mut paragraph) = response.paragraph {
+            for result in &mut paragraph.results {
+                result.shard_id = shard_id.clone();
+            }
```
jotare: The assumption that all results are from the first shard is not correct. You should rely on search to fill the shard_id, otherwise you'll have wrongly labeled stuff and retries on different nodes will have different ordering

javitonino: My bad, too focused on the merge and too reliant on autocomplete :cry:

nidx/src/searcher/shard_merge.rs
```diff
@@ -871,6 +963,89 @@ mod tests {
...
+                Limit(20),
+            )
+            .paragraph
+            .unwrap();
+            assert_eq!(merged.results[0].uuid, "foo");
+            assert_eq!(merged.results[0].shard_id, shard_bytes(SHARD_B));
+            assert_eq!(merged.results[1].uuid, "bar");
+            assert_eq!(merged.results[1].shard_id, shard_bytes(SHARD_A));
+
+            // Reversing shard assignment flips the order, confirming determinism.
```
jotare: Misleading comment, it doesn't flip any order, it maintains the same and that's what we want.

nidx/src/searcher/shard_merge.rs
```diff
@@ -583,6 +592,88 @@ mod tests {
...
+                uuid: rid.to_string(),
+                field: "a/title".to_string(),
+                sort_value: Some(document_result::SortValue::Score(nidx_protos::ResultScore {
+                    bm25: score,
+                    booster,
+                })),
+                shard_id: shard_bytes(shard_id),
+                ..Default::default()
+            };
+            // shard_id is stamped by shard_search before merge; simulate that here
```
jotare: Misleading comment. This should be at L#608

nidx/src/searcher/shard_merge.rs
```diff
@@ -583,6 +592,88 @@ mod tests {
...
+                Limit(20),
+            )
+            .document
+            .unwrap();
+            assert_eq!(merged.results[0].uuid, "foo");
+            assert_eq!(merged.results[0].shard_id, shard_bytes(SHARD_B));
+            assert_eq!(merged.results[1].uuid, "bar");
+            assert_eq!(merged.results[1].shard_id, shard_bytes(SHARD_A));
+
+            // Reversing shard assignment flips the order.
```
jotare: Misleading comment: again, reversing shard assignment doesn't flip the order, and that's what we want

javitonino: I think this is actually correct. SHARD_B is returned first (breaks the tie) so bar is returned first in this case. In the previous case, foo was returned first.

nidx/src/searcher/shard_merge.rs
javitonino: Same as above

jotare (approved):

PR #3820, opened 2026-07-21
Search after

javitonino:
### Description

Allows to paginate search after a hybrid search. Subsequent pages will only include keyword results. Since the first page is hybrid search and can return results out of order, a list of shown results is passed to continued search to skip those results.

### How was this PR tested?

### DIFF
--- a/nidx/src/searcher/shard_merge.rs
+++ b/nidx/src/searcher/shard_merge.rs
@@ -200,16 +200,16 @@ fn sort_documents_fn(order_by: OrderBy) -> Box<dyn Fn(&DocumentResult, &Document
                 (
                     Some(document_result::SortValue::Score(nidx_protos::ResultScore {
                         bm25: a_bm25,
-                        booster: a_booster,
+                        docaddr: a_docaddr,
                     })),
                     Some(document_result::SortValue::Score(nidx_protos::ResultScore {
                         bm25: b_bm25,
-                        booster: b_booster,
+                        docaddr: b_docaddr,
                     })),
                 ) => a_bm25
                     .total_cmp(&b_bm25)
                     .then(a.shard_id.cmp(&b.shard_id))
-                    .then(a_booster.cmp(&b_booster))
+                    .then(a_docaddr.cmp(&b_docaddr))
                     .is_gt(),
                 _ => {
                     unreachable!("index always return values with the same order_by as we have")
@@ -278,16 +278,16 @@ fn sort_paragraphs_fn(order_by: OrderBy) -> Box<dyn Fn(&ParagraphResult, &Paragr
                 (
                     Some(paragraph_result::SortValue::Score(nidx_protos::ResultScore {
                         bm25: a_bm25,
-                        booster: a_booster,
+                        docaddr: a_docaddr,
                     })),
                     Some(paragraph_result::SortValue::Score(nidx_protos::ResultScore {
                         bm25: b_bm25,
-                        booster: b_booster,
+                        docaddr: b_docaddr,
                     })),
                 ) => a_bm25
                     .total_cmp(&b_bm25)
                     .then(a.shard_id.cmp(&b.shard_id))
-                    .then(a_booster.cmp(&b_booster))
+                    .then(a_docaddr.cmp(&b_docaddr).reverse())
                     .is_gt(),
                 _ => {
                     unreachable!("index always return values with the same order_by as we have")
@@ -622,12 +622,12 @@ mod tests {
 
         #[test]
         fn test_merge_document_results_by_score() {
-            let document = |rid: &str, score: f32, booster: u64| DocumentResult {
+            let document = |rid: &str, score: f32, docaddr: u64| DocumentResult {
                 uuid: rid.to_string(),
                 field: "a/title".to_string(),
                 sort_value: Some(document_result::SortValue::Score(nidx_protos::ResultScore {
                     bm25: score,
-                    booster,
+                    docaddr,
                 })),
                 ..Default::default()
             };
@@ -666,12 +666,12 @@ mod tests {
             const SHARD_B: &str = "bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb";
             let shard_bytes = |s: &str| Uuid::parse_str(s).unwrap().as_bytes().to_vec();
 
-            let document = |rid: &str, score: f32, booster: u64, shard_id: &str| DocumentResult {
+            let document = |rid: &str, score: f32, docaddr: u64, shard_id: &str| DocumentResult {
                 uuid: rid.to_string(),
                 field: "a/title".to_string(),
                 sort_value: Some(document_result::SortValue::Score(nidx_protos::ResultScore {
                     bm25: score,
-                    booster,
+                    docaddr,
                 })),
                 // shard_id is set by shard_search before merge; simulate that here
                 shard_id: shard_bytes(shard_id),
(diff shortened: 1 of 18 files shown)


nucliadb/tests/nucliadb/integration/search/test_search_after.py
```diff
@@ -0,0 +1,199 @@
...
+        "reranker": "noop",
+        "top_k": 100,
+    }
+    resp = await nucliadb_reader.post(f"/kb/{kbid}/find", json=find_payload)
+    assert resp.status_code == 200, resp.text
+    baseline_data = resp.json()
+
+    baseline_ids = baseline_data["best_matches"]
+
+    assert len(baseline_ids) >= 5
```
jotare: Add a constant for 5 if you want to be more flexible. But we should never have more than 5
```suggestion
    assert len(baseline_ids) == 5
```

nucliadb/tests/nucliadb/integration/search/test_search_after.py
```diff
@@ -0,0 +1,199 @@
...
+        paginated_ids.extend(page_ids)
+
+        token = body.get("search_after")
+        token_present = token is not None
+        assert token_present == body["next_page"]
+
+        if not token or len(page_ids) < 1:
+            break
+
+    assert set(paginated_ids) == set(baseline_ids)
```
jotare, line 96: You're only checking ids, not the actual order

nucliadb/tests/nucliadb/integration/search/test_search_after.py
```diff
@@ -0,0 +1,199 @@
...
+        json={
+            "query": "potatoes",
+            "features": ["keyword"],
+            "reranker": "noop",
+            "top_k": 100,
+        },
+    )
+    assert resp.status_code == 200, resp.text
+    baseline_ids = set(resp.json()["best_matches"])
+    assert len(baseline_ids) >= 7
```
jotare: Same as above
```suggestion
    assert len(baseline_ids) == 7
```

nucliadb/tests/nucliadb/integration/search/test_search_after.py
```diff
@@ -0,0 +1,199 @@
...
+                assert not match.startswith(hybrid_rid)
+
+        if not token or not page_ids or not body["next_page"]:
+            break
+
+    # No duplicates
+    assert len(paginated_ids) == len(set(paginated_ids))
+
+    # Same results
+    assert set(baseline_ids) == set(paginated_ids)
```
jotare, line 199: Again, same results, but same order?

nucliadb_models/src/nucliadb_models/search.py
```diff
@@ -1983,6 +1983,14 @@ class FindRequest(BaseSearchRequest):
         title="Generative model",
         description="The generative model used to rephrase the query. If not provided, the model configured for the Knowledge Box is used.",
     )
+    search_after: str | None = Field(
```
jotare, line 1986: Should we hide it from the docs until we test it? Otherwise this will be published in the nucliadb-models before the server code reaches prod

nucliadb/src/nucliadb/search/search/search_after.py
```diff
@@ -0,0 +1,60 @@
...
+
+class SearchAfterToken(BaseModel):
+    # Indicates the first result to fetch from the index
+    after: SearchAfter | None = None
+
+    # Indicates the results already shown that should be skipped
+    skip: list[str] = Field(default_factory=list)
+
+    def encode(self) -> str:
+        return b64encode(self.model_dump_json().encode()).decode()
```
jotare, line 37: Didn't you want to do some obfuscation of the token?

nidx/src/searcher/shard_merge.rs
```diff
@@ -200,16 +200,16 @@ fn sort_documents_fn(order_by: OrderBy) -> Box<dyn Fn(&DocumentResult, &Document
...
                     Some(document_result::SortValue::Score(nidx_protos::ResultScore {
                         bm25: b_bm25,
-                        booster: b_booster,
+                        docaddr: b_docaddr,
                     })),
                 ) => a_bm25
                     .total_cmp(&b_bm25)
                     .then(a.shard_id.cmp(&b.shard_id))
-                    .then(a_booster.cmp(&b_booster))
+                    .then(a_docaddr.cmp(&b_docaddr))
```
jotare: Did you miss the reverse here?

nucliadb/src/nucliadb/search/search/find.py
```diff
@@ -89,6 +92,15 @@ async def _ndb_index_find(
     audit = get_audit()
     start_time = time()
 
+    # Disabled unsupported features for search_after and set the window to include enough results
+    if item.search_after is not None:
+        search_after = SearchAfterToken.decode(item.search_after)
+        item.features = [FindOptions.KEYWORD]
+        item.reranker = RerankerName.NOOP
```
jotare, line 99: Do we really want to ignore the reranker without telling the user? I think we should either return a 422 or mention it in the search_after docs (in the model at least)

nucliadb/src/nucliadb/search/search/find.py
```diff
@@ -98,11 +110,21 @@ async def _ndb_index_find(
...
+
     with metrics.time("index_search"):
         text_blocks, pb_query, pb_response, queried_shards = await text_block_search(
             kbid, parsed.retrieval
         )
 
+    # Remove skipped paragraphs for search_after
+    if search_after is not None:
+        text_blocks = [tb for tb in text_blocks if tb.paragraph_id.full() not in search_after.skip]
+        item.top_k -= len(search_after.skip)
```
jotare, line 125: Why are you subtracting this here? We don't use `item` anymore right?

nucliadb/tests/nucliadb/integration/search/test_search_after.py
javitonino: I can match the number but I think it's actually 10 (title and content). This was just a barebones check that there is enough to paginate.

nucliadb/tests/nucliadb/integration/search/test_search_after.py
javitonino: Yes. The order is not the same in both iterations due to differences in rank fusion window and such. I just want to make sure everything is returned and nothing is skipped. If anything I could add a check that things are returned in descending score order

nucliadb/src/nucliadb/search/search/search_after.py
javitonino: I was considering it, but I don't think it's needed. I don't think you can do much evil things (beside breaking your own search) with this.

nidx/src/searcher/shard_merge.rs
javitonino: Yeah, good catch. This is not really used now (since you cannot do search after on documents) but good to have it consistent.

nucliadb/src/nucliadb/search/search/find.py
javitonino: I'll add it do the docs whenever we release it. For now, I'll add it to the model. I mean, it doesn't make any sense to rerank at this point, it's just confusing, but better to be clear.

nucliadb/src/nucliadb/search/search/find.py
javitonino: I think it's used in logging but mainly doing it for consistency.

jotare (dismissed):

jotare (approved):

PR #3830, opened 2026-07-23
Backfill orm tables migration

lferran:
### Description
This has already been run in production, but we need to wire it in the automatic migrations so that onprem deployments get this aswell.

[sc-14838]

### How was this PR tested?


nucliadb/src/migrations/0050_backfill_orm_tables.py
```diff
@@ -0,0 +1,70 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+
+"""Migration #48
```
jotare:
```suggestion
"""Migration #50
```

jotare (dismissed):

jotare (dismissed):

nucliadb/tests/nucliadb/migrations/test_migration_0018.py
```diff
@@ -66,7 +68,7 @@ async def test_migration_0018_global(maindb_driver: Driver):
         # tikv needs to open a second transaction to be able to read values from
         # the first one using `scan_keys`
         async with maindb_driver.ro_transaction() as txn:
-            kb_slugs = [kb_slug async for kbid, kb_slug in datamanagers.kb.get_kbs(txn, prefix="")]
+            kb_slugs = [kb_slug async for kbid, kb_slug in datamanagers.kb.get_kbs(txn, slug_prefix="")]
```
jotare, line 71: Not sure the migration tests make sense anymore. You're using the tables in the test but the migration is still working with keys.
I don't think that's really important though

nucliadb/tests/nucliadb/integration/common/datamanagers/test_fields_v2.py
```diff
@@ -18,17 +18,15 @@
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 """
-Integration tests for nucliadb.common.datamanagers.fields_v2.
+Integration tests for nucliadb.common.datamanagers.fields.
```
jotare: You kept the filenames as test_fields_v2 and so on. I think should also remove the v2 from here.

nucliadb/src/nucliadb/common/datamanagers/fields.py
```diff
@@ -50,78 +87,242 @@ async def set(
...
-    keys = [
-        KB_RESOURCE_FIELD_STATUS.format(
-            kbid=kbid, uuid=rid, type=FIELD_TYPE_PB_TO_STR[fid.field_type], field=fid.field
+    txn: Transaction,
+    *,
+    kbid: str,
+    rid: str,
+    fields: Sequence[rpb2.FieldID],
+) -> list[wpb2.FieldStatus]:
+    """Return the deserialised FieldStatus for a list of fields, in the same order."""
```
jotare: I know those were already in the v2 files but I'm not a fan of all these redundant comments that just explain with words the already typed return value. (marking this as an example but there are plenty)

nucliadb/src/nucliadb/common/datamanagers/fields.py
```diff
@@ -50,78 +87,242 @@ async def set(
...
+        return []
+
+    async with _pg_cursor(txn) as cur:
+        await cur.execute(
+            f"""
+            SELECT field_type, field_id, status
+            FROM kb_fields
+            WHERE kbid = %s AND rid = %s
+              AND (field_type, field_id) IN (
+                {",".join(["(%s, %s)"] * len(fields))}
```
jotare, line 213: You can use ANY (https://www.postgresql.org/docs/current/functions-comparisons.html#FUNCTIONS-COMPARISONS-ANY-SOME) to avoid SQL query string manipulation. Then just map the fields list

nucliadb/tests/nucliadb/migrations/test_migration_0018.py
lferran: yeah, agree. it doesn't harm though. I'll remove it in another PR

nucliadb/tests/nucliadb/integration/common/datamanagers/test_fields_v2.py
lferran: done

nucliadb/src/nucliadb/common/datamanagers/fields.py
lferran: Unless you have a strong opinion on this one, I'd rather keep it as is. Using ANY isn't much nicer in my opinion

nucliadb/src/migrations/0050_backfill_orm_tables.py
```diff
@@ -494,8 +494,8 @@ class resources_v1:
 
     KB_RESOURCE_SHARD = "/kbs/{kbid}/r/{uuid}/shard"
 
-    @backoff.on_exception(backoff.expo, (Exception,), jitter=backoff.random_jitter, max_tries=3)
     @classmethod
+    @backoff.on_exception(backoff.expo, (Exception,), jitter=backoff.random_jitter, max_tries=3)
```
jotare: Not necessarily for now, but I think these backoff were from tikv times and pg is more reliable. Maybe we should remove them eventually

nucliadb/src/nucliadb/common/datamanagers/conversations.py
```diff
@@ -17,146 +17,224 @@
...
+                 page >= 1 → resources_pb2.Conversation (~200 messages each)
+
+The FieldConversation metadata (page count, total, page size, extract/split strategy)
+is stored in kb_fields.value for the corresponding 'c'-type field row and is written
+directly here to avoid a cross-module import cycle with fields.
+
+NOTE: deleting a kb_resources row (or its parent kbs row) automatically removes all
+related kb_conversations rows via the ON DELETE CASCADE foreign key.
+Deleting a single conversation field requires an explicit DELETE by (kbid, rid, field_id)
+because there is no FK from kb_conversations to kb_fields.
```
jotare: This is misleading, there is a FK from kb_conversation to kb_fields. We should make sure if we write docs they're accurate

nucliadb/src/nucliadb/common/file_md5.py
```diff
@@ -17,131 +17,27 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 #
+"""
+V2 implementation of file MD5 tracking using the kb_fields.md5 column
```
jotare: Another reference to v2

nucliadb/tests/nucliadb/migrations/test_migration_0050.py
```diff
@@ -0,0 +1,75 @@
...
+            )
+            await txn.set(
+                migration.module.resources_v1.KB_RESOURCE_SLUG.format(kbid=kbid, slug=slug),
+                rid.encode(),
+            )
+
+        await txn.commit()
+
+    # The test goal is coverage: this should complete without raising exceptions.
+    await migration.module.migrate_kb(execution_context, kbid)
```
jotare, line 76: It wouldn't harm adding a couple of asserts :)

nucliadb/src/nucliadb/common/datamanagers/resources.py
```diff
@@ -17,343 +17,349 @@
...
-    if datamanagers_v2_read(kbid):
-        return await resources_v2.get_basic(txn, kbid=kbid, rid=rid)
 
-    raw = await get_basic_raw(txn, kbid=kbid, rid=rid)
-    if raw is None:
-        return None
-    basic = resources_pb2.Basic()
-    basic.ParseFromString(raw)
-    return basic
+resource_exists = exists  # alias for backwards compatibility
```
jotare, line 248: Do we need bw/c for datamanagers? It is private to nucliadb

nucliadb/src/nucliadb/common/datamanagers/kb.py
```diff
@@ -17,107 +17,140 @@
...
+async def delete(txn: Transaction, *, kbid: str) -> None:
+    """Fully delete a KB row and all its associated resources, fields, and conversations."""
+    async with _pg_cursor(txn) as cur:
+        await cur.execute(
+            "DELETE FROM kbs WHERE kbid = %(kbid)s",
+            {"kbid": kbid},
+        )
+
+
+async def soft_delete(txn: Transaction, *, kbid: str) -> None:
```
jotare, line 115: NIT: delete or mark_delete and purge? To keep the same names as in ORM knowledgebox

nucliadb/src/nucliadb/common/datamanagers/resources.py
lferran: it was just to not add even more changes to the PR. I'll be fixing this later on, if that is ok

nucliadb/src/migrations/0050_backfill_orm_tables.py
lferran: yeah, can be removed

nucliadb/src/nucliadb/common/datamanagers/kb.py
lferran: I prefer leaving the soft_delete because not only marks it for deletion, but also frees the slug to be used by other kbs.

I'll also tackle all the purge logic in a later PR

jotare (approved):

PR #3831, opened 2026-07-23
Remove merging logic from search

jotare:
### Description
Merge logic is already in nidx, we can get rid of it in search

### How was this PR tested?

### DIFF
--- a/nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
+++ b/nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
@@ -196,6 +196,7 @@ async def get_node_index_counts(kbid: str) -> tuple[IndexCounts, list[str]]:
         raise HTTPException(status_code=503, detail=f"No shards found")
 
     counts = IndexCounts(fields=0, paragraphs=0, sentences=0, size_bytes=0)
+    print(results)
     for shard in results:
         if isinstance(shard, BaseException):
             logger.error("Error getting shard info", exc_info=shard)
@@ -205,4 +206,5 @@ async def get_node_index_counts(kbid: str) -> tuple[IndexCounts, list[str]]:
         counts.paragraphs += shard.paragraphs
         counts.sentences += shard.sentences
         counts.size_bytes += shard.size_bytes
+    print(counts)
     return counts, queried_shards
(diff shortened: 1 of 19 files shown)


nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
```diff
@@ -196,6 +196,7 @@ async def get_node_index_counts(kbid: str) -> tuple[IndexCounts, list[str]]:
         raise HTTPException(status_code=503, detail=f"No shards found")
 
     counts = IndexCounts(fields=0, paragraphs=0, sentences=0, size_bytes=0)
+    print(results)
```
javitonino: print

nucliadb/src/nucliadb/search/api/v1/knowledgebox.py
```diff
@@ -205,4 +206,5 @@ async def get_node_index_counts(kbid: str) -> tuple[IndexCounts, list[str]]:
         counts.paragraphs += shard.paragraphs
         counts.sentences += shard.sentences
         counts.size_bytes += shard.size_bytes
+    print(counts)
```
javitonino: pront

nucliadb/src/nucliadb/search/requesters/utils.py
```diff
@@ -104,16 +75,15 @@ async def nidx_query(
...
 
 async def nidx_query(
     kbid: str,
     method: Method,
-    pb_query: REQUEST_TYPE,
+    pb_query: SearchRequest | SuggestRequest | GraphSearchRequest,
     timeout: float | None = None,
-) -> tuple[Sequence[T | BaseException], list[str]]:
-    timeout = timeout or settings.search_timeout
+) -> tuple[SearchResponse | SuggestResponse | GraphSearchResponse | BaseException, list[str]]:
```
javitonino: I think BaseException is not needed here (exceptions are raised, not returned)

nucliadb/src/nucliadb/search/requesters/utils.py
```diff
@@ -123,95 +93,79 @@ async def nidx_query(
             detail="The knowledgebox or its shards configuration is missing",
         )
 
-    ops = []
     queried_shards = []
+    for shard_obj in shard_groups:
+        if shard_obj.nidx_shard_id is not None:
+            queried_shards.append(shard_obj.nidx_shard_id)
```
javitonino: Why build this from the database rather than from the response protobuffer?

Also, if we can return this from the protobuffer, maybe this function can just return the response, and the callers can extract the list of shards directly, making the signature of this method cleaner.

jotare:
[sc-14854]

javitonino (approved):

PR #3833, opened 2026-07-24
Use multi-shard search for entities

jotare:
### Description
Use multi-shard search in nidx instead of querying all shards

### How was this PR tested?


jotare:
[sc-14854]

javitonino (approved):

PR #3847, opened 2026-08-05
Allow choosing vectorset when retrieving embeddings of fields

lferran:
### Description
[sc-15251]

### How was this PR tested?


jotare (commented):
If we want to be consistent with the API, I think we should add the vectorset to `ResourceHydrationOptions` and use it in the augmentor as well as the /find endpoint. I know returning vectors in search is discouraged but we still support it

jotare (approved):

PR #3848, opened 2026-08-05
Improve KB datamanager and usage

lferran:
### Description

### How was this PR tested?


nucliadb/src/nucliadb/common/datamanagers/kb.py
```diff
@@ -30,20 +30,175 @@
...
-from . import cluster
+from nucliadb_protos import knowledgebox_pb2, writer_pb2
+from nucliadb_telemetry.metrics import Observer
 
 logger = logging.getLogger(__name__)
 
 
-async def get_kbs(txn: Transaction, *, slug_prefix: str = "") -> AsyncIterator[tuple[str, str]]:
+observer = Observer(
+    "datamanagers.kb",
```
jotare: We should prefix our metrics, e.g., nucliadb_datamanagers.

I'm not sure we want an observer per datamanager. Building the dashboard will need as many queries as datamanagers and at Prometheus level, it'll be the same number of time series. May be easier to add a name or datamanager label

nucliadb/src/nucliadb/common/datamanagers/kb.py
```diff
@@ -30,20 +30,175 @@
...
+UNSET: Final = _UnsetType()
+UNSET_STR: Final[str | None] = cast(str | None, UNSET)
+UNSET_CONFIG: Final[knowledgebox_pb2.KnowledgeBoxConfig | None] = cast(
+    knowledgebox_pb2.KnowledgeBoxConfig | None, UNSET
+)
+UNSET_SHARDS: Final[writer_pb2.Shards | None] = cast(writer_pb2.Shards | None, UNSET)
+UNSET_DELETED_AT: Final[datetime | None] = cast(datetime | None, UNSET)
+
+
+@dataclass(slots=True)
```
jotare, line 64: Why `slots=True`?

nucliadb/src/nucliadb/common/datamanagers/exceptions.py
```diff
@@ -19,3 +19,7 @@
 #
 class KnowledgeBoxNotFound(Exception):
     pass
+
+
+class KnowledgeBoxConflict(Exception):
```
jotare, line 24: Why do we need an extra extra conflict exception?

nucliadb/src/nucliadb/common/datamanagers/kb.py
lferran: yep, ok! I can set the datamanager type as a label sounds good too

nucliadb/src/nucliadb/common/datamanagers/exceptions.py
lferran: The KnowledgeBoxNotFound exception is also "duplicated" both in datamanagers and in orm code.
I followed the same pattern for now. I think it's ok, it doesn't harm, but we can unify them too -- although that may require many more small changes

nucliadb/src/nucliadb/common/datamanagers/kb.py
lferran: I didn't know this either, but it restricts instance attributes to a fixed set and saving a bit of memory.
Not that it is very needed here, but it doesn't harm either

nucliadb/src/nucliadb/common/datamanagers/kb.py
```diff
@@ -30,20 +30,161 @@
...
+        return kb_config
+    if column == "shards":
+        shards = writer_pb2.Shards()
+        shards.ParseFromString(bytes(value))
+        return shards
+    raise ValueError(f"Unknown KB column: {column}")  # pragma: no cover
+
+
+@observer.wrap({"type": "kb", "op": "upsert"})
+async def upsert(
```
jotare: Could you use `set` instead so we don't have more words for the same thing?
```suggestion
@observer.wrap({"type": "kb", "op": "set"})
async def set(
```

jotare (dismissed):

jotare (approved):

PR #3871, opened 2026-08-19
Less pg load on augmentor

lferran:
### Description
- Removed unnecessary field existance checks
- Optimized getting shards for resources

### How was this PR tested?
Existing shards cover us.

### DIFF
--- a/nucliadb/src/nucliadb/common/datamanagers/resources.py
+++ b/nucliadb/src/nucliadb/common/datamanagers/resources.py
@@ -385,6 +385,25 @@ async def get_shard(txn: Transaction, *, kbid: str, rid: str, for_update: bool =
     return resource.shard
 
 
+@observer.wrap({"type": "resources", "op": "get_shards"})
+async def get_shards(txn: Transaction, *, kbid: str, rids: list[str]) -> dict[str, str]:
+    """Return a mapping of rid → shard for the given rids in a single query."""
+    result: dict[str, str] = {}
+    if not rids:
+        return result
+    placeholders = [psycopg.sql.Placeholder(f"rid_{i}") for i in range(len(rids))]
+    query = psycopg.sql.SQL(
+        "SELECT rid, shard FROM kb_resources WHERE kbid = %(kbid)s AND rid IN ({placeholders})"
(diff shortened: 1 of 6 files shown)


nucliadb/src/nucliadb/common/datamanagers/resources.py
```diff
@@ -385,6 +385,25 @@ async def get_shard(txn: Transaction, *, kbid: str, rid: str, for_update: bool =
...
+async def get_shards(txn: Transaction, *, kbid: str, rids: list[str]) -> dict[str, str]:
+    """Return a mapping of rid to shard for the given rids in a single query."""
+    result: dict[str, str] = {}
+    if not rids:
+        return result
+    placeholders = [psycopg.sql.Placeholder(f"rid_{i}") for i in range(len(rids))]
+    query = psycopg.sql.SQL(
+        "SELECT rid, shard FROM kb_resources WHERE kbid = %(kbid)s AND rid IN ({placeholders})"
+    ).format(placeholders=psycopg.sql.SQL(", ").join(placeholders))
+    params = {"kbid": kbid, **{f"rid_{i}": rid for i, rid in enumerate(rids)}}
```
jotare: I found this query a bit tricky to read. Would you mind using `IN ANY(%(rids))` or something simpler?
(https://www.psycopg.org/psycopg3/docs/basic/adapt.html#lists-adaptation)

nucliadb/src/nucliadb/search/augmentor/paragraphs.py
```diff
@@ -60,15 +60,8 @@ async def augment_paragraph(
     if resource is None:
         # skip resources that aren't in the DB
         return None
-
     field_id = paragraph_id.field_id
-    # we must check if field exists or get_field will return an empty field
-    # (behaviour thought for ingestion) that we don't want
```
jotare, line 66: Could you add a comment telling that we don't care about the field not existing?

nucliadb/src/nucliadb/common/datamanagers/resources.py
lferran: you're right, plus pg can optimize the query as it will be always athe same with ANY, instead of dynamic placeholders.

jotare (approved):

PR #3878, opened 2026-08-26
Augment messages content

lferran:
### Description
This PR extends the augment endpoint to be able to efficiently fetch the raw content of multiple conversation messages in a single request.

This is used in the memory feature, where we need to fetch the original json content for memory entries that are retrieval matches or cited text blocks.
[sc-15136]

### How was this PR tested?
Extended existing test suite for augmentor and augment endpoint

### DIFF
--- a/nucliadb/src/nucliadb/models/internal/augment.py
+++ b/nucliadb/src/nucliadb/models/internal/augment.py
@@ -34,7 +34,7 @@ from nucliadb.common.ids import FieldId, ParagraphId
 from nucliadb_models import filters
 from nucliadb_models.augment import ResourceId
 from nucliadb_models.common import FieldTypeName
-from nucliadb_models.conversation import FieldConversation
+from nucliadb_models.conversation import FieldConversation, MessageFormat
 from nucliadb_models.file import FieldFile
 from nucliadb_models.link import FieldLink
 from nucliadb_models.metadata import Extra, Origin
@@ -282,6 +282,17 @@ class MessageSelector(BaseModel):
         return self
 
 
+class MessagesSelector(BaseModel):
+    """Selects the conversation field messages specified by the message ids."""
+
+    name: Literal["messages"] = "messages"
+
+    ids: list[str] = Field(
+        default_factory=list,
+        description="List of message identifiers to select",
+    )
+
+
 class PageSelector(BaseModel):
     """Selects all messages from the page of the message specified by the field
     id.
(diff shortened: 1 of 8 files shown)


nucliadb/src/nucliadb/models/internal/augment.py
```diff
@@ -348,11 +360,22 @@ class FullSelector(BaseModel):
 ]
 
 
+ConversationValueSelector = Annotated[
+    (Annotated[MessagesSelector, Tag("messages")]),
+    Discriminator(name_discriminator),
+]
+
+
```
jotare: I you already added the MessagesSelector to the ConversationSelector, why do you need an extra selector type?

nucliadb/src/nucliadb/search/augmentor/fields.py
```diff
@@ -295,19 +301,35 @@ async def db_augment_conversation_field(
...
                     augmented_message.text = message.content.text
+                    augmented_message.format = from_proto.message_content_format(message.content.format)
 
         elif isinstance(prop, FieldValue):
             db_value = await field.get_metadata()
             if db_value is None:
                 continue
             augmented.value = from_proto.field_conversation(db_value)
 
+            if isinstance(prop, ConversationValue):
```
jotare: Untested and unreachable

nucliadb/src/nucliadb/models/internal/augment.py
lferran: Because we don't offer all of the previous selectors on the ConversationValue prop right now, do we?

nucliadb/tests/search/integration/search/test_augmentor.py
```diff
@@ -402,6 +404,28 @@ async def test_augmentor_conversation_strategy_text(
         assert len(augmented_field.messages) == 5
         assert {m.ident for m in augmented_field.messages} == {"1", "5", "7", "10", "12"}
 
+        # Augmenting with the messages selector returns only those ids and
+        # includes the message format.
```
jotare, line 408: We now always return the message format, not only with this selector, right?

nucliadb_models/src/nucliadb_models/augment.py
```diff
@@ -210,6 +216,28 @@ def validate_cross_options(self):
...
+                parts = field_id.split("/")
+                if len(parts) != 4:
+                    raise ValueError(
+                        "`conversation_message_raw_content` requires all given field ids to have a subfield_id (aka: ident) in the field id. Field ids: "
+                        + ", ".join(self.given)
+                    )
+                prefixes.add("/".join(parts[:3]))
+            if len(prefixes) > 1:
+                raise ValueError(
+                    "`conversation_message_raw_content` requires all given field ids to be from the same conversation field. Field ids: "
```
jotare: Why do we require this? We can return multiple conversations in the same call to augment

lferran: yes but they all need to be in the same field augment term, no?

nucliadb/tests/search/integration/search/test_augmentor.py
lferran: no, only if the raw content is returned

jotare (approved):

PR #3879, opened 2026-08-26
Fix nats ingest-subscriber-workers ConsumerSequenceMismatchError

jotare:
### Description
Push consumers with both `queue` and `idle_heartbeat` are wrongly configured, leading to `ConsumerSequenceMismatchError`s on autoscaling. As we do want to be able to scale and we do have metrics for pending message, we can already detect if the client and consumer are no longer connected (which is the purpose of `idle_heartbeat`).

The proper solution however is to move away from push consumers and use pull consumers instead.

See: https://github.com/nats-io/nats-architecture-and-design/blob/main/adr/ADR-15.md#configuration for further reference

### How was this PR tested?


javitonino:
Replaced by https://github.com/nuclia/nucliadb/pull/3889

PR #3890, opened 2026-09-01
allow specifying file field name via tus metadata

drf7:
**Allow specifying file field name via TUS metadata**

When using the KB-level TUS upload endpoint (`POST /kb/{kbid}/tusupload`), the file field ID is always auto-generated as a random UUID. This makes it impossible to update the same field later — uploading to a resource-scoped endpoint with a different field name creates a new field instead of replacing the existing one.

This change allows clients to pass a `field` key in the `Upload-Metadata` TUS header to control the field name. If not provided, behavior is unchanged (random UUID).

**Example:**
```
Upload-Metadata: filename dGVzdC5wZGY=,field ZmlsZQ==
```
(where `ZmlsZQ==` is base64 for `"file"`)

This enables idempotent upload flows where the same field name is used for both creation and subsequent updates.

### DIFF
--- a/nucliadb/src/nucliadb/writer/api/v1/upload.py
+++ b/nucliadb/src/nucliadb/writer/api/v1/upload.py
@@ -276,9 +276,9 @@ async def _tus_post(
         metadata = {}
 
     # Allow overriding the field name via TUS metadata (e.g., for idempotent updates)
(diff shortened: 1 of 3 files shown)


lferran (approved):
lgtm!

PR #3903, opened 2026-09-14
Send title slug on upload resource

arnaugrispg:
### Description
Ensures Processing payloads include the resource `title` and `slug` across uploads, field updates, resource updates, and reprocessing.

Values from the request take precedence when available. Otherwise, the stored `Basic` metadata is used. Reprocessing reuses the already loaded resource metadata.

### How was this PR tested?
Tests cover `title` and `slug` propagation for each affected Writer entry point, including resource updates with new metadata and fallback to stored metadata.


nucliadb/src/nucliadb/writer/api/v1/resource.py
```diff
@@ -300,6 +301,7 @@ async def modify_resource(
     toprocess.kbid = kbid
     toprocess.uuid = rid
     toprocess.source = Source.HTTP
+    await set_processing_metadata_from_basic(toprocess, kbid, rid)
```
javitonino: I think this is incorrect. You are moving this call before `parse_basic_modify` which can tweak the title.

Also, you are reading from the stored basic in `set_processing_metadata_from_basic` so that's always the data in the database, which will not be correct if we are modifying the title.

And, in general, we try to avoid reading basic in the writer if possible, keeping it to ingest, to avoid issues with transactionality.

javitonino (approved):

PR #3914, opened 2026-09-17
Add S3 timeouts

jotare:
### Description

### How was this PR tested?


nucliadb_utils/src/nucliadb_utils/storages/s3.py
```diff
@@ -391,6 +391,8 @@ def __init__(
         if disable_checksums:
             config = aiobotocore.config.AioConfig(
                 None,
+                connect_timeout=30,
+                read_timeout=10,
```
lferran, line 395: we want the timeouts in both branches (if disable_checksums), don't we?

lferran (approved):

